PR #11, opened 2016-12-19
Add rules orchestration

fmueller:
This pr adds rules orchestration and integrates it for happy path into the controller.

### DIFF
--- a/src/test/java/de/zalando/zally/RestApiTest.java
+++ b/src/test/java/de/zalando/zally/RestApiTest.java
@@ -2,21 +2,62 @@ package de.zalando.zally;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import de.zalando.zally.rules.Rule;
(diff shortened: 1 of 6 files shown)


src/test/java/de/zalando/zally/RestApiTest.java
```diff
@@ -2,34 +2,72 @@
...
+
+            private final String apiName;
+
+            public CheckApiNameIsPresentRule(String apiName) {
+                this.apiName = apiName;
+            }
+
+            @Override
+            public List<Violation> validate(Swagger swagger) {
+                if (swagger != null && swagger.getInfo().getTitle().contains(apiName)) {
```
netme: You need to be careful here, I assume `getInfo()` can be `null`.

netme: https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-models/src/main/java/io/swagger/models/Swagger.java#L17

fmueller: Yes, because of test and because I'm controlling the input data for it, I do not see a big problem. The test would only fail if there would be an input validation issue.

PR #13, opened 2016-12-20
Define OAuth scopes: Added rule implementation class and test class

thomasabraham:


src/main/java/de/zalando/zally/rules/DefineOAuthScopesRule.java
```diff
@@ -0,0 +1,107 @@
...
+
+    // Extract all oauth2 scopes applied to the given operation into a simple list
+    private List<String> extractAppliedScopes(Operation operation) {
+        List<String> scopes = new ArrayList<>();
+        List<Map<String, List<String>>> security = operation.getSecurity();
+        if (security != null && !security.isEmpty()) {
+            security.forEach(securityMap -> {
+                if (securityMap != null) {
+                    securityMap.forEach((key, value) -> {
+                        if (OAUTH2.equals(key) && value != null) {
```
mkulak: We don't need to iterate over each key in Map here. Just use `Map.get(OAUTH2)`.

src/test/java/de/zalando/zally/rules/DefineOAuthScopesRuleTest.java
```diff
@@ -0,0 +1,70 @@
...
+    private static final String API_WITHOUT_SCOPE = "api_without_scopes_defined";
+    private static final int API_WITHOUT_SCOPE_VIOLATION_COUNT = 4;
+    private static final String API_WITH_DEFAULT_SCOPE = "api_with_default_scope";
+    private static final int API_WITH_DEFAULT_SCOPE_VIOLATION_COUNT = 0;
+    private static final String API_WITH_DEFINED_SCOPE = "api_with_defined_scope";
+    private static final int API_WITH_DEFINED_SCOPE_VIOLATION_COUNT = 0;
+    private static final String API_WITH_UNDEFINED_SCOPE = "api_with_undefined_scope";
+    private static final int API_WITH_UNDEFINED_SCOPE_VIOLATION_COUNT = 4;
+    private static final String API_WITH_DEFINED_AND_UNDEFINED_SCOPE = "api_with_defined_and_undefined_scope";
+    private static final int API_WITH_DEFINED_AND_UNDEFINED_SCOPE_VIOLATION_COUNT = 2;
```
mkulak: I suggest to inline all these constants since they don't make any sense outside of context

PR #15, opened 2016-12-20
Snake case for query params

nehalium:


src/main/java/de/zalando/zally/rules/SnakeCaseForQueryParamsRule.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private static String SNAKE_CASE_PATTERN = "^[a-z0-9_]*$";
+
+    List<Violation> validate(Swagger swagger) {
+        if (swagger == null) return new ArrayList<>();
+
+        List<Violation> violations = new ArrayList<>();
+        swagger.getPaths().forEach((path, pathObject) -> {
+            pathObject.getOperationMap().forEach((httpMethod, operation) -> {
+                operation.getParameters().forEach(param -> {
```
mkulak: operation.getParameters() can be null in case of operation without parameters

src/main/java/de/zalando/zally/rules/SnakeCaseForQueryParamsRule.java
```diff
@@ -0,0 +1,53 @@
...
+    private boolean isSnakeCase(String input) {
+        Pattern pattern = Pattern.compile(SNAKE_CASE_PATTERN);
+        Matcher matcher = pattern.matcher(input);
+        return matcher.matches();
+    }
+
+    private Violation getViolation() {
+        return new Violation(
+            "Must: Use snake_case (never camelCase) for Query Parameters",
+            "Must: Use snake_case (never camelCase) for Query Parameters",
```
mkulak: Title is the same as description which is not very useful. I suggest to put in description actual parameter names that violates the rule

nehalium: Updated. Please take a look.

src/main/java/de/zalando/zally/rules/SnakeCaseForQueryParamsRule.java
nehalium: I am testing for this and it appears not to be null when the parameters property is missing. The object simply has 0 parameters.

PR #18, opened 2016-12-20
Implement rule Avoid Link Headers

ishishov:


src/main/java/de/zalando/zally/rules/AvoidLinkHeadersRule.java
```diff
@@ -0,0 +1,43 @@
+package de.zalando.zally.rules;
+
+import de.zalando.zally.Violation;
+import de.zalando.zally.ViolationType;
+import io.swagger.models.*;
```
netme: Can we avoid `*` imports?

src/main/java/de/zalando/zally/rules/AvoidLinkHeadersRule.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Override
+    public List<Violation> validate(Swagger swagger) {
+        List<Violation> violations = new ArrayList();
+        if (swagger != null && swagger.getPaths() != null) {
+            for (Map.Entry<String, Path> path : swagger.getPaths().entrySet())
+                for (Map.Entry<HttpMethod, Operation> op : path.getValue().getOperationMap().entrySet()) {
+                    for (Parameter param : op.getValue().getParameters()) {
+                        if ("header".equals(param.getIn()) && "Link".equals(param.getName())) {
+                            String violationPath = path.getKey() + "/" + op.getKey() + "/" + param.getName();
```
netme: Is it possible to somehow simplify this loop? Currently we have 5 levels, maybe it's good to move some logic into some subfunctions?

src/test/java/de/zalando/zally/rules/AvoidLinkHeadersRuleTest.java
```diff
@@ -0,0 +1,33 @@
...
+        AvoidLinkHeadersRule rule = new AvoidLinkHeadersRule();
+        Swagger swagger = new SwaggerParser().read("/api_spa.yaml");
+        assertThat(rule.validate(swagger)).isEmpty();
+    }
+
+    @Test
+    public void avoidLinkHeadersInvalidJson() throws IOException {
+        AvoidLinkHeadersRule rule = new AvoidLinkHeadersRule();
+        Swagger swagger = new SwaggerParser().read("/fixtures/avoidLinkHeaderRuleInvalid.json");
+        assertThat(rule.validate(swagger).size()).isEqualTo(2);
```
netme: Maybe it's also good to test which exact violations are happening

src/main/java/de/zalando/zally/rules/AvoidLinkHeadersRule.java
```diff
@@ -0,0 +1,43 @@
...
+                            violations.add(new Violation(TITLE, DESCRIPTION, VIOLATION_TYPE, RULE_LINK, violationPath));
+                        }
+                    }
+                    for (Map.Entry<String, Response> response : op.getValue().getResponses().entrySet()) {
+                        if (response.getValue().getHeaders() != null && response.getValue().getHeaders().containsKey("Link")) {
+                            String violationPath = path.getKey() + "/" + op.getKey() + "/" + response.getKey();
```
netme: You can also use `String.join()` here

src/test/java/de/zalando/zally/rules/AvoidLinkHeadersRuleTest.java
mkulak: I suggest to add validation for Violation's fields also.

src/main/java/de/zalando/zally/rules/AvoidLinkHeadersRule.java
```diff
@@ -0,0 +1,43 @@
...
+public class AvoidLinkHeadersRule implements Rule {
+    private static final String TITLE = "Avoid Link in Header Rule";
+    private static final String DESCRIPTION = "Do Not Use Link Headers with JSON entities";
+    private static final ViolationType VIOLATION_TYPE = ViolationType.MUST;
+    private static final String RULE_LINK = "http://zalando.github.io/restful-api-guidelines/hyper-media/Hypermedia.html#must-do-not-use-link-headers-with-json-entities";
+
+    @Override
+    public List<Violation> validate(Swagger swagger) {
+        List<Violation> violations = new ArrayList();
+        if (swagger != null && swagger.getPaths() != null) {
```
superbenji: can we include the check for null for the swagger object in the controller as it effects all rules?

src/main/java/de/zalando/zally/rules/AvoidLinkHeadersRule.java
```diff
@@ -0,0 +1,43 @@
...
+                    for (Parameter param : op.getValue().getParameters()) {
+                        if ("header".equals(param.getIn()) && "Link".equals(param.getName())) {
+                            String violationPath = path.getKey() + "/" + op.getKey() + "/" + param.getName();
+                            violations.add(new Violation(TITLE, DESCRIPTION, VIOLATION_TYPE, RULE_LINK, violationPath));
+                        }
+                    }
+                    for (Map.Entry<String, Response> response : op.getValue().getResponses().entrySet()) {
+                        if (response.getValue().getHeaders() != null && response.getValue().getHeaders().containsKey("Link")) {
+                            String violationPath = path.getKey() + "/" + op.getKey() + "/" + response.getKey();
+                            violations.add(new Violation(TITLE, DESCRIPTION, VIOLATION_TYPE, RULE_LINK, violationPath));
```
mkulak: Please refactor common code among 25-27 and 31-33 in separate method.

src/main/java/de/zalando/zally/rules/AvoidLinkHeadersRule.java
ishishov: looks like IDE automatically does it when there's 5+ imports from the same package, reverted

src/main/java/de/zalando/zally/rules/AvoidLinkHeadersRule.java
ishishov: there are 3 loops levels + conditions which is ok, 'for' loops in this case is more readable than lambda on streams. But I will move violation creation in a separate method

src/main/java/de/zalando/zally/rules/AvoidLinkHeadersRule.java
ishishov: I think we can add it to the controller, but it doesn't mean it shouldn't be here as well, right?

src/test/java/de/zalando/zally/rules/AvoidLinkHeadersRuleTest.java
ishishov: will compare the validation name and type also

netme:
Would be great if you'll squash latest commits. 

```
git rebase -i 	6146614
````

PR #23, opened 2016-12-20
refactored to PatternUtil

nehalium:


src/test/java/de/zalando/zally/rules/SnakeCaseInPropNameRuleTest.java
```diff
@@ -23,37 +23,37 @@
 
     @Test
     public void isSnakeCaseForEmpty() {
-        assertThat(isSnakeCase("")).isFalse();
+        assertThat(PatternUtil.isSnakeCase("")).isFalse();
```
cjhowald: These tests should all be removed, since they are are now reduced to just testing the `PatternUtil` function, which is the responsibility of `PatternUtilTest`.

PR #25, opened 2016-12-20
Implement 'use-problem-json' rule

s0k0:


src/main/java/de/zalando/zally/rules/UseProblemJsonRule.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    public List<Violation> validate(Swagger swagger) {
+        List<Violation> violations = new ArrayList<>();
+        if (swagger == null || swagger.getPaths() == null) {
+            return violations;
+        }
+        for(Map.Entry<String, Path> path: swagger.getPaths().entrySet())
+            for(Map.Entry<HttpMethod, Operation> op : path.getValue().getOperationMap().entrySet())
+                for(Map.Entry<String, Response> response: op.getValue().getResponses().entrySet())
+                    try {
```
netme: Is it possible to minimise the scope of `try..catch` block? Probably the following scope should be good enough:

```java
try {
     Integer httpCodeInt = Integer.valueOf(response.getKey());
} catch(NumberFormatException nfe) {
}
```

src/main/java/de/zalando/zally/rules/UseProblemJsonRule.java
```diff
@@ -0,0 +1,46 @@
...
+                        Integer httpCodeInt = Integer.valueOf(response.getKey());
+                        //TODO: check if schema reference type is Problem := https://github.com/swagger-api/swagger-parser/pull/303/files
+                        if (httpCodeInt >= 400 && httpCodeInt < 600 && response.getValue().getSchema() != null
```
netme: Optional: `<= 599` will be easy to understand. Took me a while to understand why 600 :-)

s0k0: Done.

src/main/java/de/zalando/zally/rules/UseProblemJsonRule.java
s0k0: Agreed, I edited the scope.

ishishov:
👍

PR #27, opened 2016-12-20
Media type versioning

superbenji:
* added rule for media type checking
* optimized imports

Possible optimization: move pattern to pattern util BUT: it can not really be reused by other rules.


fmueller:
You should change your logic: you can only add version parameter to custom media types - not application/json.

Can we also add another better readable violation for "do not use custom media types"? If media type is not application/json and no media type versioning is used, this violation should be created.

fmueller:
It's basically:

is there another media type used than application/json or application/problem+json?
  yes:
    is media type versioning used?
      yes: ok
      no: violation, only use custom media types for media type versioning
  no: violation, do not use custom media types

src/main/java/de/zalando/zally/rules/MediaTypesRule.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Map;
+
+public class MediaTypesRule implements Rule{
+
+    private static final String TITLE = "Prefer standard media type names";
+    private static final String DESCRIPTION = "Custom media types should only be used for versioning";
+    private static final String RULE_LINK = "http://zalando.github.io/restful-api-guidelines/data-formats/DataFormats.html"+
+            "#should-prefer-standard-media-type-name-applicationjson";
+    private static final String PATTERN_APPLICATION_PROBLEM_JSON = "^application/(problem\\+)?json$";
+    private static final String PATTERN_CUSTOM_WITH_VERSIONING = "^\\w+/[-+.\\w]+;v(ersion)?=\\d+$";
```
netme: Should patterns be a part of `PatternUtil`?

superbenji: I'm not sure if this patterns can be reused by anybody. that is why I left them in the rule itself. But I can move them of cause if you think it is better to have them in the utils

fmueller: if they cannot be reused, patterns should be part of rule itself imho

PR #34, opened 2016-12-21
Pluralize resource names

mkulak:


src/main/java/de/zalando/zally/rules/PluralizeResourceNamesRule.java
```diff
@@ -0,0 +1,49 @@
...
+                        .filter(s -> !s.isEmpty() && !PatternUtil.isPathVariable(s))
+                        .filter(s -> !isPlural(s))
+                        .map(s -> createViolation(path, s))
+                        .collect(Collectors.toList()));
+            }
+        }
+        return res;
+    }
+
+    public static boolean isPlural(String resource) {
```
nehalium: I would refactor this to a separate Pluralizer class so that it can be tested separately.

src/test/java/de/zalando/zally/rules/PluralizeResourceNamesRuleTest.java
```diff
@@ -0,0 +1,92 @@
...
+                .validate(swagger)
+                .stream()
+                .map(v -> v.getPath().get())
+                .collect(Collectors.toList());
+        assertThat(results).hasSameElementsAs(Arrays.asList("/queue/configs/{config-id}", "/queue/models",
+                "/queue/models/{model-id}", "/queue/summaries"));
+    }
+
+    @Test
+    public void checkIsPluralized() {
```
nehalium: If you have a separate Pluralizer class, you can test it separately from the rule (separation of concerns).

PR #38, opened 2016-12-21
Hypen pascal case http headers

cjhowald:
- Add new hyphen pascal case http headers SHOULD rule
- Improve existing hypenated http headers MUST rule to check response headers and include path info in the violation (when exists)
- Refactored shared HTTP parsing functionality into an abstract HttpHeadersRule

### DIFF
--- a/src/main/java/de/zalando/zally/rules/HttpHeadersRule.java
+++ b/src/main/java/de/zalando/zally/rules/HttpHeadersRule.java
@@ -1,0 +1,75 @@
+package de.zalando.zally.rules;
+
+import de.zalando.zally.Violation;
+import io.swagger.models.Operation;
+import io.swagger.models.Path;
+import io.swagger.models.Response;
+import io.swagger.models.Swagger;
+import io.swagger.models.parameters.Parameter;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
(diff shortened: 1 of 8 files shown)


src/main/java/de/zalando/zally/rules/HttpHeadersRule.java
```diff
@@ -0,0 +1,75 @@
...
+import io.swagger.models.parameters.Parameter;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+public abstract class HttpHeadersRule implements Rule {
+
+    public static final Set<String> PARAMETER_NAMES_WHITELIST = new HashSet<>(Arrays.asList("ETag", "TSV", "TE",
+            "Content-MD5", "DNT", "X-ATT-DeviceId", "X-UIDH", "X-Request-ID", "X-Correlation-ID", "WWW-Authenticate",
+            "X-XSS-Protection"));
```
fmueller: Does it makes sense to also add X-Flow-ID header and all other prop headers mentioned http://zalando.github.io/restful-api-guidelines/headers/ProprietaryHeaders.html

cjhowald: Added all prop headers (that aren't already hyphenated pascal case) to the whitelist, plus a unit test to confirm.

PR #40, opened 2016-12-21
add rule: consider non-nested urls

ishishov:


src/main/java/de/zalando/zally/rules/NestedPathsCouldBeRootPathsRule.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public List<Violation> validate(Swagger swagger) {
+        List<Violation> violations = new ArrayList<>();
+        if (swagger == null || swagger.getPaths() == null) {
+            return violations;
+        }
+        for (String path : swagger.getPaths().keySet()) {
+            String[] pathSegments = path.split("/");
+            // we are only interested in paths that have sub-resource followed by a param: /path1/{param1}/path2/{param2}
+            if (pathSegments.length > 4) {
```
nehalium: Please refactor the threshold value to a class-level static final variable. Also update the description to:
`"Number of path segments (%s) is more than the recommended limit (%s). Nested paths / URLs could be top-level resources."`
and change createViolation by changing description to:
`String.format(DESCRIPTION, numberOfSegments, SEGMENT_LIMIT);`

ishishov: as discussed, agreed to leave it like it was

PR #43, opened 2016-12-21
snake case in json response and api fixes

superbenji:
* zall-api fixed (-> no violations!!!!)
* snake case json for violation object
* more concrete check in NestedPathsCouldBeRootPathsRule rule


fmueller:
LGTM

PR #46, opened 2016-12-21
Topic/GitHub integration

mkulak:
To run you need top set 2 environment variables:
github.token=<token to ur repo>
external.url=<externally visible url of your service>

It definitely needs some love in details.ftl (front-end page template for displaying violations)

### DIFF
--- a/src/main/java/de/zalando/zally/GithubController.java
+++ b/src/main/java/de/zalando/zally/GithubController.java
@@ -14,6 +14,7 @@ import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.client.RestClientException;
 import org.springframework.web.client.RestTemplate;
 
 import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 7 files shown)


src/main/java/de/zalando/zally/GithubController.java
```diff
@@ -0,0 +1,64 @@
...
+        ObjectNode response = mapper.createObjectNode().put("status", "ok");
+        if (payload.get("pull_request") != null) {
+            String repoUrl = payload.get("repository").get("url").asText();
+            String commitSha = payload.get("pull_request").get("head").get("sha").asText();
+
+            githubClient.setCommitStatus(repoUrl, commitSha, "pending", "", "Waiting for Zally approval");
+            try {
+                String swaggerPath = "swagger/swagger.yaml"; //TODO MK: unhardcode it
+                String swaggerContent = githubClient.getFileContent(repoUrl, commitSha, swaggerPath);
+                Swagger swagger = new SwaggerParser().parse(swaggerContent);
```
fmueller: Please also check for swagger == null in this case which means swagger file was not parsable --> break PR

mkulak: I addressed this by introducing `RulesValidator.validate(String)`. Since `RulesValidator` is also a Rule itself, rising violation in case of unparsable swagger file looks like his responsibility. But I can move that logic to some other place of your choice.

fmueller:
LGTM

Really like your additions to RulesValidator. 👍 

Can we merge this and consider improvements to detail page in another PR?

PR #65, opened 2016-12-26
Discussion: Kotlin

mkulak:
I propose to introduce Kotlin in this project.
**Pros:**
- Kotlin is modern language that allows to express business logic in a concise way (type inference, pattern matching, case classes). It greatly reduces NPE count and makes process of consuming any Java API much more convenient (this is important cause in the heart of Zally lies swagger-parser which is very java-ish - with a lot of unexpected nulls).
- Designed with java interop in mind. Unlike Scala Kotlin doesn't have it's own collections (it uses plain old java ones). This opens space for great compatibility: Spring boot works with Kotlin like a charm (I have 6 month experience with Kotlin + Spring boot). Also as of now Kotlin is in **assess** on [Zalando Tech Radar](https://zalando.github.io/tech-radar/) 
- Easy to grasp for java-developers (much easier than Scala). No mind boggling implicits, no higher-kinded types, no jiberish-special-characters-as-operators. 
- Good IDE support (on-par with java)

**Cons:**
- In some rare corner cases you need to tweak code for interoperability (i needed to make createViolation and isViolation public cause Kotlin doesn't have exact analog of `default` java visibility modifier). Same thing for Spring - sometimes you need to specify additional modifiers/annotations to make it work.

In this PR I just rewrote couple of rules to Kotlin to compare the code. Do not merge it unless @fmueller approves.

Remarkable thing: HttpHeadersRule (written in Kotlin) has 2 child classes (written in java) 

Let's discuss.


ishishov:
You made my day :))
If it's switched to Kotlin, I would be interested to contribute

netme:
Looks a way more cleaner, but here comes a question if it will be efficient to use Kotlin for a team, who will maintain it later. Will leave this question open to @fmueller .

fmueller:
I really like how consice this code is and clean. As Kotlin seems to be far better interoperable with Java than Scala for instance, I don't see an issue in giving it a try. For implementing rules it makes sense, I guess. But there is an open issue in Spring Boot for better Kotlin support https://github.com/spring-projects/spring-boot/issues/5537 So I propose to allow writing rules from now on in Kotlin but stick to Java in the API itself to not run into subtle interoperability issues with Spring. What do you think?

fmueller:
👍

Issue #57, opened 2016-12-23
Fix `gradle bin` for zally-cli


netme:
We have spotted that on some environments `gradle bin` command doesn't create a proper `zally.jar` file inside the `bin/` folder. That makes a `zally` command failing. Investigate and fix it.

PR #72, opened 2016-12-30
#57 Fix `gradle bin` command

netme:
Resolves zalando-incubator/zally#57

🔋 Uses more precise way to copy `zally.jar` file


fmueller:
This is in fact more precise and cleaner. :) LGTM

fmueller:
👍

PR #74, opened 2017-01-02
Add support for insecure https connections

### LINKED ISSUE #55: HTTPS connections are not working
Check why `unirest` is failing making HTTPS requests and fix it.

fmueller:
Closes #55 

Since handling different ssl keystores with java is quite heavy burden for cli tool users and we don't handle customer critical traffic, this PR configure http client of CLI with disabled ssl check.

You can verify this locally by using https://zally.architecture.zalan.do


netme:
Thanks for fixing!

netme:
👍

PR #76, opened 2017-01-03
zally-61 Auto detection of rules present in classpath.

thomasabraham:
- All rules marked with @Component annotation
- RulesValidator is not a rule anymore
- Removed RulesValidatorConfiguration


fmueller:
LGTM :-)

PR #84, opened 2017-01-04
ZALLY-77 Fix bin path in zally command

### LINKED ISSUE #77: Make `zally` command path-agnostic
### Steps to reproduce
- `cd bin/`
- `./zally path/to/swagger.yaml`

### Actual result
Command reports that `bin/zally.jar` is not found.

### Expected result
Command should still work

netme:
🔋 

Only shell script update. 

Closes #77 

More details: http://stackoverflow.com/questions/59895/getting-the-current-present-working-directory-of-a-bash-script-from-within-the-s

### DIFF
--- a/cli/bin/zally
+++ b/cli/bin/zally
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
 function zally {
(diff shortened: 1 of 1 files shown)


mkulak:
:+1:

PR #91, opened 2017-01-09
#86 Write one-jar directly to /bin

### LINKED ISSUE #86: Instructions "how to build and run cli tool" are incomplete
In [cli/README.md] I see following instructions:
1. cd cli
2. ./gradlew bin
3. ./bin/zally

After doing it on a clean project I get: 
`Error: Unable to access jarfile /Users/mkulak/src/zally/cli/bin/zally.jar`

As far as I understand shell script expects jar-file to be located in his dir but gradle by default places it in build/libs.
So we need either add to readme additional step like `copy build/libs/zally.jar to bin` or add copying to gradle task.

netme:
Closes #86 

🔋 

Since the chain of build -> oneJar -> bin provides a lot of troubles,
let's minimise the efforts and use `destinationDir` parameter of
Jar plugin:

https://docs.gradle.org/current/dsl/org.gradle.api.tasks.bundling.Jar.html#org.gradle.api.tasks.bundling.Jar:destinationDir

### DIFF
--- a/cli/build.gradle
+++ b/cli/build.gradle
@@ -33,16 +33,9 @@ task wrapper(type: Wrapper) {
     gradleVersion = '2.14.1'
 }
 
-task zallyJar(type: OneJar) {
+task bin(type: OneJar) {
     dependsOn build
     mainClass = 'de.zalando.zally.cli.Main'
     archiveName = 'zally.jar'
-}
-
-task bin {
-    dependsOn zallyJar
-    copy {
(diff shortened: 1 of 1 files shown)


fmueller:
👍 

Hope this solves our weird issues we have sometimes with it. ;-)

fmueller:
👍

netme:
@fmueller, me too. I have a feeling that it's a race condition and removing 1 step might help. Let's see.

PR #93, opened 2017-01-10
Switching to shadow plugin

maxim-tschumak:
I've faced 2 problems when building zally-cli:
1. Empty bin/zally.jar (no manifest, no classes).
2. Stackoverflow when playing with the configuration of gradle-one-jar to fix 1.

Because of this, I propose to switch to `shadow` plugin: it works, has better docu and bigger community.
I've never used Gradle before, so let me know if you have a better solution :)

PS. The first issue is maybe related to https://github.com/rholder/gradle-one-jar/issues/21


fmueller:
Let's cross fingers and hope issues are resolved.

LGTM

fmueller:
👍

netme:
Yeah, it was time to replace it. Already tried to fix issues 3 times.

👍

PR #102, opened 2017-01-11
ZALLY-92 Add Initial Integration Tests Suite

### LINKED ISSUE #92: Add an Integration Test to Run on Travis.CI
In order to make sure that CLI tool is not broken, we can add a small integration test which:

1. Compiles and starts Zally Server
2. Compiles Zally CLI
3. Runs few scenarios with local and remote files

In order to keep it simple, we can use bats framework for that.

netme:
Resolves #92 

- [x] Adds simple `.bats` script to run CLI tool agains locally running server. 
- [x] Simplifies usage of `shadowJar`

More info: https://github.com/sstephenson/bats

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,7 @@ install: true
 script:
   # Unit-test and build CLI tool
   - cd cli/
-  - ./gradlew test --info
-  - ./gradlew shadowJar
+  - ./gradlew build --info
   - cd ..
 
   # Unit-test server
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,37 @@ language: java
 jdk:
   - oraclejdk8
 
+before_install:
+  - sudo add-apt-repository ppa:duggan/bats --yes
+  - sudo apt-get update -qq
+  - sudo apt-get install -qq bats
+
 install: true
+
 script:
+  # Unit-test and build CLI tool
   - cd cli/
-  - gradle test --info
-  - cd ../server/
-  - gradle test --info
+  - ./gradlew test --info
+  - ./gradlew shadowJar
+  - cd ..
+
+  # Unit-test server
+  - cd server/
+  - ./gradlew test --info
+  - cd ..
+
+  # Launch local zally server
+  - cd server/
+  - echo "spring.profiles.active=dev" > application.properties
+  - ./gradlew bootRun > /dev/null &
+  - echo $! > /tmp/zally_server.pid
+  - cd ..
+
+  # Wait until Spring Boot will start
+  - sleep 15
+
+  # Run CLI integration test
+  - ./cli/integration_tests/tests.bats
+
+  # Kill Zally server instace
+  - kill -9 `cat /tmp/zally_server.pid`
--- a/cli/README.md
+++ b/cli/README.md
@@ -14,7 +14,7 @@ requesting violations check at a given Zally server.
 
 3. Build `zally.jar` package:
 	```bash
-	./gradlew shadowJar
+	./gradlew build
 	```
 
 4. Run it using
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -5,9 +5,37 @@ language: java
...
+
+  # Launch local zally server
+  - cd server/
+  - echo "spring.profiles.active=dev" > application.properties
+  - ./gradlew bootRun > /dev/null &
+  - echo $! > /tmp/zally_server.pid
+  - cd ..
+
+  # Wait until Spring Boot will start
+  - sleep 15
```
netme: We have to wait until https://github.com/spring-projects/spring-boot/issues/3982 will be resolved :-)

.travis.yml
```diff
@@ -5,9 +5,37 @@ language: java
...
 install: true
+
 script:
+  # Unit-test and build CLI tool
   - cd cli/
-  - gradle test --info
-  - cd ../server/
-  - gradle test --info
+  - ./gradlew test --info
+  - ./gradlew assemble
```
fmueller: I propose to use `./gradlew build` instead of test and assemble afterwards. build is executing all tests and builds the jar file.

cli/README.md
```diff
@@ -14,7 +14,7 @@ requesting violations check at a given Zally server.
 
 3. Build `zally.jar` package:
 	```bash
-	./gradlew shadowJar
+	./gradlew assemble
```
fmueller: As comment above: build instead of assemble.

cli/build.gradle
```diff
@@ -45,3 +45,5 @@ shadowJar {
 task wrapper(type: Wrapper) {
     gradleVersion = '2.14.1'
 }
+
+assemble.dependsOn shadowJar
```
fmueller: I don't know if this is still necessary if we stick to build instead of assemble.

.travis.yml
netme: @fmueller thanks for your comment, I will investigate a bit more on this topic. Also was not sure if `assemble` is the best option.

cli/build.gradle
netme: `shadowJar` is not triggered by `build`. That's why I've started doing that changes. I will try to find a better solution.

fmueller: Adding shadowJar step to `build`, maybe?

netme: Yeah, most probably it will work. I'm still quite new to `gradle`, sorry :-)

fmueller: Don't wanted to sound rude. I like your contributions a lot. ;-)

netme: You are not sounding rude, I really appreciate your feedback :-)

netme:
@fmueller I've switched `assemble` to `build` in 1357455 .

.travis.yml
```diff
@@ -5,9 +5,37 @@ language: java
...
+
 install: true
+
 script:
+  # Unit-test and build CLI tool
   - cd cli/
-  - gradle test --info
-  - cd ../server/
-  - gradle test --info
+  - ./gradlew test --info
```
fmueller: With `build` this test goal should also be executed. I think, we can remove `gradle test` here.

netme: Good point, let's modify the travis script and check it :-)

netme: @fmueller Fixed

netme: And many thanks for your comments, I really learned a lot about Gradle these days 👍

maxim-tschumak:
:+1: from my side. Bats looks very good! I would go for fewer  integration tests (one happy case, one not-very-happy case). It's easier to maintain and other corner cases should be tested on unit level, IMO.

netme:
@maxim-tschumak Thanks for checking! I agree, the reason why we add 5 of them is to cover all 4 happy-paths and one corner case. We found in the past that it's easy to break one of the possible flows in CLI tool. Later, if you don't mind, I would like to add the tests for another service URL and token. But let's discuss it in person. 

@fmueller, your approval is still missing :-)

PR #110, opened 2017-01-18
ZALLY-73 Show Help if Parameters are Missing

### LINKED ISSUE #73: Help instructions should show up if arguments were not sufficient in any form
At the moment, CLI only shows help instructions if '--help' parameter is supplied. I think it's a better user experience if these help instructions always show up when there is an issue with supplied argument list - except if we can give better error description.

netme:
Closes #73 

📗

### DIFF
--- a/cli/src/main/java/de/zalando/zally/cli/Main.java
+++ b/cli/src/main/java/de/zalando/zally/cli/Main.java
@@ -23,8 +23,10 @@ public class Main {
     @Option(opt = {"-t", "--token"}, description = "OAuth2 Security Token")
     private String token;
 
+    private static OptionParser parser;
+
     public static void main(String[] args) {
-        OptionParser parser = new OptionParser(Main.class);
+        parser = new OptionParser(Main.class);
         try {
             parser.parse(args);
         } catch (IllegalArgumentException e) {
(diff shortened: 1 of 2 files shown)


netme:
@maxim-tschumak could you please also post a comment with 👍 ? Zappr doesn't understand GH approvals yet...

maxim-tschumak:
:+1:

cli/src/main/java/de/zalando/zally/cli/Main.java
```diff
@@ -45,7 +47,9 @@ void run(String[] args) {
 
     private boolean lint(String[] args) throws RuntimeException {
         if (args.length < 1) {
-            throw new RuntimeException("Please provide a swagger file path or URL");
+            System.err.println("Please provide a swagger file path or URL\n");
+            parser.showHelp();
+            throw new RuntimeException("");
```
fmueller: If you don't have a message for the exception you can also use empty constructor parameter list: `new RuntimeException()`

netme: @fmueller, true, I've already tried it, but the library will print `(null)` to the console in this case. That's why I've decided to use this hack :-)

fmueller: Ah ok, makes sense :)

PR #111, opened 2017-01-19
Disable Avoid Javascript Keywords Rule

### LINKED ISSUE #109: Disable Avoid Javascript Keywords rule
Since our guidelines changed (https://github.com/zalando/restful-api-guidelines/pull/181) we should disable this rule for now. Can be made available for other organisations once we have configuration of rules (see #75)

fmueller:
Fixes #109

### DIFF
--- a/server/src/main/java/de/zalando/zally/rules/AvoidJavascriptKeywordsRule.kt
+++ b/server/src/main/java/de/zalando/zally/rules/AvoidJavascriptKeywordsRule.kt
@@ -2,18 +2,20 @@ package de.zalando.zally.rules
 
 import de.zalando.zally.Violation
 import de.zalando.zally.ViolationType
-import io.swagger.models.Model
 import io.swagger.models.Swagger
-import org.springframework.stereotype.Component
 
-import java.util.*
-
-@Component
+/**
+ * TODO:
+ *      - this rule is currently disabled because Zalando's API guidelines changed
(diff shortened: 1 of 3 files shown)


server/src/main/java/de/zalando/zally/rules/AvoidJavascriptKeywordsRule.kt
```diff
@@ -2,18 +2,20 @@ package de.zalando.zally.rules
...
-import org.springframework.stereotype.Component
 
-import java.util.*
-
-@Component
+/**
+ * TODO:
+ *      - this rule is currently disabled because Zalando's API guidelines changed
+ *      - will be re-enabled if we can configure ruleset
+ */
```
netme: I like this approach.

server/src/main/java/de/zalando/zally/rules/AvoidJavascriptKeywordsRule.kt
```diff
@@ -31,4 +33,4 @@ class AvoidJavascriptKeywordsRule : Rule {
 
     private fun createViolation(path: String, fieldName: String) =
             Violation(TITLE, DESC_PATTERN.format(fieldName), ViolationType.SHOULD, RULE_URL, path + "." + fieldName)
-}
+}
```
netme: Just a random thought: we should introduce some code quality tool at some point

netme:
👍

PR #116, opened 2017-01-23
Add architecture overview for planned github integration

### LINKED ISSUE #83: Define GH Integration Architecture
Since we have an already working GH integration prototype, it's time to convert it to a service. Before we create tasks for that, we need to define an architecture.

fmueller:
Fixes #83


gh-integration/README.md
```diff
@@ -0,0 +1,18 @@
...
+## Architecture Proposal
+
+                                                      +--------------+
+                      +------------------------------>+ Zally Web UI |
+                      |                               +-----+--------+
+                      |                                     |
+                      |                                     ^
+           +----------+------------+                  +-----+--------+
+           | GH Integration Server +-----------------^+ Zally Server |
+           +-----------------------+                  +-----+--------+
```
netme: Why do we have a loop here? Maybe `Zally Web UI` should operate directly with `GH Integration Server`?

netme:
Restarted builds after #118 got merged

gh-integration/README.md
fmueller: This is kinda misleading because there is no arrow-down in ascii. Zally Web UI connects to Zally Server, so no loop in there.

netme: Ah, OK, got it, then I'm fine :-)

netme:
👍

netme:
👍

Issue #113, opened 2017-01-19
Introduce Code Quality Tools to the Project


netme:
In order to keep the codebase clean, we should introduce some code quality checking tools like PMD or CheckStyle for Java.

Since we also supporting Kotlin for the rule definitions, we also need to check tooling around this language.

maxim-tschumak:
Error-prone is also a good option. (http://errorprone.info/)

netme:
@maxim-tschumak , yeah, looks like a good addition to other tools. We can use multiple of them.

mkulak:
We could look at: https://github.com/shyiko/ktlint

netme:
@mkulak thanks for the link. It looks good to me, we can give it a try.

netme:
⬆️  first part, more to come.

fmueller:
Because Kotlin as a language does not have so good integration into SaaS code quality tools, it's ok to have it integrated into the build.

For Java we can do the same with Findbugs and PMD. But I also like Codacy (https://www.codacy.com) because they are integrating smoothly with our PR-flow. What do you think?

netme:
@fmueller thanks for your input.

Codacy looks nice, I really like how it's integrated with our pull requests, we can benefit from it.

From another hand, if we will go with PMD + FindBugs (and maybe + CheckStyle), we can use them locally during the development process, this will also protect us from pushing lower quality code. 

We can also use errorprone instead of Findbugs because it operates on source instead of bytecode.

Thoughts?

netme:
Since this task is becoming quite big, opened 2 extra tickets #123 and #124 with leftovers. Closing this one.

PR #118, opened 2017-01-23
ZALLY-117 Wait for Spring Boot to Start

netme:
Related to #117 

🔋 

Hopefully resolves an issue with race condition. 

#### Possible solution:
Poll port 8080 instead of waiting for 15 seconds

PR #119, opened 2017-01-24
ZALLY-113 Introduce Checkstyle in CLI subproject

netme:
Related to #113 

🔋 

- Introduces Checkstyle
- Fixes issues found by Checkstyle 

Rules adjustments from dfbf1e2 to be discussed with @fmueller and @maxim-tschumak 

### How to test:
- Run `./gradlew check`
- Observe that only `Javadoc` issues left

### DIFF
--- a/cli/build.gradle
+++ b/cli/build.gradle
@@ -1,4 +1,5 @@
 apply plugin: 'java'
+apply plugin: 'checkstyle'
 apply plugin: 'com.github.johnrengelman.shadow'
 
 repositories {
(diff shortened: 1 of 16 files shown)


netme:
Here is an example output on Travis: https://travis-ci.org/zalando-incubator/zally/builds/194976852#L323

netme:
Restarted builds after #118 got merged

fmueller:
LGTM

We should align the ruleset to use afterwards in the next days, @netme

fmueller:
👍

netme:
@fmueller yes, makes complete sense. I wanted to open a discussion here. Right now it's a google rule set with 2 exceptions:

- Line length: 120 chars (to be aligned with IDEA defaults)
- Tab length: 4 chars (was before 2)

netme:
@fmueller and to be 100% clear, Google standard requires docstrings. There are about 10 warnings now due to that. We should also agree if we are using them or not.

PR #114, opened 2017-01-20
#56 Improve Error Reporting

netme:
Closes #56 

Side-effect: removes couple of issues fro Codacy

PR #129, opened 2017-02-01
#113 Fix small issues captured by Codacy

netme:
Related to #113 

Also fixes a bug with `ResultPrinter` summary. 

Still left: Exception handling, will be addressed in #114


fmueller:
👍

PR #133, opened 2017-02-06
#123 Fix the Rest of Codacy Issues

### LINKED ISSUE #123: Fix Issues Spotted By Checkstyle
Related to #113 

Fix the most critical issues in `zally-server` project.

To get the list of issues, run:

```bash
cd server/
gradle check
```

netme:
Closes #123


server/src/test/java/de/zalando/zally/rules/PluralizeNamesForArraysRuleTest.java
```diff
@@ -36,27 +41,24 @@ public void negativeCase() {
...
             File file = ResourceUtils.getFile("src/test/resources/" + fixture);
             return new SwaggerParser().parse(new String(Files.readAllBytes(file.toPath())));
-        } catch (FileNotFoundException e) {
-            e.printStackTrace();
-            throw new RuntimeException("Unable to read fixture", e);
         } catch (IOException e) {
             e.printStackTrace();
-            throw new RuntimeException("Unable to read fixture", e);
+            throw e;
         }
```
mkulak: Why we are catching IOException here and then re-throwing it?
I propose to remove try-catch entirely. 

Or even better: leave wrapping IOException in RuntimeException as is. I don't see much value in forcing all test methods to declare `throws Exception` instead of just leaving it as is.

BTW that's why I'm not very enthusiastic about code linters: sometimes it forces you to do suboptimal things.

netme: @mkulak , I think we can remove the `try...catch` block here, it would be good enough for jUnit. Will do it.

fmueller:
What about just merging this and improve code quality step by step with each PR?

netme:
@fmueller , I'm up. Please approve it then :-)

fmueller:
👍

PR #146, opened 2017-02-09
#139 rule implementation + test

### LINKED ISSUE #139: Rule: Detect base path property is not used
Introduce a new rule which detects if a base path can be extracted out of all resource names. If all paths have same path prefix, a could violation should be created.

mkulak:
Implements server part #139

### DIFF
--- a/server/src/main/java/de/zalando/zally/ViolationType.java
+++ b/server/src/main/java/de/zalando/zally/ViolationType.java
@@ -4,5 +4,6 @@ public enum ViolationType {
 
     MUST,
     SHOULD,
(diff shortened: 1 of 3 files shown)


fmueller:
Great :)

How would the CLI tool behave if it receives a new rule type? @netme

netme:
@fmueller, it will ignore it most probably. I will adjust it later, please don't close the ticket :-)

netme:
👍 Since tests are passed let's merge the server part. I will self-assign #139 and implement the CLI part of it.

PR #148, opened 2017-02-10
#96 white listed _link, fixed typos and kotlinize

### LINKED ISSUE #96: _links field is not whitelisted
_links field should be whitelisted in each rule, it violates our syntax rules in the guidelines but is also defined as default field for links in our guidelines. Currently Zally creates violations for it.

mkulak:
Closes #96


mkulak:
Summary: add whitelist, fix multiple typos, kotlinize

fmueller:
👍

PR #149, opened 2017-02-14
#140 Give Information of Non-Automated Rules

### LINKED ISSUE #140: Give information about not-automated rules
As user of Zalando's API Linter I'd like to know which rules are not yet automated. This would improve API design process because as an API provider I would know which rules I should read. Furthermore it would help API reviewers to know on which aspects to concentrate because they are not covered by Zally.

Server should deliver in api-violations response a simple description text about not yet automated rules. This text should be configurable. CLI should display this text as part of summary.

netme:
Closes #140

### DIFF
--- a/cli/src/main/java/de/zalando/zally/cli/ResultPrinter.java
+++ b/cli/src/main/java/de/zalando/zally/cli/ResultPrinter.java
@@ -32,6 +32,14 @@ public class ResultPrinter {
         writer = new OutputStreamWriter(outputStream);
     }
 
+    public void printMessage(String message) throws IOException {
+        if (!message.isEmpty()) {
+            printHeader(ANSI_WHITE, "Server message");
+            writer.write(message);
+            writer.flush();
+        }
+    }
+
     public void printViolations(List<JsonObject> violations, String violationType) throws IOException {
         if (!violations.isEmpty()) {
 
--- a/cli/src/main/java/de/zalando/zally/cli/ResultPrinter.java
+++ b/cli/src/main/java/de/zalando/zally/cli/ResultPrinter.java
@@ -35,7 +35,7 @@ public class ResultPrinter {
     public void printMessage(String message) throws IOException {
         if (!message.isEmpty()) {
             printHeader(ANSI_CYAN, "Server message");
-            writer.write(message);
+            writer.write(message + "\n\n");
             writer.flush();
         }
     }
--- a/cli/src/main/java/de/zalando/zally/cli/ResultPrinter.java
+++ b/cli/src/main/java/de/zalando/zally/cli/ResultPrinter.java
@@ -34,7 +34,7 @@ public class ResultPrinter {
 
     public void printMessage(String message) throws IOException {
         if (!message.isEmpty()) {
-            printHeader(ANSI_WHITE, "Server message");
+            printHeader(ANSI_CYAN, "Server message");
             writer.write(message);
             writer.flush();
         }
@@ -59,7 +59,7 @@ public class ResultPrinter {
     }
 
     public void printSummary(List<String> violationTypeNames) throws IOException {
-        printHeader(ANSI_WHITE, "Summary:");
+        printHeader(ANSI_CYAN, "Summary:");
         for (String name : violationTypeNames) {
             writer.write(name.toUpperCase() + " violations: " + counters.getOrDefault(name, 0).toString() + "\n");
         }
@@ -108,7 +108,7 @@ public class ResultPrinter {
             case "HINT":
                 return ANSI_CYAN;
             default:
-                return ANSI_WHITE;
+                return ANSI_CYAN;
         }
     }
 }
--- a/cli/src/test/java/de/zalando/zally/cli/ResultPrinterTest.java
+++ b/cli/src/test/java/de/zalando/zally/cli/ResultPrinterTest.java
@@ -126,4 +126,20 @@ public class ResultPrinterTest {
         String result = ResultPrinter.formatViolation(testColor, violation);
         assertEquals(expectedResult, result);
     }
+
+    @Test
+    public void printsProperMessage() throws Exception {
+        final String message = "Test message";
+        ResultPrinter resultPrinter = new ResultPrinter(outStream);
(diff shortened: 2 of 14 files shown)


server/src/main/resources/api/zally-api.yaml
```diff
@@ -72,6 +72,8 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/Violation'
+      message:
+        type: string
```
netme: @fmueller , I will need your quick feedback on API change

server/src/main/java/de/zalando/zally/ApiViolationsController.java
```diff
@@ -19,6 +20,9 @@
     private final RulesValidator rulesValidator;
     private final ObjectMapper mapper;
 
+    @Value("${zally.message}")
+    private String message;
```
fmueller: Can you transform this into constructor injection like the other parameters? @Value does also work on constructor argument.

server/src/main/resources/api/zally-api.yaml
fmueller: LGTM

A simple string field can be used for any type of message. I like it.

fmueller: It should be optional btw. So, server can response without it if value is not specified or empty.

server/src/test/java/de/zalando/zally/RestApiTest.java
```diff
@@ -80,6 +80,9 @@ public void shouldValidateGivenApiDefinition() throws IOException {
         JsonNode violations = rootObject.get("violations");
         assertThat(violations).hasSize(1);
         assertThat(violations.get(0).get("title").asText()).isEqualTo("dummy1");
+
+        String message = rootObject.get("message").asText();
+        assertThat(message).isEqualTo("Test message");
```
fmueller: Can you also test message not specified in configuration?

server/src/main/resources/api/zally-api.yaml
netme: 👍 Makes sense, will add it to spec

server/src/main/java/de/zalando/zally/ApiViolationsController.java
netme: I will take a look on it. Still a bit new to Spring.

server/src/test/java/de/zalando/zally/RestApiTest.java
netme: I will take a look on it

server/src/main/java/de/zalando/zally/ApiViolationsController.java
fmueller: No worries. :) As a rule in Java: always favor constructor based injection. It makes testing your code easier. If you only have constructor based injection, you don't need to start spring application context in your test and can simple mock your dependencies.

netme: Oh, thanks for explanation. I'm sure I will do the same trick in any other language, makes complete sense :-)

server/src/test/java/de/zalando/zally/RestApiTest.java
netme: @fmueller , I believe I did a small improvement in this direction here: https://github.com/zalando-incubator/zally/pull/149/commits/b52b452ad22dfbdd3a6af0adba9ae74e0216155b#diff-681343b113913a3272e8cd1141af063bR38

But most probably it's still not that you expected. Should I create another `SpringBootTest` for that?

server/src/main/resources/api/zally-api.yaml
netme: Fixed

server/src/test/java/de/zalando/zally/RestApiTest.java
fmueller: My point is, you are only testing for "property has certain value and this value is returned in message field in json payload". What happens if this value is not given in the configuration (another test case) and what happens if the value is empty (also another test case).

netme: OK, I've got your point, will try to cover it.

cli/src/main/java/de/zalando/zally/cli/ResultPrinter.java
```diff
@@ -32,6 +32,14 @@ public ResultPrinter(OutputStream outputStream) {
         writer = new OutputStreamWriter(outputStream);
     }
 
+    public void printMessage(String message) throws IOException {
+        if (!message.isEmpty()) {
+            printHeader(ANSI_WHITE, "Server message");
```
fmueller: Is ANSI_WHITE actually readable in non-dark terminals?

netme: Good point, I need to try that

fmueller:
👍

netme:
@fmueller resolving conflicts now :-)

fmueller:
@netme I see, good luck. ^^ Will give you thumbs up afterwards :)

netme:
Seems to be ready, let's see if Travis / Codacy will agree :-)

fmueller:
👍

cli/src/test/java/de/zalando/zally/cli/ResultPrinterTest.java
```diff
@@ -75,7 +75,7 @@ public void printsProperSummary() throws IOException {
         outContent.reset();
 
         resultPrinter.printSummary(Linter.violationTypes);
-        String expectedResult = resultPrinter.ANSI_WHITE + "\nSummary:\n"
+        String expectedResult = resultPrinter.ANSI_CYAN + "\nSummary:\n"
```
netme: I also found out that we were using `ANSI_WHITE` for `Summary:` section and it was looking ugly on the white background, so good spot, @fmueller :-)

fmueller: Nice finding :)

Issue #51, opened 2016-12-22
Response enhancement or more detailed report page


superbenji:
Include count of must, should and could violations in the response in order to quickly get an overview in which condition the API spec is.
Or let the frontend (Zally report) calculate this numbers and display them.

netme:
I've counted the amount of violations in CLI tool, see #52 .

Would be nice to get these counters directly from API to keep less business logic in clients.

fmueller:
+1

I also see it as a nice addition to our API.

netme:
Server part is implemented in #158, will fix CLI in another PR

PR #158, opened 2017-02-28
#51 Add Statistics to API

netme:
Related to #51

Adds API changes proposal.

### DIFF
--- a/server/src/main/java/de/zalando/zally/ViolationsCounter.java
+++ b/server/src/main/java/de/zalando/zally/ViolationsCounter.java
@@ -1,0 +1,27 @@
+package de.zalando.zally;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class ViolationsCounter {
+    private final Map<ViolationType, Integer> counters = new HashMap<>();
+
+    public ViolationsCounter(List<Violation> violations) {
+        countViolations(violations);
+    }
+
+    public Integer getCounter(ViolationType violationType) {
+        return counters.getOrDefault(violationType, 0);
+    }
+
+    private void countViolations(List<Violation> violations) {
+        violations.forEach(violation -> updateCounter(violation));
+    }
+
+    private void updateCounter(Violation violation) {
+        ViolationType violationType = violation.getViolationType();
+        Integer count = counters.getOrDefault(violationType, 0) + 1;
+        counters.put(violationType, count);
+    }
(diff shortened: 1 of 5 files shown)


server/src/main/resources/api/zally-api.yaml
```diff
@@ -87,6 +87,8 @@ definitions:
           $ref: '#/definitions/Violation'
       message:
         type: string
+      statistics:
```
fmueller: I personally find it a bit redundant to have "_violations" in each field name below and propose to change `statistics` field to `violations_count` and rename field below to `must`, `should` and so on.

netme:
@fmueller , swagger definition is updated.

server/src/main/java/de/zalando/zally/ViolationsCounter.java
```diff
@@ -0,0 +1,27 @@
...
+    public ViolationsCounter(List<Violation> violations) {
+        countViolations(violations);
+    }
+
+    public Integer getCounter(ViolationType violationType) {
+        return counters.getOrDefault(violationType, 0);
+    }
+
+    private void countViolations(List<Violation> violations) {
+        violations.forEach(violation -> updateCounter(violation));
```
fmueller: You can also do some sort of grouping by with Java streams api. I think this would simplify your code.

netme: Yeah, also thought about it, will try to find a good solution

server/src/main/java/de/zalando/zally/ViolationsCounter.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class ViolationsCounter {
+    private final Map<ViolationType, Integer> counters;
+
+    public ViolationsCounter(List<Violation> violations) {
+        counters = violations.
+                stream().
+                collect(Collectors.groupingBy(Violation::getViolationType, Collectors.summingInt(x -> 1)));
```
netme: Using `summingInt()` instead of `counting()` here to get integer counters

netme:
@fmueller , the server part is ready, if you don't mind I will fix CLI in a separate PR.

fmueller:
Ok, you decide how to develop this. Go ahead :-)

fmueller:
👍

PR #164, opened 2017-03-10
#70 Use only 1 JSON library in CLI

### LINKED ISSUE #70: Use only 1 JSON Library in Zally CLI
Currently we are using 2 JSON libraries in CLI tool. This leads to some inefficient data format transformations like:

https://github.com/zalando-incubator/zally/blob/master/cli/src/main/java/de/zalando/zally/cli/YamlReader.java#L22

Goal: Minimise amount of JSON libraries to 1

netme:
Closes #70 

### TODO:
- [x] Return `ZallyApiResponse` directly from `client.validate()`
- [x] Replace `minimaljson` with `org.json`
- [x] Add directory structure
- [x] Bump JSON dependencies
- [x] Add more tests to newly added classes
- [x] Make `violation_type` non-optional
- [ ] #189 ~~Introduce timeouts to HTTP client~~
- [ ] #188 ~~Do not pass `token` when not specified~~


netme:
@fmueller sorry for a big mess created, but could you please review?

cli/src/main/java/de/zalando/zally/cli/api/ZallyApiClient.java
```diff
@@ -46,25 +48,8 @@ public ZallyApiClient(String url, String token) {
...
-            String details = exception.getCause() == null ? exception.getMessage() : exception.getCause().getMessage();
-            throw new CliException(
-                    CliExceptionType.API,
-                    "An error occurred while querying Zally server",
-                    details
-            );
-        }
-
+    public ZallyApiResponse validate(String requestBody) throws CliException {
+        final HttpResponse<String> response = doRequest(requestBody);
```
fmueller: Minor: I think there are better names for `doRequest`, e.g. `requestViolationsReport`.

cli/src/main/java/de/zalando/zally/cli/api/ZallyApiClient.java
```diff
@@ -76,21 +61,47 @@ public JsonValue validate(String body) throws CliException {
...
-        if (!errorJson.asObject().isEmpty()) {
-            return errorJson.asObject().getString("detail", null);
+        return errorJson.optString("detail", null);
+    }
+
+    private HttpResponse<String> doRequest(final String requestBody) throws CliException {
+        try {
+            return Unirest
+                    .post(url)
+                    .header("Authorization", "Bearer " + token)
```
fmueller: This does not like OAuth is optional. Do I miss something?

cli/src/main/java/de/zalando/zally/cli/api/ZallyApiClient.java
```diff
@@ -76,21 +61,47 @@ public JsonValue validate(String body) throws CliException {
...
         }
 
-        if (!errorJson.asObject().isEmpty()) {
-            return errorJson.asObject().getString("detail", null);
+        return errorJson.optString("detail", null);
+    }
+
+    private HttpResponse<String> doRequest(final String requestBody) throws CliException {
+        try {
+            return Unirest
```
fmueller: Btw: Do we set a timeout for requests to Zally server? If not, we should for better user experience.

cli/src/main/java/de/zalando/zally/cli/readers/SpecsReaderFactory.java
```diff
@@ -1,7 +1,10 @@
-package de.zalando.zally.cli;
+package de.zalando.zally.cli.readers;
```
fmueller: I'd strongly recommend to stick to singular in package names. It's obvious that a package is grouping multiple things together and it's kinda default for naming packages what I have seen so far.

cli/src/main/java/de/zalando/zally/cli/api/ZallyApiClient.java
netme: Cool, will fix it.

cli/src/main/java/de/zalando/zally/cli/api/ZallyApiClient.java
netme: Good point , also not sure, will check it

cli/src/main/java/de/zalando/zally/cli/readers/SpecsReaderFactory.java
netme: Will rename it

cli/src/main/java/de/zalando/zally/cli/domain/Violation.java
```diff
@@ -0,0 +1,80 @@
...
+        this.description = description;
+    }
+
+    public Violation(JSONObject violationJson) {
+        this.title = violationJson.getString("title");
+        this.description = violationJson.getString("description");
+
+        if (violationJson.has("violation_type")) {
+            this.violationType = violationJson.getString("violation_type");
+        }
```
mkulak: Isn't `violation_type` mandatory? (as `title` and `description` are)

mkulak: Hm, according to spec - no. Strange (

cli/src/main/java/de/zalando/zally/cli/ViolationsFilter.java
```diff
@@ -1,33 +1,22 @@
...
 
-import com.eclipsesource.json.JsonObject;
-import com.eclipsesource.json.JsonValue;
+import de.zalando.zally.cli.domain.Violation;
 
 import java.util.List;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 public class ViolationsFilter {
```
mkulak: I think this class can be inlined.

cli/src/main/java/de/zalando/zally/cli/domain/Violation.java
netme: Actually no according to our Swagger definition:

```yaml
  Violation:
    type: object
    description: Violation Object
    required:
      - title
      - description
      ...
```

Maybe it's time to rethink it

cli/src/main/java/de/zalando/zally/cli/readers/SpecsReaderFactory.java
netme: Addressed in 03bb6e3

cli/src/main/java/de/zalando/zally/cli/api/ZallyApiClient.java
netme: Addressed in 10909f7

cli/src/main/java/de/zalando/zally/cli/api/ZallyApiClient.java
netme: Addressed in 10909f7

cli/src/main/java/de/zalando/zally/cli/api/ZallyApiClient.java
netme: @fmueller, could you please elaborate? Probably it's a bit late and my brain is off  :-D

cli/src/main/java/de/zalando/zally/cli/ViolationsFilter.java
netme: Actually you are right, it can be. Just kept it due to the good test coverage around. Should I do it?

cli/src/main/java/de/zalando/zally/cli/domain/Violation.java
fmueller: Let's change it to required then, common sense I'd say. ;)

netme: OK, changing :-)

cli/src/main/java/de/zalando/zally/cli/api/ZallyApiClient.java
fmueller: Ah, I see. I forgot to speak with you about it: it is possible to start zally server without OAuth protection in place. So, the OAuth token should also be optional and configurable in the CLI. The header don't have to be set necessarily.

netme:
Oups, custom timeouts killed the build. Need to investigate a bit more.

cli/src/main/java/de/zalando/zally/cli/api/ZallyApiClient.java
netme: @fmueller , got it. So if token is not set, we should not expose the header, right? Will do it.

cli/src/main/java/de/zalando/zally/cli/api/ZallyApiClient.java
netme: Rolled-back the change. Need to investigate how to introduce timeouts in our custom HTTP client.

cli/src/main/java/de/zalando/zally/cli/ViolationsFilter.java
mkulak: I still prefer it to be inlined. But will not push for it anyway ).

cli/src/main/java/de/zalando/zally/cli/domain/Violation.java
netme: Fixed in c79f166
API spec is updated in ffdc622

fmueller:
👍 

As discussed, client base timeout and optional token will be part of separate tickets.

fmueller:
👍

netme:
Tickets are created #188 and #189

PR #165, opened 2017-03-13
Topic/template refactor

rbarilani:
Add and use a base template that can be reused.
Use style library from free CDN.
Adjust detail view to display "violation paths" as a list.


fmueller:
👍

Issue #82, opened 2017-01-03
Group Validation Issues


netme:
Sometimes, developers are using the same wrong pattern in multiple places of swagger file. This leads to the long list of similar violations when running Zally.

Proposal: We need to group violations by rule type.

Question 1: Where it should be implemented? 
- On Zally API level
- In the client libraries

Question 2: How it will affect the output format?

mkulak:
Could you please provide an example of this behaviour.

netme:
@mkulak , valid point! Here is an example:

```text
$ ./zally ../../server/src/test/resources/api_spp.json
Found the following MUST violations
===================================
Use Problem JSON:
	Operations Should Return Problem JSON When Any Problem Occurs During Processing Whether Caused by Client Or Server
	/product-put-requests/{product_path}/POST/400
Use 429 With Header For Rate Limits:
	If Client Exceed Request Rate, Response Code Must Contain Header Information Providing Further Details to Client
	/product-put-requests/{product_path}/POST/429
Use 429 With Header For Rate Limits:
	If Client Exceed Request Rate, Response Code Must Contain Header Information Providing Further Details to Client
	/products/GET/429
Use 429 With Header For Rate Limits:
	If Client Exceed Request Rate, Response Code Must Contain Header Information Providing Further Details to Client
	/products/{product_id}/children/GET/429
Use 429 With Header For Rate Limits:
	If Client Exceed Request Rate, Response Code Must Contain Header Information Providing Further Details to Client
	/products/{product_id}/GET/429
Use 429 With Header For Rate Limits:
	If Client Exceed Request Rate, Response Code Must Contain Header Information Providing Further Details to Client
	/products/{product_id}/HEAD/429
Use Problem JSON:
	Operations Should Return Problem JSON When Any Problem Occurs During Processing Whether Caused by Client Or Server
	/products/{product_id}/PATCH/400
Use 429 With Header For Rate Limits:
	If Client Exceed Request Rate, Response Code Must Contain Header Information Providing Further Details to Client
	/products/{product_id}/PATCH/429
Use 429 With Header For Rate Limits:
	If Client Exceed Request Rate, Response Code Must Contain Header Information Providing Further Details to Client
	/products/{product_id}/updates/{update_id}/GET/429
Use 429 With Header For Rate Limits:
	If Client Exceed Request Rate, Response Code Must Contain Header Information Providing Further Details to Client
	/request-groups/{request_group_id}/summaries/GET/429
[...]

Issue #122, opened 2017-01-27
Consider using JDK-Standard 'ServiceLoader' instead of Spring


jbellmann:
... for loading rules into validator.

Hi Zally-Team, I would like you to consider using the ['ServiceLoader'-Feature](https://docs.oracle.com/javase/tutorial/ext/basics/spi.html) to load existing rules into the validator. Sure, I don't know what you plan for the future with Zally and how the Rules-interface will evolve in the future. From what I have seen so far, the ServiceLoader provided by the JDK can replace the existing mechanism to load the rules and the validator itself does not depend anymore on spring.

Removing the dependency to spring could make the validator more interesting for other projects. As an example [swagger-codegen-tooling](https://github.com/zalando-stups/swagger-codegen-tooling) comes into my mind. It would be a nice feature to validate the swagger.yaml before generating code.
Also for other integrations (Gradle-plugin) it would make sense to provide a validator with minimal dependencies.

Please have a look into the 'serviceloader'-branch to see how the 'ServiceLoader' can be used.

Regards
Jörg

fmueller:
Wow, thanks for this interesting idea and proof of concept 👍 

You are totally right, without Spring integrations in other tools should be quite easy. For this we could fledge out the rule evaluation in a library and link this library to the server.

How could #75 be implemented with this approach? We'd like to make rule set configurable on runtime. As far as I understood your code, you have to provide the configuration in manifest file in META-INF.

jbellmann:
Regarding #75 I could imagine to give every Rule-Implementation a unique id/shortname (think of modules/checks in [Checkstyle](http://checkstyle.sourceforge.net/checks.html)). Then after loading every rule available in classpath and picked up by ServiceLoader, just filter via standard 'java.util.function.Predicate<Rule>' and return a new RuleValidator-object containing only filtered rules. That would be the easiest thing to try first.

Which rules to include can be specified by a zally.yaml maybe of if nothing specified, every rule kicks in. In zally.yaml a 'include', 'exclude' section can exists to let the user define what makes sense for him/her.

For going with ids/shortnames it would helpful for a user to build then back-links to specific rules in the report (like in the checkstyle-link above). But I think I've seen something like that already, right?

Regards
Jörg

mkulak:
Regarding #75 - I totally agree with @jbellmann, as long as we have full list of `Rule` implementations  we can filter it out by any criteria. 

Regarding ServiceLoader mechanism:
[...]

PR #166, opened 2017-03-13
97 synonyms rule

### LINKED ISSUE #97: Add synonym dictionary as configurable rule
In our microservice architecture at Zalando (and I guess this holds true for other environments as well) we have a common set of property names like customer id, order id and so on. Such terms should be the same across all APIs. I think it would be a great value add if one could configure a synonym dictionary where alternative names for such general property names are listed and once an API uses such a synonym and not the general name a violation is created.

This dictionary should be configurable, for instance by a simple text file.

e.g.:
customer Id -> synonyms: cId, c-id, custId, cust_identifier...

### LINKED ISSUE #155: Add metric for how often each rule was violated
At the moment we only aggregate histograms for violation types. To improve our API guidelines and to gather data in general to improve rule implementation etc. a metric should be introduced for each rule to count how often it was violated.

This should be started after #82 was implemented (as this will have major impact on code structure and we wanna prevent big merge conflicts).

mkulak:
Closes #97  

I also upgraded Kotlin version to 1.1 and made use of new features (I waited for it since #82).

### DIFF
--- a/server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
+++ b/server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
@@ -1,0 +1,35 @@
+package de.zalando.zally.rules
+
+import de.zalando.zally.Violation
+import de.zalando.zally.ViolationType
+import io.swagger.models.Swagger
+
+class AvoidSynonymsRule : AbstractRule() {
+
+    val TITLE = "Use common property names"
+    val DESC_PATTERN = "Property names should utilize common dictionary"
+    val RULE_URL = ""
+
+    private val COMMON_DICTIONARY = listOf(
+            "customer_id" to listOf("customerid", "c_id", "cid", "custid", "cust_id"),
+            "order_id" to listOf("orderid", "o_id", "ord_id"),
+            "payment_id" to listOf("paymentid", "p_id"),
+            "parcel_id" to listOf("parcelid"),
+            "article_id" to listOf("articleid", "a_id", "art_id")
+    )
+
+    override fun validate(swagger: Swagger): Violation? {
+        val dict = COMMON_DICTIONARY.flatMap { (right, wrongList) -> wrongList.map { it to right } }.toMap()
+        val res = swagger.definitions.orEmpty().flatMap { (defName, def) ->
+            val props = def?.properties.orEmpty()
+            props.keys.filter { it in dict }.map { it to "#/definitions/$defName" }
+        }
+        return if (res.isNotEmpty()) {
+            val (names, paths) = res.unzip()
+            val details = names.toSet().groupBy(dict::get)
+                    .map { (right, wrongList) -> right + " instead of " + wrongList.joinToString(", ") }
+                    .joinToString("\n")
+            Violation(this, TITLE, "$DESC_PATTERN:\n$details", ViolationType.SHOULD, RULE_URL, paths.toSet().toList())
+        } else null
+    }
+}
--- a/server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
+++ b/server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
@@ -17,8 +17,7 @@ class AvoidSynonymsRule(@Autowired rulesConfig: Config) : AbstractRule() {
     val descPattern = "Property names should utilize common dictionary"
 
     @Suppress("UNCHECKED_CAST")
-    private val commonDictionary = rulesConfig.getConfig(javaClass.simpleName + ".dictionary")
-            .entrySet()
+    private val commonDictionary = rulesConfig.getConfig("${getName()}.dictionary").entrySet()
             .map { (key, config) -> key to config.unwrapped() as List<String> }
 
     override fun validate(swagger: Swagger): Violation? {
--- a/server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
+++ b/server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
@@ -17,7 +17,7 @@ class AvoidSynonymsRule(@Autowired rulesConfig: Config) : AbstractRule() {
     val descPattern = "Property names should utilize common dictionary"
 
     @Suppress("UNCHECKED_CAST")
-    private val commonDictionary = rulesConfig.getConfig("${getName()}.dictionary").entrySet()
+    private val commonDictionary = rulesConfig.getConfig(javaClass.simpleName + ".dictionary").entrySet()
             .map { (key, config) -> key to config.unwrapped() as List<String> }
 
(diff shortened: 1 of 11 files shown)


fmueller:
Thanks for your efforts and Kotlin evangelism :) 

This PR is way too big. For next time, please distinguish between feature implementation an refactoring.

The rule implementation itself looks good. One more thing which is needed: the dictionary has to be configurable. Also for usage at Zalando.

mkulak:
Sorry for big PR. I'm always hesitant when it comes to creating technical issues like "Refactor X" or "Upgrade language to version Y".  Usually I try to do it along with something that has business value. But yeah, this leads to mixing concerns in one PR. I separated refactoring and rule implementation in 2 different commits with corresponding messages.

Regarding configuration: I agree, synonyms dictionary needs to be configurable. Do we currently have a way to configure it? Is there a rule which I can use as an example?

fmueller:
I understand your concerns with shipping it together with business value adding code. I do it normally same way. Right, you made two different commits. I can live with this in the review. :) 

On more con: we have to constantly re-merge new code from master (because we favor small PRs in Zally).

For synonym dictionary: no, there is no example available. What about linking to custom text file (dictionary config) in our application.conf?

mkulak:
@fmueller the problem is that I need structured config like that:
```
COMMON_DICTIONARY = listOf(
            "customer_id" to listOf("customerid", "c_id", "cid", "custid", "cust_id"),
            "order_id" to listOf("orderid", "o_id", "ord_id"),
            "payment_id" to listOf("paymentid", "p_id"),
            "parcel_id" to listOf("parcelid"),
            "article_id" to listOf("articleid", "a_id", "art_id")
)
```
So simple text file is not the best fit here. I see next possible solutions:
-[] custom YAML file
-[] use [typesafe config](https://github.com/typesafehub/config)

Both are totally doable but none of them is appealing to me.
Can I have your input on that issue? Perhaps you see some other solutions?

netme:
@mkulak , I've rebased this guy

netme:
👍

mkulak:
Ok, I've implemented generic mechanism for rules configuration via typesafe config.
Definitely not the most elegant solution in terms of implementation but it allows to painlessly configure every rule.

Need your feedback @fmueller @netme

netme:
@mkulak , thanks for your contributions, the solution is looking good. If @fmueller doesn't have any other concerns we can merge it.

fmueller:
@mkulak Will review it on Monday - build is broken?

mkulak:
@fmueller ok. 
I'll fix it till Monday

server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
```diff
@@ -0,0 +1,36 @@
...
+    override val violationType = ViolationType.SHOULD
+
+    val descPattern = "Property names should utilize common dictionary"
+
+    @Suppress("UNCHECKED_CAST")
+    private val commonDictionary = rulesConfig.getConfig(javaClass.simpleName + ".dictionary").entrySet()
+            .map { (key, config) -> key to config.unwrapped() as List<String> }
+
+    override fun validate(swagger: Swagger): Violation? {
+        val dict = commonDictionary.flatMap { (canonical, synonyms) -> synonyms.map { it to canonical } }.toMap()
```
fmueller: Minor: please do not use abbreviations for variable names. Having `dict`and `res`here makes it unnecessarily hard to understand what you are doing.

server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
```diff
@@ -0,0 +1,36 @@
...
+        val res = swagger.definitions.orEmpty().flatMap { (defName, def) ->
+            def?.properties.orEmpty().keys.filter { it in dict }.map { it to "#/definitions/$defName" }
```
fmueller: This looks like you are only checking parameter names in definitions sections. Are parameters which are directly defined in path definitions also recognised?

server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
```diff
@@ -0,0 +1,36 @@
...
+class AvoidSynonymsRule(@Autowired rulesConfig: Config) : AbstractRule() {
+
+    override val title = "Use common property names"
+    override val url = ""
+    override val violationType = ViolationType.SHOULD
+
+    val descPattern = "Property names should utilize common dictionary"
+
+    @Suppress("UNCHECKED_CAST")
+    private val commonDictionary = rulesConfig.getConfig(javaClass.simpleName + ".dictionary").entrySet()
```
fmueller: If I read this correctly, we can also add configs for any other rules by this class loading mechanims, right? If so, we should probably align this with https://github.com/zalando-incubator/zally/issues/122 Sorry, I only remember it now that I see your PR in detail.

fmueller: Further question for configuration: is it possible to set the path to rule config per environment parameter? I don't wann rebuild the server and new docker image only because a configuration changed.

mkulak: Not sure I fully understand implications of #122. 
In current implementation rules are configured by putting some keys in `rules-config.conf` and then retrieving those keys from parsed `Config` object. Here I used `javaClass.simpleName` as pure convention. You are free to name your keys whatever you like (and this way some configuration can be easily shared between rules).

mkulak: It's possible for a config key to get it's value from environment variable:
```yaml
#(in a .conf file):
logging-dir: ${HOME}/logs
```
But I'm not sure I follow your thought. Could you give me an example of configuration for some rule that you'd prefer to push into environment env? (Incomplete) list of configuration data  for every rule can be found here: #183

server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
mkulak: I thought some abbreviations are so common that are widely recognisable. Examples:
`in' for `inputStream`, `i` for `index`, `res` for `result`.

But ok, if you find it hard to read, then I would try to write full names.

server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
mkulak: @fmueller 
I spotted this pattern in several (if not all) of our rules that deals with definitions. Tried to create issue but failed to provide valid example of embedded object definition. If you have such example - it would be nice to have it and test all of our rules against it.

mkulak:
Moved path to rules config file to application config (as @fmueller requested).
Ready for review.

fmueller:
👍

PR #178, opened 2017-03-17
170-ui-standalone

### LINKED ISSUE #94: Add optional OAuth2 flow to Web UI
Some environments make an integration with OAuth2 mandatory, therefore it should be optional possible to secure the Web UI (/try and /results endpoints).

Probably use https://github.com/zalando/oauth2-client-js for this.

### LINKED ISSUE #170: UI as standalone component
To increase the flexibility, to increase independence and ease the contributions of Frontend Developers I think it would make sense to have the UI as a separate component using a frontend friendly stack.
The UI will be stateless and communicating with the server only via the exposed REST API.

WDYT?

### LINKED ISSUE #175: Remove web ui related code in server
Once #170 is finished, all web ui related code should be remove from the rest api server codebase.

rbarilani:
Sorry for the big PR but well it's about setting up a completely new project.

close #175 
close #170
close #94 

What still to be done... we are already working on it:

* CI Tasks: Linting and Testing
* Integrate with Travis CI
* Add some cool badges ;-)

P.S. for Zalando's check out a preview here: https://zally-web-ui.overarching.zalan.do/


web-ui/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM registry.opensource.zalan.do/stups/node:latest
```
mfellner: I would use a definitive version since different node versions can easily break stuff. Latest one right now is `7.7-cd45`, see https://registry.opensource.zalan.do/teams/stups/artifacts/node/tags.

web-ui/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+LABEL MAINTAINER "http://zalando.github.io/"
+
+RUN mkdir -p /opt/app
+WORKDIR /opt/app
+
+# Copy Sources
+COPY ./package.json ./package.json
+COPY ./src ./src
+COPY ./scm-source.json ./scm-source.json
```
mfellner: Let's COPY scm-source after `npm install`. Otherwise the image has to be rebuilt even if `src/` or `package.json` don't change.

web-ui/README.md
```diff
@@ -0,0 +1,85 @@
+Zally WEB-UI
+============
+
+Zally Web-UI project provide a web user interface to lint your api specs.
```
mfellner: provide**s** 🇬🇧

web-ui/README.md
```diff
@@ -0,0 +1,85 @@
...
+Proudly build with NodeJS and React.
```
mfellner: buil**t** 🇬🇧

web-ui/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+```bash
+cd web-ui
+npm install --production
+npm start
+```
+
+## Configuration
+
+The web-ui rely on [dotenv](https://github.com/motdotla/dotenv) NodeJS module to handle configuration.
```
mfellner: relie**s** on **the** 🇬🇧

web-ui/README.md
```diff
@@ -0,0 +1,85 @@
...
+npm install
+```
+
+### Run in development mode
+
+```
+npm run dev
+```
+
+> This task start the server in development mode with **nodemon** and **webpack-dev-server-middleware** watching for changes.
```
mfellner: start**s** 🇬🇧

web-ui/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "zally-web-ui",
+  "version": "0.1.0",
+  "description": "Zally API Linter Web UI",
+  "main": "server.js",
```
mfellner: Should be src/server/index.js probably as this file does not exist.

web-ui/package.json
```diff
@@ -0,0 +1,48 @@
+{
```
mfellner: Is this package going to be pushed to a npm registry? Otherwise let's make it `"private": true`

web-ui/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/zalando-incubator/zally.git"
+  },
+  "keywords": [
+    "API",
+    "Linter",
+    "Zally"
+  ],
+  "author": "",
```
mfellner: Let's remove it if it's empty.

web-ui/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/zalando-incubator/zally/issues"
+  },
+  "homepage": "https://github.com/zalando-incubator/zally#readme",
+  "dependencies": {
```
mfellner: I would recommend fixed versions (without `^`) and to use `npm install --save-exact` for installing new packages. Even better, we might want to switch to yarn and add the lockfile to the Docker image to eliminate the problem of changing transitive dependencies in production.

web-ui/scripts/scm-source.js
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const scm = require('node-scm-source');
+
+fs.writeFileSync('./scm-source.json', JSON.stringify(scm(), null, 4));
```
mfellner: Why not use the built-in [scm-source command](https://github.com/zalando-stups/node-scm-source#npm-scripts-and-cli) instead?

web-ui/webpack.config.js
```diff
@@ -0,0 +1,32 @@
+const webpack = require('webpack');
+const path = require('path');
+const pkg = require('./package.json');
+
+const BUILD_DIR = path.resolve(__dirname, 'src/client/public/assets');
+const APP_DIR = path.resolve(__dirname, 'src/client/app');
+
+var config = {
+  entry: APP_DIR + '/index.jsx',
```
mfellner: We should use `path.join` here.

web-ui/webpack.config.js
```diff
@@ -0,0 +1,32 @@
...
+  output: {
+    path: BUILD_DIR,
```
mfellner: Should this path be excluded from git? We have the bundle (web-ui/src/client/public/assets/bundle.js) in the repository now.

web-ui/webpack.config.js
```diff
@@ -0,0 +1,32 @@
...
+    filename: 'bundle.js'
+  },
+  module : {
+    loaders : [
```
mfellner: This `loaders` definition is now deprecated in webpack 2 and should be called 'rules'. See https://webpack.js.org/configuration/

web-ui/Dockerfile
rbarilani: I will do... but I  I'm more for stay up to date even if something break.

web-ui/package.json
rbarilani: I will do.... but same as before I prefer using latest versions. @mfellner

web-ui/scripts/scm-source.js
rbarilani: @mfellner Didn't know about that. will do.

web-ui/webpack.config.js
rbarilani: @mfellner Why path.resolve is not fine?

web-ui/webpack.config.js
rbarilani: @mfellner It depends I prefer to keep it, so the user doesn't have to run the build that means that doesn't have to install dev dependencies.

web-ui/webpack.config.js
rbarilani: @mfellner Will do

web-ui/package.json
mfellner: I prefer using the latest versions too but stability through locked down versions saves more headaches in the long run. We could try using https://greenkeeper.io to send us automatic pull requests for new versions.

web-ui/webpack.config.js
mfellner: I see. I generally prefer not including build artifacts in the repository because they may be outdated and just add bloat. Maybe in this case it can be beneficial because then you don't need to run the frontend build just to start the whole service. Then on the other hand, we're not including the compiled Java code either!

I'd recommend adding https://github.com/srs/gradle-node-plugin and then to remove the build artifact.

mfellner: See #180

rbarilani:
@mfellner If you can merge... afterwards we should adjust codacy rules to "play" better with javascript and frontend technologies.

mfellner:
👍

fmueller:
👍 

As discussed, we merge without Codacy and exclude codacy runs for web-ui folder in future.

PR #185, opened 2017-03-19
#171 implement rule + test

### LINKED ISSUE #171: Add rule: Hostname should not contain protocol related information
A swagger file can contain field for hostname. If this field contains protocol prefix (e.g. https://) a violation should be created because this information should be placed in scheme field.

mkulak:
Closes #171


server/src/main/java/de/zalando/zally/rules/NoProtocolInHostRule.kt
```diff
@@ -0,0 +1,21 @@
...
+@Component
+class NoProtocolInHostRule : AbstractRule() {
+    override val title = "Host should not contain protocol"
+    override val url = ""
+    override val violationType = ViolationType.MUST
+    val desc = "Information about protocol should be placed in schema"
+
+    override fun validate(swagger: Swagger): Violation? {
+        val host = swagger.host.orEmpty()
+        return if (host.startsWith("http://") || host.startsWith("https://")) {
```
fmueller: I think we should be more strict here: only check for `://` in host field. Because it can also be `ws://` or something else. What do you think?

fmueller: I think we should be more strict here: only check for `://` in host field. Because it can also be `ws://` or something else. What do you think?

mkulak: I think it's ok.

fmueller:
👍

fmueller:
👍

fmueller:
👍

PR #187, opened 2017-03-20
web-ui: fix server request, close #186

### LINKED ISSUE #186: Fix server request in UI
The web ui currently only sends the link from textbox. The api definition from the link should be fetched first and then send to the server.

rbarilani:
fix server request, close #186
fetch the schema, transform to json and send the request to zally api server.

### DIFF
--- a/web-ui/src/server/fetch.js
+++ b/web-ui/src/server/fetch.js
@@ -1,13 +1,13 @@
 const fetch = require('node-fetch');
 
 function handleResponse(response) {
-  if (response.status >= 200 && response.status < 300) {
-    return Promise.resolve(response)
-  } else {
+  if(response.status >= 400) {
     var error = new Error(response.statusText || response.status);
-    error.response = response;
+    error.status = response.status || 500;
     return Promise.reject(error);
   }
+
+  return Promise.resolve(response);
 }
 
 module.exports = function () {
--- a/web-ui/src/server/fetch.js
+++ b/web-ui/src/server/fetch.js
@@ -1,0 +1,17 @@
+const fetch = require('node-fetch');
+
+function handleResponse(response) {
+  if (response.status >= 200 && response.status < 300) {
+    return Promise.resolve(response)
+  } else {
+    var error = new Error(response.statusText || response.status);
+    error.response = response;
+    return Promise.reject(error);
+  }
+}
+
+module.exports = function () {
+  return fetch
+    .apply(null, arguments)
+    .then(handleResponse);
+};
--- a/web-ui/src/server/zally-api-handler.js
+++ b/web-ui/src/server/zally-api-handler.js
@@ -1,9 +1,59 @@
 const env = require('./env');
 const logger = require('./logger');
-const request = require('request');
+const fetch = require('./fetch');
+const yaml = require('js-yaml');
+
+const parseSchema = (text) => {
+  try {
+    // try json first
+    return JSON.parse(text);
+  } catch(err) {}
+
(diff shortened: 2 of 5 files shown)


fmueller:
I excluded `web-ui` project folder from codacy analysis. Just do a dummy commit on this branch - so Codacy will start again.

web-ui/src/server/fetch.js
```diff
@@ -0,0 +1,17 @@
+const fetch = require('node-fetch');
+
+function handleResponse(response) {
+  if (response.status >= 200 && response.status < 300) {
```
mfellner: I wouldn't reject HTTP codes in the 100 and 300 ranges as errors, because they're not. Technically, only 400 and up are considered errors.

web-ui/src/server/fetch.js
```diff
@@ -0,0 +1,17 @@
...
+    return Promise.resolve(response)
+  } else {
+    var error = new Error(response.statusText || response.status);
+    error.response = response;
```
mfellner: Adding the complete response object to the error might be a bit much.  If we only the the status code then let's try this:

```javascript
error.status = response.statusCode || response.status || 500;
```

web-ui/src/server/zally-api-handler.js
```diff
@@ -1,9 +1,59 @@
...
-  var url = env.ZALLY_API_URL + req.url.replace('/zally-api', '');
-  logger.debug(`Proxying request to: ${url}`);
-  req.pipe(request(url)).pipe(res);
+  const url = env.ZALLY_API_URL + req.url.replace('/zally-api', '');
+  const apiDefinitionURL = req.body.api_definition;
+
+  logger.debug(`Fetch swagger schema: ${apiDefinitionURL}`);
+
+  fetch(apiDefinitionURL)
+    .then((response) => {
```
mfellner: I'd recommend upgrading to Node 7.6+ (by requiring it in [package.json](https://docs.npmjs.com/files/package.json#engines)) and use async/await in this function to avoid nesting. We're already using node:7.7-cd45 in the Dockerfile so we're good to go.

web-ui/src/server/zally-api-handler.js
```diff
@@ -1,9 +1,59 @@
...
+  } catch(err) {}
+
+  return yaml.safeLoad(text);
+};
 
 module.exports = function (req, res) {
-  var url = env.ZALLY_API_URL + req.url.replace('/zally-api', '');
-  logger.debug(`Proxying request to: ${url}`);
-  req.pipe(request(url)).pipe(res);
+  const url = env.ZALLY_API_URL + req.url.replace('/zally-api', '');
```
mfellner: It wasn't any clearer before, but maybe let's add at least a comment to understand what url we're actually creating here.

web-ui/src/server/zally-api-handler.js
```diff
@@ -1,9 +1,59 @@
 const env = require('./env');
 const logger = require('./logger');
-const request = require('request');
+const fetch = require('./fetch');
+const yaml = require('js-yaml');
+
+const parseSchema = (text) => {
```
mfellner: Let's keep our style consistent and use proper function declaration, i.e. `function name { }`, when declaring functions.

mfellner:
👍

rbarilani:
👍

fmueller:
As discussed, travis is read because it has some network issues and we still have an unwanted integration with codacy.

PR #191, opened 2017-03-20
188: Skip auth header if token is not specified

### LINKED ISSUE #188: Don't send Authentication header when token is not specified

netme:
Closes #188


fmueller:
👍

PR #192, opened 2017-03-20
Fix "Core Technical Concepts" header in README.md

netme:
Reason: it was rendered not properly. 

📗  Updates only docs


mkulak:
:+1:

PR #193, opened 2017-03-21
web-ui: add minor Dockerfile improvement

rbarilani:
Add Dockerfile improvements:

* install dependencies (invalidate cache just when package.json change)
* ignore not compiled files when copy the src


fmueller:
👍

PR #195, opened 2017-03-21
web-ui: client code refactor

rbarilani:
Add client code refactor:

* use native fetch instead of jquery.ajax (reduce bundle size, consistent with what it's used in the web-ui-backend)
* modularize services and components (use React "functional" components for the one without state)
* "inject" services to components via props


web-ui/src/client/app/components/violations.jsx
```diff
@@ -1,53 +1,60 @@
...
+      <h4 className="dc-h4">
+        <ViolationType type={props.violation.violation_type} />
+        {props.violation.violation_type} {'\u2013'} {props.violation.title}
+      </h4>
+
+      <p>{props.violation.description}</p>
+
+      { props.violation.rule_link && <ViolationRuleLink ruleLink={props.violation.rule_link} /> }
+
+      { props.violation.paths.length && <ViolationPaths paths={props.violation.paths} /> }
```
mfellner: React.render doesn't allow `undefined` arguments, so this could cause problems I think. See https://github.com/facebook/react/issues/3738. I would suggest two options:

A. Use the ternary operator:
```jsx
{ props.violation.paths.length ? <ViolationPaths paths={props.violation.paths} /> : null }
```

B. Create a [higher-order component](https://facebook.github.io/react/docs/higher-order-components.html) that takes a predicate function:

```jsx
<Optional predicate={() => !!props.violation.paths.length}>
  <ViolationPaths paths={props.violation.paths} />
<Optional/>
```

B2. Or as a function:
```jsx
function optional(componentClass: ReactClass, predicate: () => boolean): ReactClass {
  return return class extends React.Component { ... }
}

const OptionalViolationPaths = optional(ViolationPaths, () => !!props.violation.paths.length)

<OptionalViolationPaths paths={props.violation.paths} />
```

rbarilani: @mfellner Really like the high-order component technique, B proposal.

web-ui/src/client/app/components/util.jsx
```diff
@@ -0,0 +1,3 @@
+export function If(props) {
```
rbarilani: @mfellner what about this simple solution?

mfellner: @rbarilani I guess that works! 👍

mfellner:
👍

rbarilani:
👍

fmueller:
👍

mfellner:
👍

Issue #179, opened 2017-03-17
Add linting and code coverage.


bhaskarmelkani:
- [x] Add Linting

- [ ] Add Unit Tests(Server)

- [ ] Add Unit Tests(Client)

- [ ] Code Coverage(Reports)

- [ ] Update Travis

- [ ] Codeclimate

- [ ] Badges

mfellner:
We should consider using https://github.com/prettier/prettier for code formatting. It's already widely adopted in the JS community and solves all code formatting headaches in pull requests.

bhaskarmelkani:
#212 For unit tests of client.

bhaskarmelkani:
#213 for TravisCI and coverage.

PR #199, opened 2017-03-21
web-ui: Add ES Linting

bhaskarmelkani:
Add ES Linting for web-ui. 

Related to #179.

### DIFF
--- a/web-ui/.eslintrc
+++ b/web-ui/.eslintrc
@@ -1,0 +1,35 @@
+extends: eslint:recommended
+env:
+  es6: true
+rules:
+  indent:
+  - error
+  - 2
+  quotes:
+  - 1
+  - single
+  - avoid-escape
+  semi:
+  - 2
+  - always
+  space-before-function-paren:
+  - 2
+  - always
+  keyword-spacing:
+  - 2
+  - before: true
+    after: true
+  space-infix-ops: 2
+  spaced-comment:
+  - 2
+  - always
+  arrow-spacing: 2
+  no-console: 2
+  no-empty: 2
+  eqeqeq:
+  - 2
+  - always
+  no-unused-vars: 1
+  no-unsafe-negation: 2
+  prefer-const: 2
+  no-var: 2
--- a/web-ui/README.md
+++ b/web-ui/README.md
@@ -1,3 +1,5 @@
+[![dependencies Status](https://david-dm.org/zalando-incubator/zally/status.svg?path=web-ui)](https://david-dm.org/zalando-incubator/zally?path=web-ui)
+
 Zally WEB-UI
 ============
 
@@ -83,3 +85,7 @@ npm run build
 npm run docker:build
 ```
 
+## Contributing
+People interested contributing to the web-ui project can open issues and related pull requests. 
+
+Before opening PRs, be sure the test are running by executing `npm test`.
--- a/web-ui/package.json
+++ b/web-ui/package.json
@@ -10,7 +10,10 @@
     "start": "NODE_ENV=production node src/server/index.js",
     "dev": "NODE_ENV=development nodemon src/server/index.js",
     "build": "webpack -p",
-    "docker:build": "scm-source > scm-source.json && docker build -t pierone.stups.zalan.do/architecture/zally-web-ui:${npm_package_version}-$(date +%s) ."
+    "docker:build": "scm-source > scm-source.json && docker build -t pierone.stups.zalan.do/architecture/zally-web-ui:${npm_package_version}-$(date +%s) .",
+    "lint": "eslint src",
+    "test": "npm run lint",
+    "lint:checkstyle-report": "npm run lint -- -f checkstyle --output-file target/checkstyle.xml"
   },
   "repository": {
     "type": "git",
@@ -46,6 +49,7 @@
     "babel-loader": "6.4.0",
     "babel-preset-es2015": "6.24.0",
     "babel-preset-react": "6.23.0",
+    "eslint": "3.18.0",
     "node-scm-source": "0.2.0",
     "nodemon": "1.11.0",
     "webpack": "2.2.1",
--- a/web-ui/src/client/.eslintrc
+++ b/web-ui/src/client/.eslintrc
@@ -1,0 +1,9 @@
+extends: eslint:recommended
+env:
+  browser: true
+parserOptions:
+  sourceType: module
+  ecmaFeatures:
+    experimentalObjectRestSpread: true
+rules:
+  strict: 0
--- a/web-ui/src/server/create-http-server.js
+++ b/web-ui/src/server/create-http-server.js
@@ -1,3 +1,5 @@
+'use strict';
+
 const env = require('./env');
 const fs = require('fs');
 const http = require('http');
@@ -11,13 +13,13 @@ function createHttpsServer(app) {
 }
 
 
-function createHttpServer(app) {
+function createHttpServer (app) {
   return http.createServer(app);
 }
 
 
 module.exports = function (app) {
-  if(env.SSL_ENABLED) { return createHttpsServer(app) }
+  if (env.SSL_ENABLED) { return createHttpsServer(app); }
 
   return createHttpServer(app);
 };
--- a/web-ui/src/server/env-handler.js
+++ b/web-ui/src/server/env-handler.js
@@ -1,3 +1,5 @@
+'use strict';
+
 const fs = require('fs');
 const env = require('./env');
 const logger = require('./logger');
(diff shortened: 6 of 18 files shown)


web-ui/package.json
```diff
@@ -10,7 +10,9 @@
     "start": "NODE_ENV=production node src/server/index.js",
     "dev": "NODE_ENV=development nodemon src/server/index.js",
     "build": "webpack -p",
-    "docker:build": "scm-source > scm-source.json && docker build -t pierone.stups.zalan.do/architecture/zally-web-ui:${npm_package_version}-$(date +%s) ."
+    "docker:build": "scm-source > scm-source.json && docker build -t pierone.stups.zalan.do/architecture/zally-web-ui:${npm_package_version}-$(date +%s) .",
+    "lint": "eslint .",
+    "lint:checkstyle-report": "npm run lint -- -f checkstyle --output-file target/checkstyle.xml"
```
rbarilani: @bhaskarmelkani Can you add a `test` task that run the lint and in the future will run tests.
If Yes can you also add that to the README as a mandatory task that should run successfully?

web-ui/src/server/env-handler.js
```diff
@@ -1,3 +1,5 @@
+'use strict';
```
rbarilani: @bhaskarmelkani 'use strict' is not required since we are using and requiring node 7 and up

web-ui/package.json
bhaskarmelkani: Done

web-ui/src/server/env-handler.js
bhaskarmelkani: Have added 'use strict'. As guess its required.

bhaskarmelkani:
👍

web-ui/.eslintrc
```diff
@@ -0,0 +1,25 @@
+{
+  "extends": "eslint:recommended",
+  "env": {
+    "es6": true,
+    "node": true
```
mfellner: * We should have separate .eslintrc files for client and server. Some common rules can stay in the toplevel .eslintrc though. But env.node does not apply to `client/`. So I would suggest putting only configuration that applies to **both** client and server into the toplevel file.
* Let's write .eslintrc files in [yaml syntax](https://www.json2yaml.com). Much cleaner.
* I would recommend using the [babel plugin](https://github.com/babel/eslint-plugin-babel) and the [babel-eslint parser](https://github.com/babel/babel-eslint).

web-ui/.eslintrc
```diff
@@ -0,0 +1,25 @@
...
+  },
+  "rules": {
+    "strict": 2,
+    "indent": ["error", 2],
+    "quotes": [1, "single", "avoid-escape"],
+    "semi": [2, "always"],
```
mfellner: I would suggest using [eslint-prettier](https://github.com/prettier/prettier-eslint) and the prettier formatter for consistent code style instead of configuring individual rules in eslint.

web-ui/README.md
```diff
@@ -1,3 +1,5 @@
+[![dependencies Status](https://david-dm.org/zalando-incubator/zally/status.svg?path=web-ui)](https://david-dm.org/zalando-incubator/zally?path=web-ui)&nbsp;&nbsp;&nbsp;&nbsp;
```
mfellner: What's going on with all those whitespaces after the URL? I don't think we need them.

web-ui/package.json
```diff
@@ -10,7 +10,10 @@
     "start": "NODE_ENV=production node src/server/index.js",
     "dev": "NODE_ENV=development nodemon src/server/index.js",
     "build": "webpack -p",
-    "docker:build": "scm-source > scm-source.json && docker build -t pierone.stups.zalan.do/architecture/zally-web-ui:${npm_package_version}-$(date +%s) ."
+    "docker:build": "scm-source > scm-source.json && docker build -t pierone.stups.zalan.do/architecture/zally-web-ui:${npm_package_version}-$(date +%s) .",
+    "lint": "eslint .",
```
mfellner: We only need to lint the src directory. So should be `"eslint src"`.

web-ui/src/client/.eslintrc
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "eslint:recommended",
+  "parserOptions": {
+    "sourceType": "module",
+    "ecmaFeatures": { "experimentalObjectRestSpread": true }
+  },
+
+  "env": {
+    "es6": true,
+    "node": false,
```
mfellner: If we don't declare `node: true` in the root .eslintrc then we don't need to turn it off here.

web-ui/src/server/create-http-server.js
```diff
@@ -1,3 +1,5 @@
+'use strict';
```
mfellner: I am quite sure that we do not need 'use strict'; on the **server side**!

web-ui/webpack.config.js
```diff
@@ -1,11 +1,13 @@
+'use strict';
```
mfellner: webpack.config.js does not need 'use strict'; because it doesn't run in a browser.

web-ui/src/server/create-http-server.js
rbarilani: @mfellner I also thought that node 7 was already 'use strict' by default but then @bhaskarmelkani  shows me that without use strict doesn't trigger an error if you write code like this:

```js
a = 2;
console.log(a)
```

web-ui/.eslintrc
rbarilani: @bhaskarmelkani Agree with @mfellner about yaml and top level common eslint ... about babel eslint parser, it would only work for client code since server side doesn't use babel so it will be inconsistent. Then it's a matter of taste when I develop fast I don't like lint noise, I check them sometimes before committing or opening a PR.

web-ui/.eslintrc
rbarilani: @mfellner Here I disagree. I would like to use the standard eslint and don't introduce a new tool.

web-ui/.eslintrc
bhaskarmelkani: Sure @mfellner, makes sense I will modularise .eslintrc and use yml format due to its good readability.

web-ui/webpack.config.js
bhaskarmelkani: @mfellner I think we don't need 'use strict' when we are writing ES6 modules and they are transpiled, as the transpiler uses 'use strict' by default else it is needed.

web-ui/src/server/create-http-server.js
bhaskarmelkani: We don't need 'use strict' on server side when we are writing ES6 modules as the transpiler takes care of it.
But if we don't, as we are doing in our case, we need it.

web-ui/.eslintrc
bhaskarmelkani: Also read this in the README of [prettier](https://github.com/prettier/prettier)

Warning: This is a beta, and the format may change over time. If you aren't OK with the format changing, wait for a more stable version.

web-ui/README.md
bhaskarmelkani: @mfellner Was adding one more badge, but them removed it. Will remove the spaces.

web-ui/package.json
bhaskarmelkani: Sure, will fix it.

web-ui/.eslintrc
mfellner: @rbarilani We can still use the standard eslint! I didn't mean linking to prettier-eslint (eslint-prettier) but to the [eslint prettier plugin](https://github.com/not-an-aardvark/eslint-plugin-prettier) instead.

When using the --fix option it runs prettier to format the code. The real question is if we are ready to adopt a tool that is going to format all the code with an opinionated style (I'd still configure it to use single-quotes and maybe line-width 100 or so)?

web-ui/src/server/create-http-server.js
mfellner: @bhaskarmelkani Ok, thanks for clarifying. Let's keep it then! 👍

web-ui/.eslintrc
mfellner: @bhaskarmelkani It's already pretty stable and many projects use it. But there is a chance that it could introduce problems so we may want to hold off using it.

bhaskarmelkani:
👍

mfellner:
👍

rbarilani:
👍

mfellner:
@bhaskarmelkani @rbarilani we can "rebase merge" on GitHub, no need to update the PR...

mfellner:
👍

bhaskarmelkani:
cool 👍

PR #207, opened 2017-03-22
web-ui: Added Unit Tests for server

bhaskarmelkani:
Added unit tests for server with coverage report. 

Related to #179.

### DIFF
--- a/web-ui/.istanbul.yml
+++ b/web-ui/.istanbul.yml
@@ -1,0 +1,14 @@
+instrumentation:
+  root: src/server
+  extensions:
+    - .js
+  default-excludes: true
+  include-all-sources: true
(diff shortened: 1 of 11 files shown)


bhaskarmelkani:
@mfellner Hey I was planning to use Jest for React but didn't know that its for server as well.
It will be good to replace mocha+sinon+chai with Jest as both frontend and backend can then be tested in similar way.

mfellner:
@bhaskarmelkani Cool! You can take a look at https://github.com/zalando-incubator/tessellate where we're using Jest on the server side.

Should we migrate to Jest in a follow-up pull request?

bhaskarmelkani:
Sure (Y)

PR #219, opened 2017-03-29
202 fix definitions rules

### LINKED ISSUE #202: Several rules that checks json are doing it wrong
Common pattern for checking json: iterate over content of **definitions** section.
This approach has 2 issues:
 - in-place definitions in paths are skipped
 - definitions inside other definitions are skipped

Full list of affected rules:
`AvoidSynonymsRule.kt`, 
`CommonFieldNamesRule.kt`
`PluralizeNamesForArraysRule.kt`
`SnakeCaseInPropNameRule.kt`

Suggestion: introduce abstract DefinitionsRule (like HttpHeadersRule) that will take care of properly iterating over all json definitions.

mkulak:
Closes #202 

Instead of implementing common ancestor for all rules that deals with definitions, I opted for providing convenient way to iterate over all object definitions. This solution is more flexible cause it gives greater freedom for each rule. 

Again, it's extremely hard to cover all cases. Hopefully it covers significant percent of possible situations. Time will tell.


fmueller:
👍

PR #220, opened 2017-03-30
210 base path

### LINKED ISSUE #210: Increase the length for the 'Base path can be extracted' Hint ?
@maschleg and I have got this hint when checking our brand new API :)

```
Found the following HINT violations
===================================

Base path can be extracted
  All paths start with prefix '/ap'. This prefix could be part of base path.
  http://zalando.github.io/restful-api-guidelines/naming/Naming.html
```

We have multiple paths which start with `/ap`: `/api`, `/applications`, etc.

The rule definitely makes sense. Though, I believe increasing the length would make the hint more effective. Alternatively, we can have this hint only for paths which have the same word in the prefix: `api-version`, `api-definition` -> `api/version`, `api/definition`.

mkulak:
Closes #210 

Minor refactoring in separate commit.


server/src/main/java/de/zalando/zally/rules/AbstractRule.kt
```diff
@@ -2,20 +2,18 @@ package de.zalando.zally.rules
 
 abstract class AbstractRule() : Rule {
 
-    override fun getName(): String {
-        return javaClass.simpleName.toLowerCase()
-    }
+    override val name: String = javaClass.simpleName
```
mkulak: I need this later for prettier config.

fmueller:
👍

fmueller:
👍

PR #228, opened 2017-04-01
web-ui: add msg component tests

rbarilani:
web-ui: add msg component tests.

@mfellner @bhaskarmelkani Two doubts are in my head now that I start working with Jest, React plus node all together.....

* For the "react client" we should use snapshot testing or doing like in this PR and test the DOM? 
I see some potential problems in the approach used in this PR with DOM testing, caused by the usage of css class names to locate dom nodes... can be problematic to maintain but on the other hand it "really" assures that the dom structure is the one that we expect.

* We should use the folder structure that seems "suggested" in JEST documentation?

```
---\
  |-- client
      |-- components
          |-- dress-code.jsx
          |-- __tests__
             |-- dress-code.test.jsx
  |-- server
      |-- env-handler.js
      |-- __tests__
         |-- env-handler-spec.js
```

With the above folder structure I see two benefits:
a) tests are close to the source code (personally something that I like)
b) don't go too crazy with relative paths (eg .```import {Msg} from '../../../../../src/client/app/components/dress-code.jsx``` will be ```ìmport {Msg} from '../dress-code.jsx'```)

@mfellner @bhaskarmelkani WDYT?

### DIFF
--- a/web-ui/src/client/app/components/dress-code.jsx
+++ b/web-ui/src/client/app/components/dress-code.jsx
@@ -4,12 +4,13 @@ export class Msg extends React.Component {
   constructor(props) { super(props) }
 
   render() {
+    const type = this.props.type || 'info';
     return (
-      <div className={"dc-msg dc-msg--" + this.props.type }>
+      <div className={"dc-msg dc-msg--" + type }>
         <div className="dc-msg__inner">
 
           <div className="dc-msg__icon-frame">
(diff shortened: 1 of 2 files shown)


bhaskarmelkani:
I guess for unit tests,  to me it makes sense to have the the directory structure suggested by JEST.

rbarilani:
👍

bhaskarmelkani:
👍

Issue #197, opened 2017-03-21
web-ui: Improve error UI and UI in general


rbarilani:
- [ ] ~~Try to show a meaningful error message when the /api-violations call fail~~
- [x] Clean the previous violations list if something wrong happen when we call /api-violations
- [x] Show a nice no violations message (good job!) if the schema is valid!

rbarilani:
close this in favor of #282

PR #235, opened 2017-04-03
web-ui: Improve error

bhaskarmelkani:
Improve error.
Comply with [problem schema](https://zalando.github.io/problem/schema.yaml#/Problem).
Related to #197.

### DIFF
--- a/web-ui/src/server/errors.js
+++ b/web-ui/src/server/errors.js
@@ -1,0 +1,16 @@
+'use strict';
+
+function ProblemError (error) {
+  Error.captureStackTrace(this, this.constructor);
+  this.type = error.type || 'about:blank';
+  this.title = error.title;
+  this.status = error.status;
+  this.detail = error.detail;
+  this.instance = error.instance;
+}
+require('util').inherits(ProblemError, Error);
+
+
+module.exports = {
+  ProblemError
+};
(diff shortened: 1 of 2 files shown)


web-ui/src/server/errors.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+function ProblemError (error) {
+  Error.captureStackTrace(this, this.constructor);
+  this.type = error.type || 'about:blank';
+  this.title = error.title;
+  this.status = error.status;
+  this.detail = error.detail;
+  this.instance = error.instance;
```
rbarilani: @bhaskarmelkani BTW right now zally api accept also urls... so zally-api-handler will be removed, we will call directly the API from the browser or just doing a simple proxy if we face CORS problems.

In any case I would likely create the ProblemError with something like this:

```js
class ProblemError extends Error {
  constructor(message, extra = {}) {
    super(message);
    Object.assign(this, {
      type: 'about:blank',
      title: 'UnknownProblemErrorTitle',
      detail: 'UnknownProblemErrorDetail',
      status: 500,
      instance: 'about:blank'
    }, extra);
    Error.captureStackTrace(this, this.constructor); // after initialize properties
  }
}

const error = new ProblemError('Specific message for developers shown only on server side', {
  type: 'http://www.google.com',
  title: 'Cannot fetch http://www.google.com, fetch fatal error',
  detail: 'Something went wrong!'
});


// try out your self!
console.log(JSON.stringify(specificProblemError, null, 2));

throw error;
```

PR #237, opened 2017-04-04
web-ui: add refresh token feature #196

### LINKED ISSUE #196: web-ui: refresh token
Try to refresh the token when some http call fails with 401 error.
If the access_token gets refreshed retry the original http call.

### LINKED ISSUE #213: Setup TravisCI and coverage reports for webUI
* ~~Check if we can have an independent setup fro webui in travis CI.~~
* Add code coverage reports forWebUI.

rbarilani:
Retry failed 401 requests after refresh token, if refresh token fail redirect to login.
Add two new ENV variables: DEBUG, OAUTH_REFRESH_TOKEN_URL.
Modularize OAuth services.
Use Aurelia Http Client to be able to intecept requests and failed responses.

close #196
close #213


web-ui/src/client/app/services/oauth-interceptor.js
```diff
@@ -0,0 +1,115 @@
...
+  request (request) {
+    if (OAuthProvider.hasAccessToken()) {
+      request.headers.append('Authorization', `Bearer ${OAuthProvider.getAccessToken()}`);
+    }
+    return request;
+  },
+  responseError (response, request) {
+
+    if (!shouldHandleResponseError(response, request)) {
+      return Promise.reject(response);
```
maxim-tschumak: Maybe add some log information here, in order to have a track of unforeseen errors.

maxim-tschumak:
:+1:

web-ui/src/client/app/services/oauth-interceptor.js
rbarilani: @maxim-tschumak Here is just a skip... so it's not an error for the interceptor, it will maybe be an error for the cosumer.

maxim-tschumak:
:+1:

PR #246, opened 2017-04-06
web-ui: codecov integration, adjust codecov upload command

rbarilani:
web-ui: codecov integration, adjust codecov upload command.

NOTE: @fmueller we should delete and recreate the project @ http://codecov.io and then maybe codecov will just consider webui reports.


rbarilani:
👍

fmueller:
👍

PR #248, opened 2017-04-06
182 Fix ProblemJson Rule

### LINKED ISSUE #182: UseProblemJsonRule is incorrect
Current implementation of UseProblemJsonRule is incorrect:
```
if (httpCode in 400..599 && responseEntry.value?.schema?.type != "ref") { //TODO MK: implement real check for Problem usage
    listOf("${pathEntry.key} ${opEntry.key} ${responseEntry.key}")
} else emptyList()
```

It checks only for `schema.type` being ref. 

*DoD:*
UseProblemJsonRule checks whether error response is json object with problem structure (including ref to such object)

netme:
Closes #182


server/src/main/java/de/zalando/zally/rules/UseProblemJsonRule.kt
```diff
@@ -35,4 +40,17 @@ class UseProblemJsonRule : AbstractRule() {
             } catch (nfe: NumberFormatException) {
                 null
             }
+
+    private fun isProperRef(response: Response) : Boolean {
+        return if (response.schema?.type == "ref") {
+            val schema : RefProperty = response.schema as RefProperty
+            (schema.`$ref` == "#/definitions/Problem")
```
netme: Unfortunately, swagger parser imports the ref and marks it as internal, that's why I also need to check fields set.

fmueller:
👍

server/src/main/java/de/zalando/zally/rules/UseProblemJsonRule.kt
```diff
@@ -35,4 +40,17 @@ class UseProblemJsonRule : AbstractRule() {
...
+        } else {
+            false
+        }
+    }
+
+    private fun isProblemDefinitionCorrect(definition: Model?) : Boolean {
+        return (definition?.properties?.keys == setOf("type", "title", "status", "detail", "instance"))
```
fmueller: Does this mean, I only can have these fields in the problem object or does this mean at least these field? The latter one is correct because you are allowed to extend the problem object according to your needs.

server/src/main/java/de/zalando/zally/rules/UseProblemJsonRule.kt
fmueller: Ok ;)

server/src/main/java/de/zalando/zally/rules/UseProblemJsonRule.kt
netme: I see, then some adjustments are still needed. Will do it.

netme:
@fmueller , I've addressed your comments in  6d7cc5c...d7b7cb7

server/src/main/java/de/zalando/zally/rules/UseProblemJsonRule.kt
```diff
@@ -35,4 +42,32 @@ class UseProblemJsonRule : AbstractRule() {
             } catch (nfe: NumberFormatException) {
```
mkulak: In Kotlin 1.1 method `String.toIntOrNull()` is in standard library. Our custom method can be removed.

server/src/main/java/de/zalando/zally/rules/UseProblemJsonRule.kt
```diff
@@ -35,4 +42,32 @@ class UseProblemJsonRule : AbstractRule() {
...
                 null
             }
+
+    private fun isProperRef(swagger: Swagger, response: Response) : Boolean {
+        return if (response.schema?.type == "ref") {
+            val schema : RefProperty = response.schema as RefProperty
+            return isProblemDefinitionCorrect(swagger, swagger.definitions[schema.simpleRef])
```
mkulak: `return` can be removed here

server/src/main/java/de/zalando/zally/rules/UseProblemJsonRule.kt
```diff
@@ -35,4 +42,32 @@ class UseProblemJsonRule : AbstractRule() {
             } catch (nfe: NumberFormatException) {
                 null
             }
+
+    private fun isProperRef(swagger: Swagger, response: Response) : Boolean {
+        return if (response.schema?.type == "ref") {
```
mkulak: Seems like current implementation prohibits inlined Problem definition. 
1. Is it true? 
2. If yes, do we want to support this usecase?

server/src/main/java/de/zalando/zally/rules/UseProblemJsonRule.kt
```diff
@@ -35,4 +42,32 @@ class UseProblemJsonRule : AbstractRule() {
...
+            return isProblemDefinitionCorrect(swagger, swagger.definitions[schema.simpleRef])
+        } else {
+            false
+        }
+    }
+
+    private fun isProblemDefinitionCorrect(swagger: Swagger, definition: Model?) : Boolean {
+        val properties = getProperties(swagger, definition)
+        return requiredFields.filter { field -> !properties.contains(field) }.isEmpty()
+    }
```
mkulak: I propose to do this:
```
private fun isProblemDefinitionCorrect(swagger: Swagger, definition: Model?) = getProperties(swagger, definition).containsAll(requiredFields)
```
and inline this method.

server/src/main/java/de/zalando/zally/rules/UseProblemJsonRule.kt
```diff
@@ -35,4 +42,32 @@ class UseProblemJsonRule : AbstractRule() {
...
+
+    private fun getProperties(swagger: Swagger, definition: Model?) : Set<String> {
```
mkulak: I propose to implement it like that to avoid nested `when`:
```
private fun getProperties(swagger: Swagger, definition: Model?): Set<String> =
        when (definition) {
            is ComposedModel -> definition.allOf.orEmpty().flatMap { getProperties(swagger, it) }.toSet()
            is RefModel -> getProperties(swagger, swagger.definitions[definition.simpleRef])
            else -> definition?.properties?.keys.orEmpty()
        }
```
For me it looks a bit more readable.

server/src/main/java/de/zalando/zally/rules/UseProblemJsonRule.kt
netme: @mkulak , fair point, implemented in c5710b7

netme:
@mkulak , many thanks for your valuable comments, I've learned something and addressed them in aa7d6bd

mkulak:
:+1:

PR #264, opened 2017-04-11
258 Require only title and status for problem-json

### LINKED ISSUE #258: Should Problem JSON check more liberal?
Currently we assume that all fields of Problem JSON have to be in the payload - and we check for this. I'd like to discuss if it makes sense if we are more liberal on this check and make some fields optional? Because in the schema none of the field is marked as required, see https://zalando.github.io/problem/schema.yaml

netme:
Closes #258


mkulak:
:+1:

fmueller:
Codacy hangs, so I merge it manually.

PR #266, opened 2017-04-12
web-ui: copy scm-source in the expected root folder

rbarilani:
Fix scm-source copied in the wrong folder causing violations.


fmueller:
👍

PR #270, opened 2017-04-12
Revert extra health and metrics endpoint port

fmueller:
This causes more issues than it brings us anything - since metrics endpoint is secured by oauth anyway.


netme:
👍

PR #271, opened 2017-04-13
web-ui: Add logout

bhaskarmelkani:


web-ui/README.md
```diff
@@ -68,6 +69,16 @@ DEBUG=false
 * **OAUTH_TOKENINFO_URL**: The url used to validate the access token and retrieve token information
 * **OAUTH_REFRESH_TOKEN_URL**: The url used to refresh the access token
 * **OAUTH_SCOPES**: Comma separated list of scopes that the user should grant to the app
+* **OAUTH_USERNAME_FIELD_PATH**: Path of the field that stores username in the retrived token information.
+Eg.
```
rbarilani: I would suggest something like:

**OAUTH_USERNAME_PROPERTY**: Property that can be found in the ```/tokeninfo``` response representing the username of the connected user (eg. ```uid``` or ```user.uid``` if nested)

web-ui/src/server/env.js
```diff
@@ -19,7 +19,8 @@ const defaults = {
   OAUTH_SCOPES:'',
   DEV_PORT: 8441,
   DEV_SSL_ENABLED: false,
-  DEBUG: false
+  DEBUG: false,
+  OAUTH_USERNAME_FIELD_PATH: 'cn'
```
rbarilani: After change the property to  `OAUTH_USERNAME_PROPERTY` reorder the usage so is grouped with the others OAUTH_* properties.

web-ui/src/client/app/components/user-info-container.jsx
```diff
@@ -0,0 +1,54 @@
...
+
+export default function UserInfoContainer(props){
+
+  const showUserInfo = window.env.OAUTH_ENABLED;
+
+  if(!showUserInfo){
+    return null;
+  }
+
+  const css = `
```
rbarilani: Move the style to top level

web-ui/README.md
bhaskarmelkani: Done

web-ui/src/server/env.js
bhaskarmelkani: Done

web-ui/src/client/app/components/user-info-container.jsx
bhaskarmelkani: Done

bhaskarmelkani:
👍

PR #273, opened 2017-04-13
272 Do not specify path with linter URL

### LINKED ISSUE #272: Define /api-violations Path inside ZallyApiClient
Currently we assume that /api-violations path should be always defined inside URL parameter. We need to accept only base URL and append /api-violations to it every time we are doing a request.

netme:
Closes #272


fmueller:
👍

Issue #257, opened 2017-04-10
web-ui: editor/url mode


rbarilani:
**As** a web-ui user
**I Want** to copy and paste a Swagger schema content
**So that** I can validate it without the need of publish the version of the schema online

Right now the web-ui only allows the validation of a schema that is reachable by URL while the rest api supports also the validation of a schema provided as a string.
Add an editor where the user can paste in a schema and validates it.
I would like to suggest adding a navigation bar (related to #256) where the user can navigate to different modes/views (url or editor, default to url view).
The editor view will be on a 2 columns layout on desktop screen: right column with the editor (suggested [react-ace](https://www.npmjs.com/package/react-ace)), left column with the violations list and or error/success messages.

Probably this feature needs a routing mechanism:
Navigate to /#/editor or navigate to /#/url to open related views so the user can also directly go to the mode that is more interested into just by providing the right url in the address bar. (see [react-router](https://reacttraining.com/react-router/))

rbarilani:
closed via #274

PR #274, opened 2017-04-16
256 257 editor and logout

### LINKED ISSUE #256: web-ui: logout
**As** a logged-in web-ui user
**I Want** to logout 
**So that** I can quickly login with another user and/or clean my user session

I would like to suggest adding a "navigation" bar with the current logged in user identifier and a logout call to action. 
When the user clicks logout, clear the user session (remove tokens from local storage) and redirect him to login (```requestToken```).

rbarilani:
Big PR.

close #256 #257 

@bhaskarmelkani I merged into this branch your work for #256 with some refactor, plus:

* add editor mode
* introduce routing with react router
* use browser history instead of hash history (because was not playing well with implicit grant flow and for having nicer urls)
* introduce SASS for styling 
* lint also jsx with React support
* add client services tests
* add navigation bar with logo
* add footer with github link and copyright
* add mixed visual improvements
screen shot 2017-04-16 at 11 54 54
screen shot 2017-04-16 at 11 58 00


fmueller:
👍

PR #278, opened 2017-04-18
web-ui: update url placeholder, close #276

### LINKED ISSUE #276: Fix placeholder value in URL text box
The link which is given as an example in URL text box is not working, the link to raw view has to be provided instead.

rbarilani:
web-ui: update url placeholder, close #276


netme:
👍

PR #280, opened 2017-04-18
web-ui: fix boolean as string ENV variables parsing

rbarilani:
Small hotfix to correcly evaluate boolean like ENV variable passed as strings.
Eg. ```OAUTH_ENABLED="true" npm start```

```OAUTH_ENABLED``` will be parsed as a boolean:

```js
const env = require('./env');

console.log(env.OAUTH_ENABLED === true); // true
```


fmueller:
👍

PR #288, opened 2017-04-19
web-ui: show 'meaningful' error message from rest api, close #282

### LINKED ISSUE #282: web-ui: show error message text using rest api response.
web-ui: show error message text using rest api response.

When and if zally rest api error responses will expose something that we can directly show to the end user use that otherwise fallback to general error message (aka something went wrong...).

**As** A user 
**I Want** to read a meaningful error message when something wrong happened
**So That** I can understand the problem root cause

rbarilani:
web-ui: show 'meaningful' error message from rest api, close #282

* update tests using shallow rendering also for "abstract" violations component
* getApiViolations catch the error and "reject" with json body


fmueller:
👍

PR #289, opened 2017-04-19
Fix handling of whitelisted property names

### LINKED ISSUE #96: _links field is not whitelisted

fmueller:
Closes #96

### DIFF
--- a/server/src/main/java/de/zalando/zally/rules/SnakeCaseInPropNameRule.kt
+++ b/server/src/main/java/de/zalando/zally/rules/SnakeCaseInPropNameRule.kt
@@ -16,9 +16,11 @@ class SnakeCaseInPropNameRule : AbstractRule() {
     override val code = "M012"
     val description = "Property names must be snake_case (and never camelCase)"
 
+    private val whitelist = setOf("_links")
+
     override fun validate(swagger: Swagger): Violation? {
         val result = swagger.getAllJsonObjects().flatMap { (def, path) ->
-            val badProps = def.keys.filterNot(PatternUtil::isSnakeCase)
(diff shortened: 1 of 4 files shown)


server/src/main/java/de/zalando/zally/rules/SnakeCaseInPropNameRule.kt
```diff
@@ -16,9 +16,11 @@ class SnakeCaseInPropNameRule : AbstractRule() {
     override val code = "M012"
     val description = "Property names must be snake_case (and never camelCase)"
 
+    private val whitelist = setOf("_links")
```
mkulak: I think this value should be in config. You can find example in `AvoidJavascriptKeywordsRule`

server/src/main/java/de/zalando/zally/rules/SnakeCaseInPropNameRule.kt
```diff
@@ -16,9 +16,11 @@ class SnakeCaseInPropNameRule : AbstractRule() {
...
+
     override fun validate(swagger: Swagger): Violation? {
         val result = swagger.getAllJsonObjects().flatMap { (def, path) ->
-            val badProps = def.keys.filterNot(PatternUtil::isSnakeCase)
+            val badProps = def.keys.filterNot { d -> PatternUtil.isSnakeCase(d) || whitelist.contains(d) }
```
mkulak: This can also be written as:
```
val badProps = def.keys.filterNot { PatternUtil.isSnakeCase(it) || it in whitelist }
```
(not insisting, just FYI)

fmueller:
@mkulak I incorporated your points. :)

mkulak:
👍

Issue #290, opened 2017-04-19
Editor adding additional strings


wilq87:
Editor is behaving very strange (adding additional string values) when trying to edit properties in yaml file. Had a property "type:" and after pressing 7 times "t" after "e" letter it became:
"typetttttttddddddypettttttddddddypetttttddddddypettttddddddypetttddddddypettddddddypetddddddype:"

fmueller:
@rbarilani Can you give your 2 cents please?

rbarilani:
@fmueller Working on it.

rbarilani:
@wilq87 Can you be so kind to give us some more informations to reproduce the bug:

* Browser Version
* The schema that you are trying to modify.

Thx in advance.

wilq87:
@rbarilani 
Browser:
Version 57.0.2987.98 Built on Ubuntu , running on Ubuntu 16.04 (64-bit)

copy pasted API yaml for wholesale-article

rbarilani:
@wilq87 I we deployed a new version can you be so gentle to try again with the same schema?

wilq87:
@rbarilani works like a charm! thanks

rbarilani:
@wilq87 👍 Great. Thanks for your contribution.

PR #292, opened 2017-04-19
web-ui: catch and show parsing error

rbarilani:
Catch and show an error to the user if schema can't be parsed (contains YAML errors).
Probably closing #290


fmueller:
👍

fmueller:
👍

PR #297, opened 2017-04-20
web-ui: remove zalando oauth specific env variables and logic, yeah!

rbarilani:
web-ui: remove zalando oauth specific env variables and logic, yeah!

**NOTE**: with the new platform IAM for non confidential clients (the one using OAuth Implicit Grant Flow) the client id doesn't rotate so we don't need to read credentials.json file.


rbarilani:
👍

fmueller:
👍

PR #302, opened 2017-04-25
301 Enhance Monitoring for Violated Synonyms

### LINKED ISSUE #301: Enhance monitoring of violated synonyms
**As** an operator
**I would** like to see which synonym from synonym dictionary was violated
**so that** I can adjust the synonym dictionary on a data-driven approach

netme:
Closes #301

### DIFF
--- a/server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
+++ b/server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
@@ -55,6 +55,10 @@ class AvoidSynonymsRule(
     }
 
     fun submitStatistics(names : List<String>) {
-        names.map { name -> metricServices.submit("histogram.rules.avoid-synonyms.synonym." + name, 1.0) }
+        names.groupBy { it }.map {
+            metricServices.submit(
+                    "histogram.rules.avoid-synonyms.synonym." + it.key, it.value.count().toDouble()
+            )
+        }
     }
 }
--- a/server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
+++ b/server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
@@ -6,10 +6,12 @@ import de.zalando.zally.ViolationType
 import de.zalando.zally.utils.getAllJsonObjects
 import io.swagger.models.Swagger
 import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.actuate.metrics.dropwizard.DropwizardMetricServices
 import org.springframework.stereotype.Component
 
 @Component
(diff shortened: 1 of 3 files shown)


server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
```diff
@@ -6,33 +6,54 @@ import de.zalando.zally.ViolationType
...
                     .joinToString("\n")
             Violation(this, title, "$descPattern:\n$details", violationType, url, paths.toSet().toList())
-        } else null
+        } else {
+            null
+        }
+    }
+
+    fun submitStatistics(names : List<String>) {
+        names.map { name -> metricServices.submit("histogram.rules.avoid-synonyms.synonym" + name, 1.0) }
```
fmueller: Can you put a `.` before the `name`? That's our naming scheme for metrics

server/src/main/resources/application.yml
```diff
@@ -25,4 +25,4 @@ rules-config-path: "rules-config.conf"
 
 # FYI: S001 is disabled because Zalando's API guidelines changed.
 zally:
-  ignoreRules: S001,S010
```
fmueller: Leave it ignored. This is a configuration change later.

netme: OK, will change it back.

server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
netme: Good spot. Sure

netme:
@fmueller , comments are addressed

server/src/main/java/de/zalando/zally/rules/AvoidSynonymsRule.kt
```diff
@@ -6,33 +6,54 @@ import de.zalando.zally.ViolationType
...
                     .joinToString("\n")
             Violation(this, title, "$descPattern:\n$details", violationType, url, paths.toSet().toList())
-        } else null
+        } else {
+            null
+        }
+    }
+
+    fun submitStatistics(names : List<String>) {
+        names.map { name -> metricServices.submit("histogram.rules.avoid-synonyms.synonym." + name, 1.0) }
```
fmueller: After having a second look to semantics of histogram: would it make sense to give number of violated paths per name instead of `1.0`? This way we would know, how often a synonym name was violated in an API review. By providing `1.0`we only know that it was violated. What do you think?

netme: @fmueller , yes, nice idea, will try to fix it.

netme:
@fmueller , I've addressed your comment in bdc2974

fmueller:
👍

Issue #299, opened 2017-04-24
Show supported rules in CLI


netme:
After merging #275 we can implement CLI support for listing supported rules. 

Possible interface:

```bash
zally [--rules | -R]
```

PR #327, opened 2017-05-11
299: Use ViolationType Enum Instead of String Values

netme:
Clean up for #299 

Reason: it's really tricky to maintain hardcoded values, might be easier to use Enums instead.


fmueller:
👍

netme:
Thanks for checking, @fmueller

PR #329, opened 2017-05-12
328 Do not Require ProblemJson in Payload for HEAD and OPTION

### LINKED ISSUE #328: Not verify payloads for payload-less HTTP operations
**Current Behaviour**

Zally checks for HEAD and OPTIONS operations if decent error payload (Problem JSON) is provided - although such HTTP operations do not have a payload.

**Expected Behaviour**

Zally does not check payloads for payload-less HTTP operations.

netme:
Closes #328

### DIFF
--- a/server/src/main/java/de/zalando/zally/rules/UseProblemJsonRule.kt
+++ b/server/src/main/java/de/zalando/zally/rules/UseProblemJsonRule.kt
@@ -3,6 +3,7 @@ package de.zalando.zally.rules
 import de.zalando.zally.Violation
 import de.zalando.zally.ViolationType
 import io.swagger.models.ComposedModel
+import io.swagger.models.HttpMethod
 import io.swagger.models.Model
 import io.swagger.models.RefModel
 import io.swagger.models.Response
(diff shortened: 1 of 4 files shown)


server/src/main/java/de/zalando/zally/rules/UseProblemJsonRule.kt
```diff
@@ -54,4 +55,8 @@ class UseProblemJsonRule : AbstractRule() {
             else -> definition?.properties?.keys.orEmpty()
         }
     }
+
+    private fun shouldContainPayload(httpMethod : HttpMethod) : Boolean {
+        return httpMethod.name.toLowerCase() !in listOf("head", "options")
+    }
```
mkulak: Also possible:
```
private fun shouldContainPayload(httpMethod : HttpMethod) : Boolean =
        httpMethod.name.toLowerCase() !in listOf("head", "options")
```

netme: Oh, that's cool, forgot that Kotlin supports this

mkulak: Actually since "can contain payload" is property of HTTP method, it can be even something like this:
```
private fun HttpMethod.shouldContainPayload() : Boolean =
        name.toLowerCase() !in listOf("head", "options")
```
But it's just cosmetics.

netme: Done in 0b963d7

mkulak:
👍

PR #331, opened 2017-05-15
326 Rename COULD to MAY

### LINKED ISSUE #326: Change 'could' rule type to 'may'
The wording in Zalando's Restful API guidelines changed, could rules are now named with may - according to RFC. see https://github.com/zalando/restful-api-guidelines/issues/226

We need to adapt server, cli and web ui accordingly.

netme:
Closes #326

### DIFF
--- a/server/src/main/java/de/zalando/zally/ApiViolationsController.java
+++ b/server/src/main/java/de/zalando/zally/ApiViolationsController.java
@@ -71,9 +71,6 @@ public class ApiViolationsController {
     private void setCounters(List<Violation> violations, ObjectNode result) {
         ViolationsCounter counter = new ViolationsCounter(violations);
         for (ViolationType violationType : ViolationType.values()) {
-            if (violationType == ViolationType.COULD) {
-                continue;
-            }
             result.put(violationType.toString().toLowerCase(), counter.getCounter(violationType));
         }
     }
--- a/server/src/main/java/de/zalando/zally/ApiViolationsController.java
+++ b/server/src/main/java/de/zalando/zally/ApiViolationsController.java
@@ -71,6 +71,9 @@ public class ApiViolationsController {
     private void setCounters(List<Violation> violations, ObjectNode result) {
         ViolationsCounter counter = new ViolationsCounter(violations);
         for (ViolationType violationType : ViolationType.values()) {
+            if (violationType == ViolationType.COULD) {
+                continue;
+            }
             result.put(violationType.toString().toLowerCase(), counter.getCounter(violationType));
         }
     }
--- a/server/src/main/java/de/zalando/zally/ViolationType.java
+++ b/server/src/main/java/de/zalando/zally/ViolationType.java
@@ -4,6 +4,7 @@ public enum ViolationType {
 
     MUST("must"),
     SHOULD("should"),
+    COULD("could"),     // DEPRECATED, see #326
     MAY("may"),
     HINT("hint");
 
--- a/server/src/main/java/de/zalando/zally/ViolationType.java
+++ b/server/src/main/java/de/zalando/zally/ViolationType.java
@@ -4,7 +4,7 @@ public enum ViolationType {
 
     MUST("must"),
     SHOULD("should"),
-    COULD("could"),
+    MAY("may"),
     HINT("hint");
 
     private final String metricIdentifier;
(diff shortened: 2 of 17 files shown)


server/src/main/java/de/zalando/zally/rules/NestedPathsMayBeRootPathsRule.kt
```diff
@@ -7,13 +7,13 @@ import io.swagger.models.Swagger
...
 @Component
-class NestedPathsCouldBeRootPathsRule : AbstractRule() {
+class NestedPathsMayBeRootPathsRule : AbstractRule() {
     override val title = "Consider Using (Non-) Nested URLs"
     override val url = "http://zalando.github.io/restful-api-guidelines/resources/Resources.html" +
-            "#could-consider-using-non-nested-urls"
-    override val violationType = ViolationType.COULD
+            "#may-consider-using-non-nested-urls"
+    override val violationType = ViolationType.MAY
     override val code = "C001"
```
netme: Question: which codes should we use for `MAY`? 
- we can keep `C`
- something like `A` or `Y`

fmueller: I'd stick with C.

server/src/main/resources/api/zally-api.yaml
```diff
@@ -41,7 +41,7 @@ parameters:
     x-extensible-enum:
       - MUST
       - SHOULD
-      - COULD
```
fmueller: We are breaking API backward compatibility if we remove a value in this enum. We can only add `MAY` as an extra value of this enum.

server/src/main/java/de/zalando/zally/ViolationType.java
```diff
@@ -4,7 +4,7 @@
 
     MUST("must"),
     SHOULD("should"),
-    COULD("could"),
```
fmueller: Please do not remove Could because there are clients out there which are depending on this field.

netme:
@fmueller , thanks, I was thinking about compatibility topic, but didn't found an elegant solution. Will probably keep `COULD` as an extra parameter. 

How about web-ui changes btw, is `COULD` also somehow hard-coded there? Maybe we can ask some feedback from @rbarilani .

rbarilani:
@netme At the moment we are not using COULD. but yep we use the rule type to change the color of the label... see https://github.com/zalando-incubator/zally/blob/master/web-ui/src/client/app/components/rules.jsx

netme:
Thanks @rbarilani for a quick answer. Any preference on style I need to use for `MAY` rule? Anyway, the only `MAY` rule is currently disabled on the backend configuration level, we'll not see any immediate impact.

server/src/main/java/de/zalando/zally/ViolationType.java
netme: COULD is back

server/src/main/java/de/zalando/zally/ViolationType.java
```diff
@@ -4,7 +4,8 @@
...
+    COULD("could"),     // DEPRECATED, see #326
```
fmueller: Maybe we can use `@Deprecated` here?

server/src/main/java/de/zalando/zally/ApiViolationsController.java
```diff
@@ -71,6 +71,9 @@ private void reportViolationMetrics(List<Violation> violations) {
     private void setCounters(List<Violation> violations, ObjectNode result) {
         ViolationsCounter counter = new ViolationsCounter(violations);
         for (ViolationType violationType : ViolationType.values()) {
+            if (violationType == ViolationType.COULD) {
```
fmueller: This is not backward compatible. What if clients rely on `COULD` counter? It's not bad to add new fields in an API response payload and not delete old ones. ;)

server/src/main/java/de/zalando/zally/ViolationType.java
netme: Done in 7570e28

server/src/main/java/de/zalando/zally/ApiViolationsController.java
netme: Reverted in aeda083

netme:
@fmueller, also added an integration test to simulate how old CLI will behave.

fmueller:
👍

PR #332, opened 2017-05-16
Fix for special characters suffix in api definition urls

### LINKED ISSUE #287: Retry fetching of API definition URL without trailing special characters
The server should retry fetching of an API definition from given URL without trailing special characters - if there are any.

fmueller:
Closes #287 

Also:
- updates Kotlin and Problem Web library
- improved structure of class `ApiDefinitionReader`
- some minor refactorings


netme:
Just a minor comment, it's usually nice to split multiple changes into multiple commits :-) 
But PR is still readable.

netme:
👍

Issue #298, opened 2017-04-24
web-ui: support OAuth authorization code flow


rbarilani:
**As** a developer
**I want** to choose which authentication flow the web-ui should use (Implicit or Authorization code)
**So that** I can secure my application following my security requirements

**As** a user
**I want** to login and stay logged in until my refresh token doesn't expire
**So that** I don't have to login every time the access token expire

Make the OAUTH flow configurable, supporting Implicit or Authorization code flows.

**NOTE** with implicit grant flow and using Zalando as OAuth Provider the refresh token will not be returned in the future but the app should not break.

rbarilani:
moved to internal repo

Issue #294, opened 2017-04-19
web-ui: OAuth show an error to the user if firewall fails


rbarilani:
**As** A user
**I Want** to be notified if something went wrong during login
**So That** I can be aware of the root cause

rbarilani:
moved to internal repo

PR #334, opened 2017-05-16
web-ui: remove authentication logic from core and other zalando specifics

### LINKED ISSUE #308: web-ui: refactor the app as an express "subapp", publish it as a npm package
**As** a developer
**I want** to be able to mount the web-ui on existing express app
**So that** I can easily integrate with an existing express server

Refactor the app, removing all the part that are not "core" and publish it as a npm package.

The developer should be able to use it as a standalone without any "opinionated" way of configure it (right now done with dotenv) and be able to mount it to an existing express app.

### The proposal:

**Install**
```bash
npm install zally-web-ui
```
> Package name TBD, can be:

* ```zally-web-ui```
* ```@zally/web-ui```
* ```@zalando-incubator/zally-web-ui```

**run as standalone**
```js
const ui = require('zally-web-ui')();
ui.listen(3000);
```

**mount to an existing express app and customize it**
```js
const myLogger = require('./my-logger');
const app = require('express')();
const ui = require('zally-web-ui')({
  env: {
     ZALLY_API_URL: 'https://locahost:8080',
     DEBUG: true
  },
  logger: myLogger
});

app.use('/ui', ui);

app.listen(3000);
```

components to be removed from the core:

* oauth routes and oauth support
* dotenv
* ~webpack-dev-server support~

rbarilani:
close #308 
related to #294 #298 

**The web-ui project was moved on his own repo: https://github.com/zalando-incubator/zally-web-ui, to allow easy npm integration, independent CI job, etc.**

~This PR removes all zalando specifics from the core: health endpoint, scm-source, specific docker build, authentication (all moved to another repo that use this package as the core an then extend it with zalando specifics)~

~@fmueller can we move the web-ui folder to another repo? this would help when using npm and git reference to install the package: `npm install git+https://github.com/zalando-incubator/zally-web-ui`~


bhaskarmelkani:
👍

fmueller:
@rbarilani We stick with mono-repository approach. Otherwise we would have to separate `cli` as well.

rbarilani:
@fmueller At least CLI and REST API server are based on the same stack.
The point here is that we have to publish the package on npm repository every time we have a change that we wanna test on our internal running instance and it's overkill (while if zally-web-ui it's another repo we can do "npm install git+ssh://...." and grab also unpublished/unstable version ) The other option is cloning the repository with a sparse checkout (to checkout just the web-ui folder) during the build and do some other tricks... basically not using npm to manage the dependency to the zally-web-ui core (not really optimal).

Just my 2 cents.

fmueller:
@rbarilani Let's discuss this offline. ;-)

rbarilani:
👍

fmueller:
👍

PR #339, opened 2017-05-18
server: add cors support for /support-rules endpoint

rbarilani:
server: add cors support for /support-rules endpoint and add OAuth Configuration (without the access it's always denied - 403)

This to allow zally-web-ui integration without a proxy.

@fmueller does it make sense to have CORS support globally instead of per route basis?


fmueller:
👍

fmueller:
@rbarilani Globally does not make sense imho since we have some non-api endpoints which don't care about CORS...

rbarilani:
@fmueller ... like `/metrics` and stuff, right? Ok.

PR #340, opened 2017-05-18
313 Use Correct HTTP Codes

### LINKED ISSUE #313: Add rule for using correct HTTP status code per HTTP method
Check if an API uses status codes which are not allowed for given http method. See [Must: Use Specific HTTP Status Codes](http://zalando.github.io/restful-api-guidelines/http/Http.html) for specification.

netme:
Closes #313


server/src/main/java/de/zalando/zally/rules/UseSpecificHttpStatusCodes.kt
```diff
@@ -0,0 +1,31 @@
...
+
+@Component
+class UseSpecificHttpStatusCodes : AbstractRule() {
+    override val title = "Use Specific HTTP Status Codes"
+    override val url = "http://zalando.github.io/restful-api-guidelines/http/Http.html" +
+        "#must-use-specific-http-status-codes"
+    override val violationType = ViolationType.MUST
+    override val code = "M016"
+
+    private val allowedStatusCodes = mapOf(
```
mkulak: I think it should be in config

mkulak: Oh, its wip. Ok, will comment later )

netme: Addressed the comment in e43e029

netme:
@fmueller, @mkulak could you please take a look on this PR?

server/src/main/java/de/zalando/zally/rules/UseSpecificHttpStatusCodes.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.stereotype.Component
+
+@Component
+class UseSpecificHttpStatusCodes(@Autowired rulesConfig: Config) : AbstractRule() {
+    override val title = "Use Specific HTTP Status Codes"
+    override val url = "http://zalando.github.io/restful-api-guidelines/http/Http.html" +
+        "#must-use-specific-http-status-codes"
+    override val violationType = ViolationType.MUST
+    override val code = "M016"
+    private val description = "Operatons should use specific HTTP status codes"
```
netme: @fmueller, I thought more deeply about it, might be a bit hard to do it since we are raising only 1 violation for all paths + methods, I can still list supported status codes for every HTTP method, but it might be a bit too much. Any other ideas?

fmueller: Hm ya, right. Can be too much text... let's discuss this tomorrow offline in the office.

fmueller:
👍

PR #341, opened 2017-05-19
server: use dash for path specified on RestController annotation

rbarilani:
server: use dash for path specified on RestController annotation


fmueller:
👍

Issue #353, opened 2017-05-24
Run Go CLI Tests on Travis


netme:
#### Acceptance criteria
1. Go CLI tests are run on Travis CI with every build
2. If possible, Go CLI test coverage report is uploaded to Codacy

maxim-tschumak:
I don't have much experience with TravisCI, but as far as I understood, there is no support for multi-language environments. ([TravisCI issue](https://github.com/travis-ci/travis-ci/issues/4090))

netme:
@maxim-tschumak , I think it's possible, we had built node.js together with Java before. We can check it on Monday, if it won't be officially supported we'll find some workaround.

PR #368, opened 2017-06-07
353 Temporary disable Golang test run

netme:
Tests will be enabled once Travis will be using Go 1.8

Related to #353


maxim-tschumak:
:+1:

PR #369, opened 2017-06-08
Add api deduplication to statistics

### LINKED ISSUE #357: Enhance statistics report by api deduplication
As an API Guild Member
I'd like to see how often the same API was submitted
so that I can see how much the API definition was improved over time.

For this, the statistic job should be aware of API names and data structure as well. API name should be extracted from submitted API definition.

fmueller:
Closes #357

### DIFF
--- a/server/src/main/java/de/zalando/zally/apireview/ApiNameParser.java
+++ b/server/src/main/java/de/zalando/zally/apireview/ApiNameParser.java
@@ -1,0 +1,21 @@
+package de.zalando.zally.apireview;
+
+import io.swagger.models.Swagger;
+import io.swagger.parser.SwaggerParser;
+
+public class ApiNameParser {
+
+    private ApiNameParser() {}
+
+    public static String extractApiName(String apiDefinition) {
+        try {
+            Swagger swagger = new SwaggerParser().parse(apiDefinition);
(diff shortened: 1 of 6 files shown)


server/src/main/java/de/zalando/zally/apireview/ApiNameParser.java
```diff
@@ -0,0 +1,21 @@
...
+public class ApiNameParser {
+
+    private ApiNameParser() {}
+
+    public static String extractApiName(String apiDefinition) {
+        try {
+            Swagger swagger = new SwaggerParser().parse(apiDefinition);
+            if (swagger == null || swagger.getInfo() == null) {
+                return null;
+            }
```
netme: optional: Actually this `if` is not needed because later the `NullPointerException` will be catched :-)

netme:
👍

server/src/main/java/de/zalando/zally/apireview/ApiNameParser.java
fmueller: Ya, that's right but on the other hand, I know here in advance that it can be `null`. So make this obvious is good styl imho. The catching block afterwards is only for unknown exceptions from `SwaggerParser`. :)

PR #370, opened 2017-06-09
Fix CORS support when OAuth is enabled

### LINKED ISSUE #244: CORS browser support doesn't work when OAuth authentication is enabled
CORS browser support doesn't work when OAuth authentication is enabled.

When browser try to send an OPTIONS preflight request to check if CORS is possible the server respond with 401 because the Authorization Header is not present whereas it should not check for credentials for that preflight request.
Take in mind that the OPTIONS preflight request sent by the browser is not under the control of the developer so he/she can't add the Authorization Header.

For more info here a [stack overflow question](http://stackoverflow.com/questions/15734031/why-does-the-preflight-options-request-of-an-authenticated-cors-request-work-in) from somebody facing a similar problem.

fmueller:
Closes #244


netme:
👍

PR #375, opened 2017-06-13
Adding generation type for JPA entites

maxim-tschumak:
Right now, Zally server cannot use real Postgres database due to missing generation type for the JPA resources. This PR fixes this. 

Required for https://github.bus.zalan.do/team-architecture/management/issues/1


netme:
👍

PR #379, opened 2017-06-14
Change UseSpecificHttpStatusCodes to should violation

### LINKED ISSUE #374: Response status code check is too strict
#313/#340 added a check that verified that each operation only returns one of the status codes listed in http://zalando.github.io/restful-api-guidelines/http/Http.html#must-use-specific-http-status-codes (it also checks that the HTTP method fits).

I don't think that rule is meant to be understood as "don't ever use any other status code than those listed here" (the list says "Our list of most commonly used and best understood HTTP status codes") – so a violation against this should not be a "MUST" violation.

Also, it disallows the use of "default", which (with a response-body of a problem object) is IMHO fine.

fmueller:
Closes #374 

It's a quick fix until Zalando's API guidelines were adapted accordingly.

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/UseSpecificHttpStatusCodes.kt
+++ b/server/src/main/java/de/zalando/zally/rule/UseSpecificHttpStatusCodes.kt
@@ -14,7 +14,7 @@ class UseSpecificHttpStatusCodes(@Autowired rulesConfig: Config) : AbstractRule(
     override val title = "Use Specific HTTP Status Codes"
     override val url = "http://zalando.github.io/restful-api-guidelines/http/Http.html" +
         "#must-use-specific-http-status-codes"
-    override val violationType = ViolationType.MUST
+    override val violationType = ViolationType.SHOULD
     override val code = "M016"
     private val description = "Operatons should use specific HTTP status codes"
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/de/zalando/zally/rule/UseSpecificHttpStatusCodes.kt
```diff
@@ -14,7 +14,7 @@ class UseSpecificHttpStatusCodes(@Autowired rulesConfig: Config) : AbstractRule(
     override val title = "Use Specific HTTP Status Codes"
     override val url = "http://zalando.github.io/restful-api-guidelines/http/Http.html" +
         "#must-use-specific-http-status-codes"
-    override val violationType = ViolationType.MUST
+    override val violationType = ViolationType.SHOULD
     override val code = "M016"
```
netme: Should the code also be also changed to `S`?

netme: Or maybe some comment should be added, now it's a bit confusing :-)

fmueller: I'll add a comment in the code and reference the issue.

fmueller:
@netme Added comment, can you please re-check it? :)

netme:
👍 

Thanks for adding a comment

netme:
Restarted builds, let's see if it will help

fmueller:
This won't help because I have to adapt our integration tests. ;-)

netme:
👍

netme:
Thanks for fixing

PR #385, opened 2017-06-19
add rule: should explicitly define collection format #336

### LINKED ISSUE #336: Add should rule for collections format value in API definitions
Zalando's API guidelines changed: a should rule was added that one should define the collections format value for query parameters explicitly (only `multi` and `csv` are allowed), see https://github.com/zalando/restful-api-guidelines/pull/236

Zally should have a `SHOULD` rule for this as well.

maschleg:
Closes #336 

first time kotlin, so I am thankful for every improvement hint :)

Not ready jet I have to add tests for ``fromPaths`` cases.


server/src/main/java/de/zalando/zally/rule/QueryParameterCollectionFormatRule.kt
```diff
@@ -0,0 +1,45 @@
...
+            path.parameters.extractInvalidQueryParam(name) + path.operations.flatMap { operation ->
+                operation.parameters.extractInvalidQueryParam(name)
+            }
+        }
+
+        val allHeaders = fromParams + fromPaths
+        val paths = allHeaders
+            .map { "${it.first} ${it.second}" }
+            .toSet()
+            .toList()
```
netme: Maybe `distinct()` ?

server/src/main/java/de/zalando/zally/rule/QueryParameterCollectionFormatRule.kt
```diff
@@ -0,0 +1,45 @@
...
+    override val violationType = ViolationType.SHOULD
+    override val code = "S011"
+    val formatsAllowed = listOf("csv", "multi")
+    val violationDescription = "CollectionFormat should be one of: " + formatsAllowed
+
+    override fun validate(swagger: Swagger): Violation? {
+        fun Collection<Parameter>?.extractInvalidQueryParam(path: String) =
+            orEmpty().filterIsInstance<QueryParameter>()
+                .filter { it.`type` == "array" && it.`collectionFormat` !in formatsAllowed }
+                .map { path to it.name }
```
netme: You can also define this function outside

server/src/main/java/de/zalando/zally/rule/QueryParameterCollectionFormatRule.kt
```diff
@@ -0,0 +1,45 @@
...
+import io.swagger.models.parameters.Parameter
+import io.swagger.models.parameters.QueryParameter
+
+class QueryParameterCollectionFormatRule : AbstractRule() {
+
+    override val title = "Explicitly define the Collection Format of Query Parameters"
+    override val url = "http://zalando.github.io/restful-api-guidelines/http/Http.html#should-explicitly-define-the-collection-format-of-query-parameters"
+    override val violationType = ViolationType.SHOULD
+    override val code = "S011"
+    val formatsAllowed = listOf("csv", "multi")
```
netme: Optional: can be also part of configuration. 
Example: https://github.com/zalando-incubator/zally/blob/master/server/src/main/java/de/zalando/zally/rule/UseSpecificHttpStatusCodes.kt#L23

server/src/main/java/de/zalando/zally/rule/QueryParameterCollectionFormatRule.kt
```diff
@@ -0,0 +1,45 @@
...
+    val violationDescription = "CollectionFormat should be one of: " + formatsAllowed
```
netme: Optional:

```kotlin
val violationDescription =  "CollectionFormat should be one of: {formatsAllowed}"
```

server/src/main/java/de/zalando/zally/rule/QueryParameterCollectionFormatRule.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    override fun validate(swagger: Swagger): Violation? {
+        fun Collection<Parameter>?.extractInvalidQueryParam(path: String) =
+            orEmpty().filterIsInstance<QueryParameter>()
+                .filter { it.`type` == "array" && it.`collectionFormat` !in formatsAllowed }
```
netme: Most probably escaping for `collectionFormat` is not required

server/src/main/java/de/zalando/zally/rule/QueryParameterCollectionFormatRule.kt
```diff
@@ -0,0 +1,45 @@
...
+                .map { path to it.name }
+
+        val fromParams = swagger.parameters.orEmpty().values.extractInvalidQueryParam("parameters")
+        val fromPaths = swagger.paths.orEmpty().entries.flatMap { (name, path) ->
+            path.parameters.extractInvalidQueryParam(name) + path.operations.flatMap { operation ->
+                operation.parameters.extractInvalidQueryParam(name)
+            }
```
netme: It's also possible to implement it using pattern-matching, but it will require some refactoring. Here is a small example: https://github.com/zalando-incubator/zally/blob/master/server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt#L51-L57

I will ignore it for now.

server/src/main/java/de/zalando/zally/rule/QueryParameterCollectionFormatRule.kt
netme: Actually, same for `type` :-)

netme:
👍

server/src/main/java/de/zalando/zally/rule/QueryParameterCollectionFormatRule.kt
maschleg: cool maybe we can make a pair programming session :)

server/src/main/java/de/zalando/zally/rule/QueryParameterCollectionFormatRule.kt
maschleg: The external configuration file impairs the rule readability and makes the rule setup more complex. I don't see the benefit of it. I think we should focus more on configurability of the rule set.

server/src/main/java/de/zalando/zally/rule/QueryParameterCollectionFormatRule.kt
netme: @maschleg sure, I'm up

netme:
@maschleg could you please also rebase it and I'll merge it afterwards

netme:
👍

PR #387, opened 2017-06-20
335 Check for password flow when using OAuth2

### LINKED ISSUE #335: Check for OAuth2 Flow Property
In our guidelines we state that the OAuth2 definition should always contain `flow: password`, see http://zalando.github.io/restful-api-guidelines/security/Security.html

Zally's OAuth rules should be enhanced to also validate this as well.

netme:
Closes #335 

Decided to implement it as a separate rule to make it pluggable and to keep separate statistics.


maxim-tschumak:
I've seen many APIs with incorrect OAuth2 flow in API reviews - nice to have this rule automated!
:+1:

netme:
Thanks @maxim-tschumak for checking it! It was a quick win then :-)

fmueller:
Yes, great we have this now :-)

PR #400, opened 2017-06-28
Fixing unresolvable hostname

### LINKED ISSUE #399: Fix unresolvable hostname in RestApiViolationsTest
`RestApiViolationsTest.[shouldReturn404WhenHostNotRecognised,shouldStoreUnsuccessfulApiReviewRequest]` fail on some machines.

There is a chance that `bad.host.localhost` host can be resolved in some setups (because it's a generic subdomain of localhost).

To make the tests more stable we have to use another unresolvable hostname.

maxim-tschumak:
Fixes #399.


fmueller:
👍

PR #404, opened 2017-06-30
Rule for Extensible Enums

### LINKED ISSUE #160: Add rule for extensible enums
If enum is used for a field a should rule violation should be fired because we recommend usage of `x-extensible-enum`, see http://zalando.github.io/restful-api-guidelines/compatibility/Compatibility.html ("Should: Used Open-Ended List of Values (x-extensible-enum) Instead of Enumerations")

maxim-tschumak:
Fixes #160 

My first contact with Kotlin, please be very critical! :)

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
@@ -1,13 +1,26 @@
 package de.zalando.zally.rule
 
-import com.google.common.collect.Sets
 import de.zalando.zally.violation.Violation
-import de.zalando.zally.violation.ViolationType.MUST
 import de.zalando.zally.violation.ViolationType.SHOULD
+import io.swagger.models.Model
 import io.swagger.models.Operation
+import io.swagger.models.Path
 import io.swagger.models.Swagger
-import io.swagger.models.auth.OAuth2Definition
+import io.swagger.models.parameters.Parameter
+import io.swagger.models.parameters.SerializableParameter
+import io.swagger.models.properties.BinaryProperty
+import io.swagger.models.properties.BooleanProperty
+import io.swagger.models.properties.DateProperty
+import io.swagger.models.properties.DateTimeProperty
+import io.swagger.models.properties.DoubleProperty
+import io.swagger.models.properties.EmailProperty
+import io.swagger.models.properties.FloatProperty
+import io.swagger.models.properties.IntegerProperty
+import io.swagger.models.properties.LongProperty
+import io.swagger.models.properties.PasswordProperty
+import io.swagger.models.properties.Property
 import org.springframework.stereotype.Component
+import io.swagger.models.properties.StringProperty
 
 @Component
 class ExtensibleEnumRule : AbstractRule() {
@@ -16,10 +29,90 @@ class ExtensibleEnumRule : AbstractRule() {
         "#should-prefer-compatible-extensions"
     override val violationType = SHOULD
     override val code = "S012"
-    private val DESC = "Prefer using extensible extensions on the server-side"
 
     override fun validate(swagger: Swagger): Violation? {
-        return null
+        val enumProperties = enumPropertiesInDefinitions(swagger)
+        val definitionsWithEnumProperties = definitionsWithEnums(swagger)
+
+        val pathsWithDefinitionsWithEnums = swagger.paths.orEmpty().entries
+            .filter { (_, path) -> havingParameters(path, definitionsWithEnumProperties) }
+            .map { it.key }
+
+        val enumParameters = swagger.paths.orEmpty().values
+            .flatMap{ it.get?.parameters.orEmpty().filter { isEnum(it) } }
+            .map { it.name }
+
+        val pathsWithEnumParameters = swagger.paths.orEmpty().filter { havingParameters(it.value, enumParameters) }.keys
+
+
+        val enums = (enumProperties + enumParameters).distinct()
+        val paths = (pathsWithDefinitionsWithEnums + pathsWithEnumParameters).distinct()
+
+        if (enums.isNotEmpty())
+            return Violation(rule = this, paths = paths, ruleLink = url, violationType = violationType,
+                title = title, description = "Properties/Fields $enums are not extensible enums")
+        else
+            return null
+    }
+
+    private fun havingParameters(path: Path?, names: List<String>): Boolean {
+        fun isIn(op: Operation?): Boolean = op?.parameters.orEmpty().any { it.name in names }
+
+        return isIn(path?.post) || isIn(path?.get) || isIn(path?.put) || isIn(path?.delete) || isIn(path?.options)
+            || isIn(path?.head) || isIn(path?.patch)
+    }
+
+    private fun isEnum(parameter: Parameter?) = (parameter as? SerializableParameter)?.enum?.orEmpty()?.isNotEmpty() ?: false
+
+    private fun definitionsWithEnums(swagger: Swagger): List<String> = swagger.definitions.orEmpty().entries
+        .filter { (_, def) -> enumProperties(def).isNotEmpty() }
+        .map { it.key }
+
+    private fun enumPropertiesInDefinitions(swagger: Swagger): List<String> = swagger.definitions.orEmpty().entries
+        .flatMap { (_, def) -> def.properties.orEmpty().filter { (_, prop) -> isEnum(prop) }.map { it.key } }
+
+    private fun enumProperties(def: Model?): List<String> {
+        return def?.properties.orEmpty()
+            .filter { (_, prop) -> isEnum(prop) }
+            .map { it.key }
+    }
+
+    private fun isEnum(property: Property): Boolean {
+        fun <T> notEmpty(values: List<T>?) = values.orEmpty().isNotEmpty()
+
+        return when (property) {
+            is StringProperty -> notEmpty(property.enum)
+            is BinaryProperty -> notEmpty(property.enum)
+            is DateProperty -> notEmpty(property.enum)
+            is DateTimeProperty -> notEmpty(property.enum)
+            is BooleanProperty -> notEmpty(property.enum)
+            is DoubleProperty -> notEmpty(property.enum)
+            is EmailProperty -> notEmpty(property.enum)
+            is FloatProperty -> notEmpty(property.enum)
+            is IntegerProperty -> notEmpty(property.enum)
+            is LongProperty -> notEmpty(property.enum)
+            is PasswordProperty -> notEmpty(property.enum)
+            else -> false
+        }
+    }
+
+    fun Property.hasEnumValues(): Boolean {
+        fun <T> notEmpty(values: List<T>) = values.orEmpty().isNotEmpty()
+
+        return when (this) {
+            is StringProperty -> notEmpty(enum)
+            is BinaryProperty -> notEmpty(enum)
+            is DateProperty -> notEmpty(enum)
+            is DateTimeProperty -> notEmpty(enum)
+            is BooleanProperty -> notEmpty(enum)
+            is DoubleProperty -> notEmpty(enum)
+            is EmailProperty -> notEmpty(enum)
+            is FloatProperty -> notEmpty(enum)
+            is IntegerProperty -> notEmpty(enum)
+            is LongProperty -> notEmpty(enum)
+            is PasswordProperty -> notEmpty(enum)
+            else -> false
+        }
     }
 
 }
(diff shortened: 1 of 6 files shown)


server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
```diff
@@ -0,0 +1,94 @@
...
+class ExtensibleEnumRule : AbstractRule() {
+    override val title = "Prefer Compatible Extensions"
+    override val url = "http://zalando.github.io/restful-api-guidelines/compatibility/Compatibility.html" +
+        "#should-prefer-compatible-extensions"
+    override val violationType = SHOULD
+    override val code = "S012"
+
+    override fun validate(swagger: Swagger): Violation? {
+        val properties = enumProperties(swagger)
+        val parameters = enumParameters(swagger)
```
fmueller: Do these checks also cover Enums in definitions?

server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
```diff
@@ -0,0 +1,94 @@
...
+
+        val enumNames = (properties.keys + parameters.keys).distinct()
+        val enumPaths = (properties.values + parameters.values).distinct()
+        if (enumNames.isNotEmpty())
```
fmueller: You can put this on one line and make it more expressive: if (condition) value1 else value2

server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
fmueller: Ok, I see below that this covers parameters in definitions as well. Maybe a more precise name for this method would be better.

server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
```diff
@@ -0,0 +1,94 @@
...
+
+    private fun enumParameters(swagger: Swagger): Map<String, String> {
+        fun enumsIn(operation: Operation?): List<String> {
+            fun isEnum(parameter: Parameter?) =
+                (parameter as? SerializableParameter)?.enum?.orEmpty()?.isNotEmpty() ?: false
+
+            return operation?.parameters.orEmpty().filter { isEnum(it) }.map { it.name }
+        }
+
+        val parametersAndPaths = mutableMapOf<String, String>()
```
fmueller: Can you try to work with immutable Map and only flatMap and map operations? Not mandatory for me but would be a big plus to have it more functional.

server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
```diff
@@ -0,0 +1,94 @@
...
+            }
+        }
+
+        return parametersAndPaths
+    }
+
+    private fun isEnum(property: Property): Boolean {
+        fun <T> isNotEmpty(list: List<T>?) = list.orEmpty().isNotEmpty()
+
+        return when (property) {
```
fmueller: Is there a Property base type with enum field we could use instead of this long list?

server/src/test/java/de/zalando/zally/rule/ExtensibleEnumRuleTest.kt
```diff
@@ -0,0 +1,59 @@
...
+    val rule = ExtensibleEnumRule()
+
+    @Test
+    fun returnsNoViolationIfEmptySwagger() {
+        assertThat(rule.validate(Swagger())).isNull()
+    }
+
+    @Test
+    fun returnsViolationIfAnEnumInModelProperty() {
+        val swagger = getFixture("enum_in_model_property.yaml")
```
fmueller: Can you create a test Swagger object dedicated to each test? You find a Test helper at https://github.com/zalando-incubator/zally/blob/master/server/src/test/java/de/zalando/zally/TestUtil.kt

server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
maxim-tschumak: It was my initial implementation (see 3. commit). I decided to switch to imperative style for better readability and to match the style of the other rule implementations.

server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
maxim-tschumak: I've checked this - unfortunately not

server/src/test/java/de/zalando/zally/rule/ExtensibleEnumRuleTest.kt
maxim-tschumak: It's easier to have a sample `yaml` or `json` for the tests of this rule: much less code. This rule checks multiple places in the swagger tree and this implies a lot of code to create a proper swagger instance to test.

server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
maxim-tschumak: The current implementation distinguishes between `properties` (defined and those which are in the paths) and `properties` (fields in definitions). The wording is the same as in Swagger spec.

server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
mkulak: Our `SwaggerParser` never stops to surprise me ).
BTW what is physical meaning of `enum` in PasswordProperty? BooleanProperty? BinaryProperty?
I think we can safely omit check for some of the types. Perhaps even check for just StringProperty (I'm not aware of usage of enums outside String context).

server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
mkulak: Functional is more readable for me:
```kotlin
    private fun enumProperties(swagger: Swagger): Map<String, String> =
        swagger.definitions.orEmpty().flatMap { (defName, model) ->
            val enumPropNames = model.properties.orEmpty().filter { (_, prop) -> isEnum(prop) }.map { it.key }
            enumPropNames.map { propName -> propName to "#/definitions/$defName/properties/$propName" }
        }.toMap()
```

mkulak: Also fun inside fun inside fun looks a bit weird:
```
    private fun enumParameters(swagger: Swagger): Map<String, String> {

        fun enumsIn(operation: Operation?): List<String> {
            fun isEnum(parameter: Parameter?) =

```

I propose change this: 
```kotlin
    private fun enumParameters(swagger: Swagger): Map<String, String> {

        fun enumsIn(operation: Operation?): List<String> {
            fun isEnum(parameter: Parameter?) =
                (parameter as? SerializableParameter)?.enum?.orEmpty()?.isNotEmpty() ?: false

            return operation?.parameters.orEmpty().filter { isEnum(it) }.map { it.name }
        }

        val parametersAndPaths = mutableMapOf<String, String>()

        swagger.paths.orEmpty().forEach { (pathName, path) ->
            path.operationMap.orEmpty().forEach { (opName, op) ->
                enumsIn(op).forEach { parameterName ->
                    parametersAndPaths.put(parameterName, "#/paths$pathName/$opName/parameters/$parameterName")
                }
            }
        }

        return parametersAndPaths
    }

```
to this:
```kotlin
    private fun Parameter?.isEnum() = (this as? SerializableParameter)?.enum.orEmpty().isNotEmpty()

    private fun enumParameters(swagger: Swagger): Map<String, String> =
            swagger.paths.orEmpty().flatMap { (pathName, path) ->
                path.operationMap.orEmpty().flatMap { (opName, op) ->
                    op.parameters.orEmpty().filter { it.isEnum() }.map { param ->
                        param.name to "#/paths$pathName/$opName/parameters/${param.name}"
                    }
                }
            }.toMap()

```

server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
```diff
@@ -0,0 +1,95 @@
...
+        val properties = enumProperties(swagger)
+        val parameters = enumParameters(swagger)
+
+        val enumNames = (properties.keys + parameters.keys).distinct()
+        val enumPaths = (properties.values + parameters.values).distinct()
+        if (enumNames.isNotEmpty()) return Violation(
+            rule = this, violationType = violationType, title = title, ruleLink = url, paths = enumPaths,
+            description = "Properties/Parameters $enumNames are not extensible enums")
+
+        return null
```
mkulak: 'if' in Kotlin is expression so you can write like this:
```kotlin
        return if (enumNames.isNotEmpty()) Violation(
                rule = this, violationType = violationType, title = title, ruleLink = url, paths = enumPaths,
                description = "Properties/Parameters $enumNames are not extensible enums")
        else null
```

Also in my opinion `violationType = violationType, title = title` doesn't provide any information to reader but adds noise.

server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
mkulak: Regarding consistency: which rules are using `forEach`? 
I see 0 usages of it in main kotlin code and 49 usages of `flatMap`. Sorry if it sounds like holy war, you asked to be very critical and I'm trying hard :)

maxim-tschumak:
@mkulak thanks for the comments - you've convinced me to switch back to functional style. Please have another look.

server/src/main/java/de/zalando/zally/rule/ExtensibleEnumRule.kt
```diff
@@ -34,62 +34,48 @@ class ExtensibleEnumRule : AbstractRule() {
...
+        return pathsOperationsAndEnums
+            .filter { (_, opAndEnums) -> opAndEnums.isNotEmpty() }
+            .flatMap { (pathName, opAndEnums) -> opAndEnums.map { (op, enums) -> "#/paths$pathName/$op" to enums } }
+            .flatMap { (operationPath, enums) -> enums.map { it to "$operationPath/parameters/$it" } }.toMap()
     }
 
-    private fun isEnum(property: Property): Boolean {
-        fun <T> isNotEmpty(list: List<T>?) = list.orEmpty().isNotEmpty()
+    private fun Property.isEnum(): Boolean {
+        fun <T> hasValues(list: List<T>?) = list.orEmpty().isNotEmpty()
```
netme: A bit easier:

```kotlin
private fun Property.isEnum(): Boolean = when (this) {
    is StringProperty -> this.enum.hasValues()
    is BinaryProperty -> this.enum.hasValues()
    is DateProperty -> this.enum.hasValues()
    is DateTimeProperty -> this.enum.hasValues()
    is BooleanProperty -> this.enum.hasValues()
    is DoubleProperty -> this.enum.hasValues()
    is EmailProperty -> this.enum.hasValues()
    is FloatProperty -> this.enum.hasValues()
    is IntegerProperty -> this.enum.hasValues()
    is LongProperty -> this.enum.hasValues()
    is PasswordProperty -> this.enum.hasValues()
    else -> false
}

private fun List<String>?.hasValues() : Boolean {
    return this.orEmpty().isNotEmpty()
}
```

netme: Or even a bit shorter:

```kotlin
    private fun Property.isEnum(): Boolean = when (this) {
        is StringProperty, is BinaryProperty, is DateProperty, is DateTimeProperty,
        is BooleanProperty, is DoubleProperty, is EmailProperty, is FloatProperty,
        is IntegerProperty, is LongProperty, is PasswordProperty -> this.enum.hasValues()
        else -> false
    }
```

maxim-tschumak: @netme The short version doesn't compile.
I was making use of Smart Cast feature of Kotlin: compiler tracks `if` and `when` statements with `is`-checks and casts the value to the needed type. In your example compiler doesn't know which concrete type we are expecting -> doesn't make any type conversions -> `this` remains an abstract `Property` which doesn't ultimately have the `enum` field.
Am I missing something?

netme: My fault 🙈 
Please ignore it

netme:
Rebased

netme:
👍

PR #413, opened 2017-07-12
189: Introduce HTTP Timeouts

### LINKED ISSUE #189: Introduce Timeouts to HTTP Client Library
In order to improve user experience, we need to introduce proper timeouts to HTTP client in CLI.

netme:
Closes #189

### DIFF
--- a/cli-go/zally/utils/http_client.go
+++ b/cli-go/zally/utils/http_client.go
@@ -1,0 +1,17 @@
+package utils
+
+import (
+	"net/http"
+	"time"
+)
+
+const httpTimeout = 5 * time.Second
+
+// DoHTTPRequest makes an HTTP request with timeout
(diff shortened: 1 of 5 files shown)


cli-go/zally/utils/http_client.go
```diff
@@ -0,0 +1,17 @@
+package utils
+
+import (
+	"net/http"
+	"time"
+)
+
+const httpTimeout = 5 * time.Second
```
maxim-tschumak: minor: how about `const httpTimeout = time.Duration(5 * time.Second)`?

maxim-tschumak:
:+1:

cli-go/zally/utils/http_client.go
netme: Sorry, missed that comment. I though about it, but found it less clean for defining a constant

PR #414, opened 2017-07-13
excluding zalando specific config from default spring profile (#372)

### LINKED ISSUE #372: Rename production spring profile
Current Zalando specific OAuth implementation is covered as `production` spring profile. This is kinda misleading: we should rename it to `zalando-oauth` and also not make it the default profile. We have to adapt all deployment scripts accordingly after this change.

maxim-tschumak:
Fixes #372


netme:
👍

PR #423, opened 2017-07-18
208: Check that OAuth2 is not used together with http

### LINKED ISSUE #208: Zally should report at least a warning if HTTP scheme is used in combination with OAuth2.0
OAuth2.0 enforces HTTPS as protocol but Zally does not report at least a warning if the spec has both `http` and `https` as schemes defined in combination with `oauth2` as security definition.

netme:
Closes #208


maxim-tschumak:
:+1:

PR #424, opened 2017-07-19
Adding a whitelist for not pluralized prefixes in the paths (#315)

### LINKED ISSUE #315: Pluralize resource names should not fire for base path or possible base path extractions
**Current behaviour**

The rule that one should not use singular in resource names also fires for parts of the `basePath` property. For instance, for `/api` it should not fire.

**Expected behaviour**

The rule `Pluralize resource names` only fires for the names in an endpoint which are not part of the `basePath` or be part of potential `basePath` extractions (combine it with output of Hint rule `Base Patch Extraction`).

maxim-tschumak:
Closes #315.

FYI:
The paths of the parsed `Swagger` object **don't contain** the `basePath` as a prefix. Our current implementation of the `Pluralize resource names` doesn't check the `basePath` - which is OK.

The 'false-positive' occurs, when an API contains following paths (`/api` is not declared to be a `basePath`):
* `api/products`
* `api/products/{id}`

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/PluralizeResourceNamesRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/PluralizeResourceNamesRule.kt
@@ -1,24 +1,31 @@
 package de.zalando.zally.rule
 
+import com.typesafe.config.Config
 import de.zalando.zally.util.PatternUtil
 import de.zalando.zally.util.WordUtil.isPlural
 import de.zalando.zally.violation.Violation
 import de.zalando.zally.violation.ViolationType
 import io.swagger.models.Swagger
+import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 
 @Component
-class PluralizeResourceNamesRule : AbstractRule() {
+class PluralizeResourceNamesRule(@Autowired rulesConfig: Config) : AbstractRule() {
     override val title = "Pluralize Resource Names"
     override val url = "http://zalando.github.io/restful-api-guidelines/naming/Naming.html" +
         "#must-pluralize-resource-names"
     override val violationType = ViolationType.SHOULD
     override val code = "S008"
     private val DESC_PATTERN = "Resources %s are singular (but we are not sure)"
+    private val allowedPrefixes = rulesConfig.getConfig(name).getStringList("whitelist_prefixes")
 
     override fun validate(swagger: Swagger): Violation? {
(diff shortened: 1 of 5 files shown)


server/src/main/java/de/zalando/zally/rule/PluralizeResourceNamesRule.kt
```diff
@@ -1,24 +1,31 @@
...
 @Component
-class PluralizeResourceNamesRule : AbstractRule() {
+class PluralizeResourceNamesRule(@Autowired rulesConfig: Config) : AbstractRule() {
     override val title = "Pluralize Resource Names"
     override val url = "http://zalando.github.io/restful-api-guidelines/naming/Naming.html" +
         "#must-pluralize-resource-names"
     override val violationType = ViolationType.SHOULD
     override val code = "S008"
     private val DESC_PATTERN = "Resources %s are singular (but we are not sure)"
+    private val allowedPrefixes = rulesConfig.getConfig(name).getStringList("whitelist_prefixes")
```
netme: Very good

server/src/main/java/de/zalando/zally/rule/PluralizeResourceNamesRule.kt
```diff
@@ -1,24 +1,31 @@
...
 
     override fun validate(swagger: Swagger): Violation? {
         val res = swagger.paths.keys.flatMap { path ->
-            path.split("/".toRegex()).filter { s -> !s.isEmpty() && !PatternUtil.isPathVariable(s) && !isPlural(s) }
+            val allParts = path.split("/".toRegex())
+            val partsToCheck = if (allParts.size > 1 && allowedPrefixes.contains(allParts.first())) allParts.drop(1)
```
maxim-tschumak: `val partsToCheck = if (allowedPrefixes.contains(allParts.firstOrNull())) allParts.drop(1)  else allParts` doesn't do the same: the current implementation checks if there are at least **2** parts in the path and (only) then checks the first one. This is intended in order to generate a violation if an API has an `/api` path (without having `/api/some-resources`)

`drop(1)` drops `n=1` first elements and returns the remaining list
> /**
>  * Returns a list containing all elements except first [n] elements.
>  */
> public fun <T> Iterable<T>.drop(n: Int): List<T>

netme: Ah, got it, then I've overlooked it

netme:
:+1:

Issue #108, opened 2017-01-17
Replace Bats Framework for Integration Testing


netme:
`Bats` framework is doing its job now, but sometimes it's nightmare to debug it when something is changing on CLI.

## Steps:
1. Find a better tool to test CLI and discuss it with Zally team
2. Reimplement existing integration tests
3. Adjust `.travis.yml`

netme:
@mkulak since you now feel the pain of Bats framework as well, how about kotlinizing the integration tests? ;-)

fmueller:
This has low priority for me. Let's concentrate on improving UX in CLI first.

innokenty:
Closed via #456

PR #425, opened 2017-07-19
cli-go: cli: Used alternative validation approach in Bats tests #108

akryvko:
I've already had my first "Bats hurt" =) so please consider this testing approach. It uses slightly relaxed condition - checks that the substring appears anywhere in the output. I believe it is strong enough but still can save us some time.
Another possible solution: use relative line numbers, e.g. from the end of the file

       [ "${lines[${#lines[@]}-1]}" = "HINT violations: 1" ]
        [ "${#lines[@]}" -eq 71 ]

### DIFF
--- a/cli-go/integration_tests/tests.bats
+++ b/cli-go/integration_tests/tests.bats
@@ -3,53 +3,49 @@
 @test "CLI displays usage when no parameters specified" {
 	run cli-go/zally/zally
 	[ "$status" -eq 0 ]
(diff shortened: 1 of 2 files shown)


maxim-tschumak:
We may still replace `Bats` eventually, but this makes the maintenance of the integration tests much easier. Thanks!
:+1:

netme:
Nice fix!

Issue #28, opened 2016-12-20
Check if problem-json is content type for error status codes


fmueller:
If https://github.com/swagger-api/swagger-parser/pull/303/files is merged, we can check actual content type.

netme:
The issue was closed with:

```
Hi, we don't support this in the spec, so have to punt in the parser :( Consider adding an extension to do this, and I'd be happy to add it to the repo as well.
```

Most probably we need a plan B :-)

netme:
Did few requests against latest master:

```bash
curl -i -X GET http://localhost:8080/
HTTP/1.1 405 
X-Application-Context: application:dev
Allow: POST
Content-Type: application/problem+json
Transfer-Encoding: chunked
Date: Thu, 09 Mar 2017 21:41:15 GMT

{"title":"Method Not Allowed","status":405,"detail":"Request method 'GET' not supported"}
```

```bash
curl -i -X POST http://localhost:8080/
HTTP/1.1 400 
X-Application-Context: application:dev
Content-Type: application/problem+json
Transfer-Encoding: chunked
Date: Thu, 09 Mar 2017 21:42:12 GMT
Connection: close

{"title":"Bad Request","status":400,"detail":"Required request body is missing: public org.springframework.http.ResponseEntity<com.fasterxml.jackson.databind.JsonNode> de.zalando.zally.ApiViolationsController.validate(com.fasterxml.jackson.databind.JsonNode)"}
```

```bash
curl -i -X POST -H "Content-Type: application/json" http://localhost:8080/ -d '{"abcde": 123'
HTTP/1.1 400 
X-Application-Context: application:dev
Content-Type: application/problem+json
Transfer-Encoding: chunked
Date: Thu, 09 Mar 2017 21:43:30 GMT
Connection: close

{"title":"Bad Request","status":400,"detail":"Could not read document: Unexpected end-of-input: expected close marker for Object (start marker at [Source: java.io.PushbackInputStream@4a44189f; line: 1, column: 1])\n at [Source: java.io.PushbackInputStream@4a44189f; line: 1, column: 40]; nested exception is com.fasterxml.jackson.core.io.JsonEOFException: Unexpected end-of-input: expected close marker for Object (start marker at [Source: java.io.PushbackInputStream@4a44189f; line: 1, column: 1])\n at [Source: java.io.PushbackInputStream@4a44189f; line: 1, column: 40]"
```

It's working as expected. Closing this issue.

netme:
[...]

PR #426, opened 2017-07-20
#28 checks if produces application/json in the problem-json rule

cogitor:
A possible solution for #28, when validating the Problems, check if the Produces set contains application/json, both on the top level and in operations


server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt
```diff
@@ -2,12 +2,7 @@ package de.zalando.zally.rule
 
 import de.zalando.zally.violation.Violation
 import de.zalando.zally.violation.ViolationType
-import io.swagger.models.ComposedModel
-import io.swagger.models.HttpMethod
-import io.swagger.models.Model
-import io.swagger.models.RefModel
-import io.swagger.models.Response
-import io.swagger.models.Swagger
+import io.swagger.models.*
```
netme: I bet `ktlint` will complain here :-)

server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt
```diff
@@ -56,7 +54,23 @@ class UseProblemJsonRule : AbstractRule() {
         }
     }
 
+    private fun producesJson(swagger: Swagger, operation: Operation): Boolean {
```
netme: Optional: you can also do:

```kotlin
private fun producesJson(swagger: Swagger, operation: Operation): Boolean = 
    if (operation.produces == null || operation.produces.isEmpty()) {
    ...
```

server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt
```diff
@@ -56,7 +54,23 @@ class UseProblemJsonRule : AbstractRule() {
...
+        if (operation.produces == null || operation.produces.isEmpty()) {
```
netme: Sounds like not Kotlin :-) 
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/is-null-or-empty.html

server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt
```diff
@@ -56,7 +54,23 @@ class UseProblemJsonRule : AbstractRule() {
...
+            return swagger.produces != null && swagger.producesContainsJson()
```
netme: This `null` check can be also removed using something like:
```kotlin
swagger.produces.orEmpty().producesContainsJson()
```

server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt
```diff
@@ -56,7 +54,23 @@ class UseProblemJsonRule : AbstractRule() {
...
+        }
+    }
+
+    // support for application/json also with set charset e.g. "application/json; charset=utf-8"
+    private fun Swagger.producesContainsJson() =
+        produces.any { it.startsWith("application/json") }
+
+    // support for application/json also with set charset e.g. "application/json; charset=utf-8"
+    private fun Operation.producesContainsJson() =
+        produces.any { it.startsWith("application/json") }
```
netme: If `produces` is a string here and on line 67, why not to declare that on `String.isApplicationJson` level?

netme: And what is API produces `application/xml`?

server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt
cogitor: that would be funny cause this was IntelliJ changing automatically :D

server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt
cogitor: oh that's nice

server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt
netme: Yeah, intellij is bad with default settings, same for Java

server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt
cogitor: true.. but the isNullOrEmpty exists only for strings? I would need it for Lists.. now I could write it of course `private fun List.isNullOrEmpty() ...` :)

server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt
cogitor: that's much nicer!

server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt
cogitor: ah, well, it's actually a List.. and we're checking if any of the strings in the list is application/json or a derivative of the same ..

server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt
netme: Maybe `list.orEmpty().isEmpty()` will help you here.

server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt
netme: Ah, sorry, overlooked it. Then at least we can introduce this method once on the `List` object.

server/src/main/java/de/zalando/zally/rule/UseProblemJsonRule.kt
cogitor: yes, that seems very readable and avoids the null check 👍

cogitor:
@netme thanks for the feedback, I've some made changes based on it now

netme:
👍

netme:
👍

Issue #429, opened 2017-07-20
Simplify handling of JSONs in HTTP controllers


innokenty:
Main goal is to get rid of `JsonNode` objects, manual parsing of jsons and hardcoded fieldname strings.

netme:
Implemented in #428

innokenty:
No wait, there is a bit more to simplify in other controllers! Or should it be one PR per issue?

netme:
@innokenty sorry, didn't realise it :-) Nope, multiple PRs are OK. Could you please move it to back to "In Progress"?

innokenty:
Did already!

innokenty:
Implemented in #428 #430 #431 #432

PR #431, opened 2017-07-21
server: add DTO beans for /supported-rules response (#429)

innokenty:


server/src/main/java/de/zalando/zally/rule/Rule.kt
```diff
@@ -1,18 +1,16 @@
...
-import com.fasterxml.jackson.annotation.JsonProperty
 import de.zalando.zally.dto.Violation
 import de.zalando.zally.dto.ViolationType
 import io.swagger.models.Swagger
 
 interface Rule {
 
     val title: String
-    @get:JsonProperty("type") val violationType: ViolationType
+    val violationType: ViolationType
```
netme: That might be an issue since it's called `type` in API, not `violation_type`

maxim-tschumak: Why not just renaming the DOJOs property/fields names?

netme: I think with `type` it won't work since it's reserved keyword in Kotlin

maxim-tschumak:
:+1:

server/src/main/java/de/zalando/zally/rule/Rule.kt
innokenty: That's the point: `Rule` is not used anymore for API reasons. `RuleDTO` serves this purpose now and `Rule` is just an internal interface. Note that `RuleDTO` has this field called just `type`.

netme:
👍

PR #432, opened 2017-07-21
server: fork ViolationDTO class from Violation (#429)

innokenty:
### Change
Class `Violation` is now intended only for internal usage.
Class `ViolationDTO` represents the API structure.

### Reasoning
In my opinion this is a bad practice to unmarshal some internal classes into the response body and rely on `@JsonIgnore` and `@JsonProperty` annotations for API compliance. When those two are separate classes it makes the life much easier. As well as testing (you'll see soon)!

### P.S.
Second commit is moving a class **only**. Probably it's easier to review the first commit separately.


maxim-tschumak:
:+1:

Issue #433, opened 2017-07-21
Use DTOs in integration tests instead of JsonNode


innokenty:
The DTO beans were introduced in https://github.com/zalando-incubator/zally/issues/429 ✅

innokenty:
Implemented in https://github.com/zalando-incubator/zally/pull/434 & https://github.com/zalando-incubator/zally/pull/436

innokenty:
forgot some tests, reopening

PR #434, opened 2017-07-21
server: refact tests using DTOs instead of JsonNode (#433)

innokenty:
Please review refactoring 1 test with DTOs as a proof of concept.
Leave comments, but please don't merge: I plan to refactor more tests in this PR.

### DIFF
--- a/server/src/main/java/de/zalando/zally/dto/ApiDefinitionRequest.kt
+++ b/server/src/main/java/de/zalando/zally/dto/ApiDefinitionRequest.kt
@@ -3,7 +3,6 @@ package de.zalando.zally.dto
 import com.fasterxml.jackson.annotation.JsonRawValue
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize
 import de.zalando.zally.util.JsonRawValueDeserializer
-import org.apache.commons.io.IOUtils
 
 data class ApiDefinitionRequest (
 
@@ -13,15 +12,11 @@ data class ApiDefinitionRequest (
 
     var apiDefinitionUrl: String? = null
 ) {
+    /** for java invocations: it doesn't have overloaded constructors */
     companion object Factory {
 
-        // for java invocations: it doesn't have overloaded constructors
         fun fromJson(json: String) = ApiDefinitionRequest(json)
 
         fun fromUrl(url: String) = ApiDefinitionRequest(apiDefinitionUrl = url)
-
-        fun fromJsonResource(resource: String) = fromJson(
-                IOUtils.toString(ClassLoader.getSystemResourceAsStream(resource))
-        )
     }
 }
--- a/server/src/main/java/de/zalando/zally/dto/ApiDefinitionRequest.kt
+++ b/server/src/main/java/de/zalando/zally/dto/ApiDefinitionRequest.kt
@@ -3,6 +3,7 @@ package de.zalando.zally.dto
 import com.fasterxml.jackson.annotation.JsonRawValue
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize
 import de.zalando.zally.util.JsonRawValueDeserializer
+import org.apache.commons.io.IOUtils
 
 data class ApiDefinitionRequest (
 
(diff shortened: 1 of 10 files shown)


server/src/test/java/de/zalando/zally/apireview/RestApiBaseTest.java
```diff
@@ -36,10 +43,25 @@ public void cleanDatabase() {
         ObjectNode requestBody = new ObjectMapper().createObjectNode();
         requestBody.set("api_definition", body);
         return restTemplate.postForEntity(
-            getUrl(),
+            API_VIOLATIONS_URL,
```
innokenty: I plan to delete `getUrl()` method. It's actually never used, because this `sendRequest()` method is not abstract: it is intended only for `/api-violations` invocations.

In the bright refactored communist future the base test class will simply have different methods each invoking different endpoints. Urls will be stored in constants.

server/src/main/java/de/zalando/zally/dto/ApiDefinitionRequest.kt
```diff
@@ -11,4 +12,16 @@ data class ApiDefinitionRequest (
...
+    companion object Factory {
+
+        // for java invocations: it doesn't have overloaded constructors
+        fun fromJson(json: String) = ApiDefinitionRequest(json)
+
+        fun fromUrl(url: String) = ApiDefinitionRequest(apiDefinitionUrl = url)
+
+        fun fromJsonResource(resource: String) = fromJson(
+                IOUtils.toString(ClassLoader.getSystemResourceAsStream(resource))
+        )
```
netme: Would you mind to keep `fromJsonResource` logic in some test utility? We don't use it in normal use cases.

server/src/test/java/de/zalando/zally/apireview/RestApiBaseTest.java
```diff
@@ -36,10 +43,25 @@ public void cleanDatabase() {
...
+                API_VIOLATIONS_URL, request, ApiDefinitionResponse.class
+        );
+
+        assertThat(entity.getStatusCode()).isEqualTo(HttpStatus.OK);
+        return entity.getBody();
+    }
+
+    protected final ApiDefinitionResponse sendApiViolationsRequestFromResource(String resource) {
+        return sendApiViolationsRequest(
+                ApiDefinitionRequest.Factory.fromJsonResource(resource)
```
netme: So then we can call `fromJson` here instead of `fromJsonResource`

innokenty: Yeah, I agree it's better, thanks.

innokenty: Except I don't like just `fromJson` as a name: in the superclass we don't have the context of `ApiDefinitionRequest` anymore. So it's not clear W H A T this method actually makes out of the given json. And it's also not the json we provide, but a resource name.

So I would call it probably something like: `readApiDefinition(String resource)`

netme: Sure, that should work

innokenty:
Well, now it's definitely a CI bug:
```
The command "sudo apt-get install -qq bats" failed and exited with 100 during .
```

Is there a way to rebuild?

server/src/test/java/de/zalando/zally/util/ResourceUtil.java
```diff
@@ -0,0 +1,23 @@
...
+        return IOUtils.toString(ClassLoader.getSystemResourceAsStream(resourceName));
+    }
+
+    public static ApiDefinitionRequest readApiDefinition(String resourceName) throws IOException {
+        ApiDefinitionRequest request = new ApiDefinitionRequest();
+        String apiDefinitionJson = resourceToString(resourceName);
+        request.setApiDefinition(apiDefinitionJson);
+        return request;
+    }
+}
```
netme: That's cool

server/src/test/java/de/zalando/zally/util/ErrorResponse.kt
```diff
@@ -0,0 +1,7 @@
+package de.zalando.zally.util
+
+class ErrorResponse (
```
netme: `data class`?

netme:
👍

netme:
👍

Issue #265, opened 2017-04-12
Check is Swagger spec is compliant before running rules against it.


mkulak:
Some rules can give you false positives when swagger spec is itself invalid. Error messages in this case can be very confusing. 
I propose to validate spec and if it's not valid emit special violation (without trying to check it with other rules). Much like we do if spec cannot be parsed.

netme:
@mkulak that would be great. We can even go a bit further and do some swagger file linting like http://editor.swagger.io/ . Any ideas on the libs we can use for that?

fmueller:
If the swagger spec is not compliant, the parser should throw an exception - my assumption so far. Is this not the case?

mkulak:
No, this is not. See this issue: #261

fmueller:
Ok, then we need this. 👍

PR #435, opened 2017-07-21
#265 validate spec schema

akryvko:
Validation against Swagger schema http://swagger.io/v2/schema.json gives sometimes very useful hints. I tried to display all of them and enrich with details. That is why json-based rules introduced - they work basing on json/yaml trees and output multiple violations at once. I think they also might be useful for $ref checks and in cases when Swagger parser does not support something.


maxim-tschumak:
@akryvko please create a corresponding issue. In the meantime, I'll review the proposed changes.

server/src/main/java/de/zalando/zally/MetricsExporter.java
```diff
@@ -2,14 +2,13 @@
...
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Slf4jReporter;
+import java.util.concurrent.TimeUnit;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.context.event.ApplicationReadyEvent;
 import org.springframework.context.ApplicationListener;
 import org.springframework.stereotype.Component;
 
-import java.util.concurrent.TimeUnit;
```
maxim-tschumak: You have a lot of changes triggered by ordering of the imports in your IDE. Please change your IDE's behaviour according to https://pages.github.bus.zalan.do/team-architecture/rules-of-play/development/code-style.html#import-layout

JuntaCr: Issue exists - #265

JuntaCr: Pushed fixed import ordering, thanks.

server/src/main/java/de/zalando/zally/rule/InvalidApiSpecificationRule.kt
```diff
@@ -0,0 +1,19 @@
+package de.zalando.zally.rule
+
+import com.fasterxml.jackson.databind.JsonNode
+import de.zalando.zally.dto.ViolationType
+
+open class InvalidApiSpecificationRule : JsonRule() {
```
maxim-tschumak: What are the benefits of having this additional abstraction level? (`InvalidApiSpecificationRule` is super type of `InvalidApiSchemaRule`)

akryvko: Original idea: that is because we need some rule to inform that Swagger parsing failed - in RulesValidator. This one should not know about schema validation. But, probably it is fine. Ill try to use one class. Or we can introduce 2 rules with diff IDs and descriptions.

akryvko: removed InvalidApiSpecificationRule, seems not bad

maxim-tschumak:
:+1:

maxim-tschumak:
Thanks for your contribution!

PR #439, opened 2017-07-25
server: Check Schema - use local schema #265

### LINKED ISSUE #265: Check is Swagger spec is compliant before running rules against it.

akryvko:
As per discussion with @netme - we should handle situations when the online swagger json schema is not reachable. So it will be loaded from local resource (with dependant one) in case web download failed or configuration parameter 'swagger_schema_url' is not set  

Closes #265


netme:
Quick question: Will it be possible in the future to update offline schemas automated way?

netme:
👍

Issue #406, opened 2017-07-05
Disable Violation Pragmas


tfrauenstein:
feature enhancement idea:

When drafting an API, the API designer usually runs Zally multiple times.
He than checks Zallys violation output and decides how to handle each violation.
Sometimes API is changed as to avoid the violation, sometimes API is not changed on purpose -- often the case for SHOULD and MAY violations. 

Improve API designer convenience via support of pragma directives inside API definitions that disable checking of specific rules for specific OpenAPI elements. 
By this, API designers can avoid efforts for redundant checking of violations for multiple Zally runs during API evolution. 
It is also a must-feature for integration of Zally in automated CD or operation monitoring tooling.

mkulak:
@tfrauenstein do I understand it correctly that you propose to modify swagger file itself? If so, can you give an example how these pragmas may look like? As the file with pragmas still needs to stay compatible with OpenAPI format.

As an alternative I can propose to add additional field to json payload (and corresponding cmd arg). This way we could avoid API file modification.

tfrauenstein:
@mkulak Yes, I would prefer annotation inside OpenAPI definition to avoid multiple sources and correlation problems. As JSON does not support comments (IMO bad design choice, however) the only way I see is via specific fields. OpenAPI explicitly supports enrichments via vendor specific "x-" tags -- this would be the straight forward way to go for Zally pragma extensions.

netme:
Really nice idea, I think many linters are supporting that.
We can also go a bit further and let Zally user disable rules from command line tool configurations.

innokenty:
Hi guys, so I'm thinking of starting to implement this functionality. Let's discuss our visions of how this would look like.

We want to disable **specific rules** for **specific json nodes**, is that right? I checked the [OpenAPI docs](https://github.com/OAI/OpenAPI-Specification/blob/master/guidelines/EXTENSIONS.md), basically every element contains the possibility to add vendor-specific tags (as @tfrauenstein correctly pointed out).

As I see it, we should agree on some specific syntax to disable any set of rules, that can be inserted on any level. Is this also the goal as you see it?

In my mind it should look kinda like this:
```json
{
  "swagger" : "2.0",
  "x-zally": {
    "disable": [
      "SnakeCaseInPropNameRule",
      "LimitNumberOfSubresourcesRule"
    ]
  },
  "paths" : {
    "/products" : {
      "get" : {
        ...
        "x-zally": {
          "disable": [
[...]

PR #440, opened 2017-07-25
server: implement vendor extension to ignore rules by code (#406)

innokenty:
Example syntax:
```json
{
  "x-zally-ignore": [
    "M001",
    "M999"
  ],
  "swagger" : "2.0",
  ...
}
```

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/AbstractRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/AbstractRule.kt
@@ -2,6 +2,8 @@ package de.zalando.zally.rule
 
 abstract class AbstractRule : Rule {
 
+    val zallyIgnoreExtension = "x-zally-ignore"
+
     override val name: String = javaClass.simpleName
 
     override fun equals(other: Any?): Boolean {
--- a/server/src/main/java/de/zalando/zally/rule/JsonRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/JsonRule.kt
@@ -4,6 +4,13 @@ import com.fasterxml.jackson.databind.JsonNode
 
 abstract class JsonRule : AbstractRule() {
 
+    fun accepts(swagger: JsonNode): Boolean {
+        val ignoredCodes = swagger.get(zallyIgnoreExtension)
+        return ignoredCodes == null
+                || !ignoredCodes.isArray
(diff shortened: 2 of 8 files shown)


innokenty:
Asked review from all the people, engaged in the conversation here: https://github.com/zalando-incubator/zally/issues/406#issuecomment-317413093

maxim-tschumak:
@innokenty Have you looked at the `de.zalando.zally.rule.RulesPolicy`? I believe the implementation of these 2 functionalities should be combined:
1. configuring the server via the applicaiton configuration
2. configuring an api validation configuration via vendor extension field

Please check if it's possbile.

innokenty:
Application configuration is already implemented. That's the `RulesPolicy` class you mentioned, that's right.

I wanted to implement parsing this vendor extension there as well, but faced a problem: it depends on the input file. We have two types of rules: json and swagger, and they get the api definition in different formats. That's why there are two implementations.

And if you take a closer look at my Kotlin code snippets in `JsonRule.accepts` and `SwaggerRule.accepts` — you can see that they are slightly different. One works with `JsonNode` and another one with `Swagger` object.

That's my story 👆 
Is that what you meant? What do you think?

maxim-tschumak:
I see your point. I'm still not 100% ok with the design (static and dynamic configuration of which rules are exluded), but let's keep it like this for now and see how it develops. Thank you for the contribution!
:+1:

maxim-tschumak:
:+1:

innokenty:
Alright, thanks! I don't see a big conflict in there, those two configs serve absolutely different purposes: one disables a rule completely, because, I don't know.. It doesn't work properly. When the other config helps teams with not-standard compliant APIs still use zally with not much pain.

innokenty:
I don't like that the code is duplicated between json and swagger implementations, but that looks unavoidable.

The duplicated ways to exclude rules look totally ok to me.

maxim-tschumak:
There are two slightly different perspectives:
* **maintainer/operator/zally developer** doesn't want to remove a rule implementation if the respective guideline was removed from the [RESTful API Guidelines](http://zalando.github.io/restful-api-guidelines/).
* **zally user** with non-standard compliant APIs.


> The duplicated ways to exclude rules look totally ok to me.

Agree! :)

PR #441, opened 2017-07-25
server: UseProblemJsonRule rule should not NPE on specs w/o definitions #182

akryvko:
Spotted that on specs with references to /definitions ( $ref: '#/definitions/IdMappingResults') UseProblemJsonRule throws NPE if the /definitions section is absent or empty.

This spec is incorrect, ofc, and this situation is not covered by any of existing rules, including schema validation. So additionally we can consider implementing Reference-check rule, someth like Swagger editor does: 

 
>Resolver error at paths./identifier-types/{identifier_type}/source-ids/{source_identifier}.get.responses.200.schema.$ref
>Could not resolve reference: #/definitions/IdMappingResults
>Jump to line 47


maxim-tschumak:
Great!
:+1:

PR #443, opened 2017-07-25
server: Check Schema - download schema on build #265

### LINKED ISSUE #265: Check is Swagger spec is compliant before running rules against it.

akryvko:
Closes #265

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -26,7 +26,6 @@ an API check.
     ```
     The bootRun task is configured to run with 'dev' profile by default.
 
-
 ## Usage
 
 We recommend to use [Zally's CLI tool](../cli). Otherwise you can use any other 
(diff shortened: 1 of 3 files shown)


maxim-tschumak:
Is it possible to make the download optional?
I think it would be better to still have a default `swagger-schema` and `json-schema` definitions locally and update them if developer wants to (doesn't happen really often, right?)

I'm a long distance commuter and work sometimes in the train (no internet connection). I have all the needed dependencies in the local cache but the download of the schemas will fail.

server/src/main/resources/schemas/json-schema.json
```diff
@@ -1,150 +0,0 @@
-{
```
netme: Could you please also add these files to `.gitignore` then?

akryvko:
Then let's keep these files in git. It will be more safe for offline usage and more predictable. I've added only a task for updating them.

netme:
@akryvko sounds good. Could you please also update `README.md` and let's merge it!

netme:
👍

maxim-tschumak:
:+1:

maxim-tschumak:
:+1:

netme:
👍

PR #448, opened 2017-07-26
server: implement api parameter to ignore rules by code (#406)

### LINKED ISSUE #406: Disable Violation Pragmas

innokenty:
Closes #406

Example request:
```json
{
  "api_definition": {
    "never_gonna_give_you": "up!"
  },
  "ignore_rules": ["M001", "M999"]
}
```

### DIFF
--- a/server/src/main/java/de/zalando/zally/apireview/ApiViolationsController.java
+++ b/server/src/main/java/de/zalando/zally/apireview/ApiViolationsController.java
@@ -53,8 +53,8 @@ public class ApiViolationsController {
     public ApiDefinitionResponse validate(@RequestBody ApiDefinitionRequest request) {
         metricServices.increment("meter.api-reviews.requested");
 
-        final String apiDefinition = retrieveApiDefinition(request);
-        final List<Violation> violations = rulesValidator.validate(apiDefinition);
+        String apiDefinition = retrieveApiDefinition(request);
+        List<Violation> violations = rulesValidator.validate(apiDefinition, request.getIgnoreRules());
         apiReviewRepository.save(new ApiReview(request, apiDefinition, violations));
 
         ApiDefinitionResponse response = buildApiDefinitionResponse(violations);
--- a/server/src/main/java/de/zalando/zally/dto/ApiDefinitionRequest.kt
+++ b/server/src/main/java/de/zalando/zally/dto/ApiDefinitionRequest.kt
@@ -10,7 +10,9 @@ data class ApiDefinitionRequest (
     @JsonDeserialize(using = JsonRawValueDeserializer::class)
     var apiDefinition: String? = null,
 
-    var apiDefinitionUrl: String? = null
+    var apiDefinitionUrl: String? = null,
+
+    var ignoreRules: List<String>? = emptyList()
 ) {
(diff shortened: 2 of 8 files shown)


server/src/main/java/de/zalando/zally/apireview/ApiViolationsController.java
```diff
@@ -53,8 +53,8 @@ public ApiViolationsController(ApiValidator rulesValidator,
     public ApiDefinitionResponse validate(@RequestBody ApiDefinitionRequest request) {
         metricServices.increment("meter.api-reviews.requested");
 
-        final String apiDefinition = retrieveApiDefinition(request);
-        final List<Violation> violations = rulesValidator.validate(apiDefinition);
+        String apiDefinition = retrieveApiDefinition(request);
+        List<Violation> violations = rulesValidator.validate(apiDefinition, request.getIgnoreRules());
```
netme: Oh, why not `final` anymore?

innokenty: Dunno, kinda a bad practice in my opinion. I kept removing those finals in tests for several days now, thought you're ok with it.

innokenty: There are several discussions about this, by the way, take a look:

https://stackoverflow.com/questions/154314/when-should-one-use-final-for-method-parameters-and-local-variables

netme: Hehe, it's just another opinion on that, but I don't mind in general :-)

netme:
👍

server/src/main/java/de/zalando/zally/apireview/ApiViolationsController.java
innokenty: https://softwareengineering.stackexchange.com/questions/98691/excessive-use-final-keyword-in-java

innokenty: But in this given case both of those finals make no sense: strings are final anyway and the list is still mutable inside. Yeah, well, you protect from redirecting the pointers, is this really needed?

netme: Yeah, I don't want to go deep into that discussions, I think it will give at least some protection :-) 
But as mentioned earlier, I'm fine with removing them.

netme:
👍

netme:
@maxim-tschumak could you please also take a look ? (1 approval is not enough anymore 😸 )

maxim-tschumak:
:+1:

PR #449, opened 2017-07-26
server: move API Guidelines URL to config (#346)

### LINKED ISSUE #346: Move Restful API Guidelines base URL to config variable
While checking the existing rules I've found that Zalando Restful API Guidelines base URL is hard-coded in every rule. This leads to some inconsistencies, e.g. some of rules have `http` protocol and some `https`. 

Let's move base URL to application properties, so it will be consistent and we can change it any time.

innokenty:
Closes #346

### DIFF
--- a/server/src/test/java/de/zalando/zally/rule/RestSupportedRulesTest.java
+++ b/server/src/test/java/de/zalando/zally/rule/RestSupportedRulesTest.java
@@ -6,6 +6,7 @@ import de.zalando.zally.dto.ViolationType;
 import de.zalando.zally.util.ErrorResponse;
 import org.junit.Test;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.ResponseEntity;
 import org.springframework.test.context.TestPropertySource;
 
@@ -24,15 +25,22 @@ public class RestSupportedRulesTest extends RestApiBaseTest {
     @Autowired
     private List<Rule> implementedRules;
 
+    @Value("${zally.apiGuidelinesBaseUrl:something went wrong!}")
+    private String baseUrl;
+
     @Test
     public void testRulesCount() {
         assertThat(getSupportedRules().size()).isEqualTo(implementedRules.size());
     }
 
     @Test
-    public void testRulesFieldsAreNotNull() {
+    public void testRulesFields() {
         for (RuleDTO rule : getSupportedRules()) {
-            assertThat(rule.getCode()).as("Rule: " + rule.getTitle()).isNotNull();
+            assertThat(rule.getCode()).isNotEmpty();
+            assertThat(rule.getTitle()).isNotEmpty();
+            assertThat(rule.getType()).isNotNull();
+            assertThat(rule.getUrl()).isNotNull();
+            assertThat(rule.getUrl()).startsWith(baseUrl);
         }
     }
 
--- a/server/src/main/java/de/zalando/zally/dto/RuleDTO.kt
+++ b/server/src/main/java/de/zalando/zally/dto/RuleDTO.kt
@@ -1,7 +1,6 @@
 package de.zalando.zally.dto
 
 import com.fasterxml.jackson.annotation.JsonProperty
-import de.zalando.zally.rule.Rule
 
 data class RuleDTO(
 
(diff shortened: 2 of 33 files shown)


innokenty:
Any idea, why one test in cli-go started to fail? I don't see how it can be related: because there is no actual server running during tests, if I get it right.

akryvko:
Actually we are running server, see in travis file -  ./gradlew bootRun > /dev/null &
And do integration testing against most recent server codebase
Probably output slightly changed:
#   `[[ "$output" =~ "	http://zalando.github.io/restful-api-guidelines/hyper-media/Hypermedia.html#must-do-not-use-link-headers-with-json-entities" ]]' failed

innokenty:
Yes thanks, I see that the output has changed, but I have no idea why so far. I tried to find the config, with which the server is running during tests, but no success.

netme:
Time to get rid of Bats 😃  ( #108 )

netme:
@innokenty, as @akryvko mentioned it's running by Travis with default config, here are the exact lines: 
https://github.com/zalando-incubator/zally/blob/master/.travis.yml#L32-L36 

The failing test is doing `GET /supported-rules`, most probably the `ignoredRules` parameter of the default config is not taken into account:

https://travis-ci.org/zalando-incubator/zally/builds/257737109#L3214

innokenty:
Thanks! No, the problem is different and related to changes in this PR. I started the server locally, requested the `/supported-rules` endpoint and I see that the base url from the config is not picked up. No idea why. Just all the rules have urls like this:
```
"url" : "/naming/Naming.html#should-reserved-javascript-keywords-should-be-avoided",
```

Whereas during server integration tests — it works!

innokenty:
is `server/src/main/resources/rules-config.conf` even used?

innokenty:
should be, `application.yml:33` says:
```
rules-config-path: "rules-config.conf"
```

netme:
```
... and related to changes in this PR.
```
Definitely it is, the master is green :-)

Yeah, please check, I'm not that familiar with the config we are using. Can only say that it's working perfectly for the rules configuration, maybe it's initialized too late...

Why don't you move the base URL to `application.properties`?

innokenty:
> the master is green
yeah, I already checked that with this funny PR: https://github.com/zalando-incubator/zally/pull/450

innokenty:
I dunno, I can move. Just felt that it's kinda a rule config, but probably you're right

netme:
There is a badge for that btw:

innokenty:
Not sure what the issue was still. but I'm moving the property to `application.yml` and this way it works for all environments. Will force-push soon! 💪

innokenty:
PLZ don't merge yet!

innokenty:
And now it's ready. Also the changes got way more simple. Please review/merge!

server/src/test/java/de/zalando/zally/rule/RestSupportedRulesTest.java
```diff
@@ -24,15 +25,22 @@
     @Autowired
     private List<Rule> implementedRules;
 
+    @Value("${zally.apiGuidelinesBaseUrl:something went wrong!}")
```
netme: Hehe, I like the default value :-)

netme:
👍

maxim-tschumak:
:+1:

maxim-tschumak:
:+1:

server/src/test/java/de/zalando/zally/rule/RestSupportedRulesTest.java
innokenty: Yeah, I think it's quite a useful trick for the tests.

innokenty:
@maxim-tschumak I think you need to actually click "approve" now, so that it could be merged.

maxim-tschumak:
@innokenty No, there should be 2x :+1: after the last commit. @netme has to approve again :)

innokenty:
Says the opposite on the top:
screen shot 2017-07-27 at 10 11 41

innokenty:
kek

maxim-tschumak:
Zappr is the one who doesn't allow to merge if there are not enough approvals (zappr approval = :+1: in the comments after the last change) - I know, approval definition is somehow overloaded here :)

innokenty:
Ah, wow, so that's a feature!! I thought you're just thumbs-upping out of excitement 😆

netme:
👍

Issue #221, opened 2017-03-31
Add support for YAML in api_definition


netme:
Since `api_definition` is now accepting objects, currently we can only pass Swagger definition in JSON format.

To make it possible to send `yaml`, we need to introduce content negotiation functionality. 

We should support `application/yaml` and some other types.

PR #451, opened 2017-07-26
server: Add support of Yaml validation request #221

akryvko:
#221

### DIFF
--- a/server/src/test/resources/fixtures/api_violations_request.yaml
+++ b/server/src/test/resources/fixtures/api_violations_request.yaml
@@ -1,0 +1,46 @@
+api_definition:
+  swagger: '2.0'
+  info:
+    version: 1.0.0
+    title: Swagger Petstore
+    description: A sample API that uses a petstore as an example to demonstrate features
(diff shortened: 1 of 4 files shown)


server/src/test/resources/fixtures/api_violations_request.yaml
```diff
@@ -0,0 +1,46 @@
...
+            description: User is not authenticated
+          403:
+            description: User is not authorized
+          404:
+            description: Identifier is not found
+            schema:
+              $ref: '#definitions/Problem'
+        security:
+          - oauth2:
+            - 'cross-device-graph-service.read'
```
netme: Optional: could you please remove non-necessary empty lines from this file?

netme:
👍

server/src/test/resources/fixtures/api_violations_request.yaml
akryvko: they are added intentionally for line-numbers feature testing. Let's keep them for a while

maxim-tschumak:
:+1:

PR #456, opened 2017-07-28
cli-go: rewrite some tests.bats with go tests (#108)

innokenty:
#108 

Please review and merge if you like. Moar tests coming soon.


netme:
@innokenty that sounds really reasonable, however I have a couple of proposals.

Can we move integration tests to a separate go project?
We can automate starting up zally service, downloading binary builds of zally and testing them etc. 
(e.g. https://github.com/zalando-incubator/zally/blob/master/cli/integration_tests/tests.bats#L65)

What do you think?

innokenty:
Why would we write them in Go then? Could be bash. Or `tests.bats` 😉 
These test now invoke the app as a go function, but from the very outside entrypoint so to say. SO the benefit is you can debug it, or tweak some options. Makes sense? Otherwise let's just stay with `.bats` file because that's exactly what it does: invokes the compiled zally file with params and checks the console output, doesn't it?

netme:
@innokenty good point and I don't have an immediate answer. Let me talk to @maxim-tschumak first :-)

maxim-tschumak:
@innokenty @netme At some point, we will have integration tests for the front end as well. Does it make sense to have the same tech stack for them? I believe not, and that's why I prefer to have CLI integration tests in Golang as a part of CLI component.

innokenty:
@maxim-tschumak Do I understand you correct, that you agree with the approach proposed in this PR and you would like me to rewrite the rest of the `tests.bats` this way?

maxim-tschumak:
@innokenty Yes, I do :) and thanks!

netme:
Let's start with this approach then. +1

innokenty:
Then I'll continue, must be not a lot of work now 👍 
You can either merge, or wait for a couple more commits today.

netme:
Let's wait for couple of more commits, will be easier to see the bigger picture.

innokenty:
Squashed 👆 
Ready for merge ✅

netme:
screen shot 2017-07-31 at 18 37 41

He-he, GitHub has some troubles. Will check the PR once GH will be back online.

.travis.yml
```diff
@@ -36,7 +36,7 @@ script:
   - cd ..
 
   # Wait until Spring Boot will start
-  - while ! echo "GET / HTTP/1.0\n" | nc localhost 8080; do sleep 3; done
+  - while ! echo "GET / HTTP/1.0\n" | nc localhost 8080; do sleep 1; done
```
netme: Nice improvement

cli-go/zally/integration_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type callable func() error
+
+// https://stackoverflow.com/questions/10473800/in-go-how-do-i-capture-stdout-of-a-function-into-a-string
```
netme: 👍

netme:
👍

cli-go/zally/integration_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+		assert.Nil(t, e)
+	})
+}
+
+func TestIntegrationWithZallyApiDefinition(t *testing.T) {
+	t.Run("integrationWithZallyApiDefinition", func(t *testing.T) {
+		out, e := RunAppAndCaptureOutput([]string{"", "lint", "../../server/src/main/resources/api/zally-api.yaml"})
+
+		assert.Contains(t, out, "[33mSHOULD[0m [33mUse Specific HTTP Status Codes[0m")
```
maxim-tschumak: The color codes/commands is still a bit ugly. Make it sense to switch to [`assert.Regexp`](http://godoc.org/github.com/stretchr/testify/assert#Regexp)?

maxim-tschumak:
:+1:

cli-go/zally/integration_test.go
netme: @maxim-tschumak I think it's OK for now, we are matching exact colorised strings

innokenty: Yeah, I so far just rewrote the tests. Now it's easy to add any kind of extra checks and so on.

Issue #78, opened 2017-01-03
Find a way to provide line numbers in violation report


netme:
As we have discussed on API guild meeting, it would be nice to have line numbers in violations report. It will improve the usability of the tool.

### Possible ways of solving this issue
- Check the currently used library if it can be more verbose
- Find another library
- UGLY: try to match statements with lines

thomasabraham:
We directly get a JsonNode in APIViolationsController. All information about line numbers is lost at this point. Hence there is nothing we can do after this. 

Another thing to be considered is that we are converting yaml files to JSON in CLI and sending. Even for JSON we are doing a parsing in the CLI part which means the JSON send to server side don't have true line number information. So there is two part to solve this problem.

1. CLI should create/send request body containing a JSON that maintain the consistency of lines. What this means is that if line number 1 in the actual files becomes say line number 7 in the request body, then line number 2 in the actual file should be line number 8, 3 should be 9 and so on. This way server can start counting from the start of actual definition inside wrapper and get accurate line numbers. And for yaml, CLI have to do an intelligent parsing which maintain a similar consistency. 
2. In the server side we should accept string and parse it on our own (could use library). Here we'll create a JsonNode that represent the API definition as we have today. Also we'll create another HashMap where each of the path within the previous JsonNode is a key and the value will be the line number of that key in the actual file. Now if each rule can put the path(path inside json node)  where the violation happen then it can be converted to line numbers using this hashmap.

fmueller:
Thanks for bringing this discussion to the table.

I looked a bit more deeper into our code:
- swagger-parser library converts every input (json or yaml) into its own json object
- line and column numbers will be lost anyway by this
- what we need is a new parser library which would mean a lot of more effort

I would like to postpone further discussion as it seems to be a lot more effort to build an own parser, in contrast we can improve user experience a lot by other features. Navigating through violations by path names and entity names should be sufficient for now.

I will file an issue for swagger-parser with request for line numbers and column numbers.

alexkops:
"We" already have an own Swagger parser, don't know if that helps?
https://github.com/zalando/api-first-hand/tree/master/swagger-parser/src/main/scala/de/zalando/swagger

fmueller:
Thanks @alexkops :) Personally, I'd like to omit the effort of maintaining our own parser library and focus on delivering business value. Line numbers and column numbers are nice to have but no show stopper. Missing GitHub integration and error reporting are, for instance.

[...]

PR #458, opened 2017-07-28
#78 provide line numbers in violation report  ** NOT FOR MERGE ** ..yet

akryvko:
This is not completed PR, but a live prototype of how we can implement line numbers in the report.
Main idea: we can build association (map) between JsonNode/Swagger object and line number. 
We need: 
- Parser which knows location
- node/swagger object factory which knows what object we are creating 
(see JsonNodeFactoryLocationListener).  

In case of nested api_definition we have to build this association right in the deserializer, otherwise line numbers will be lost. (see ApiDefinitionDeserializer)

Each rule should provide list of json node pointers (pathes) to uniq identify place in spec (see AvoidTrailingSlashesRule). Having node->line map and node pointer it is easy to get line numbers.
(see: RestApiLineNumberTest)

Todo:
- make Zally understand Yaml requests [done]
- make this feature configurable: on/off
- make cli tools / ui client send specification preserving original formatting
-- stop converting yaml to json
-- send json as is, without transformations
- send line number in separated field (make path object: [path/line])
- implement returning of valid pointers in all rules
- consider building not jsonNode->line map, but SwaggerObject->line (but in this case JsonRules will be not covered). Or do both
- consider changing api: do not nest api body into api_definition parameters, send as is. Create another api path for Urls  

Issue: #78


netme:
Current status: had a chat with @akryvko, we will keep this PR on hold, still need to think about proper solution.

netme:
We have decided to close this PR, but let's keep the branch.

Issue #478, opened 2017-08-14
[GHE Integration] Trigger PR checks via webhook


netme:
### Acceptance criteria
1. REST API endpoint designed in #477 is implemented

netme:
Closed via #488

PR #488, opened 2017-08-17
GH-478 Zally Github Integration

vecmezoni:
Fixes https://github.com/zalando-incubator/zally/issues/483 https://github.com/zalando-incubator/zally/issues/481 https://github.com/zalando-incubator/zally/issues/480 https://github.com/zalando-incubator/zally/issues/479 https://github.com/zalando-incubator/zally/issues/477 https://github.com/zalando-incubator/zally/issues/482


github-integration/src/main/java/de/zalando/zally/integration/config/JacksonObjectMapperConfiguration.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    @Bean
+    @Primary
+    fun createObjectMapper(): ObjectMapper {
+        return configure(ObjectMapper())
+    }
+
+    @Bean
+    @Qualifier("yamlObjectMapper")
+    fun createYamlObjectMapper(): ObjectMapper {
```
maxim-tschumak: As far as I know, it should be enough to have one `objectMapper` (with `YAMLFactory`). This `objectMapper` will be able to marshal/unmarshal JSONs and YAMLs. Correct me if I'm wrong.

github-integration/src/main/java/de/zalando/zally/integration/zally/Configuration.kt
```diff
@@ -0,0 +1,11 @@
+package de.zalando.zally.integration.zally
+
+import com.fasterxml.jackson.annotation.JsonProperty
+
+class Configuration {
+    @JsonProperty("swagger_path")
+    lateinit var swaggerPath: String
```
maxim-tschumak: It's possible that there are multiple api definitions in one repository. This property should be an array/set. And please rename it to `apis` or `api_definitions`.

github-integration/src/main/java/de/zalando/zally/integration/zally/Configuration.kt
```diff
@@ -0,0 +1,11 @@
...
+
+    @JsonProperty("ignored_rules")
+    lateinit var ignoredRules: List<String>
```
maxim-tschumak: At some point, we decided against ignoring rules for the whole api definition. This should be only possible iff the user sets the rule id on a specific path. (see #495)

I think we should remove the `ignored_rules` setting.

github-integration/src/main/resources/api/zally-github-api.yaml
```diff
@@ -0,0 +1,85 @@
+swagger: '2.0'
```
maxim-tschumak: There is a tool for checking the API definitions ... ;)

Have you checked the API with Zally? There are some violations.

github-integration/build.gradle
```diff
@@ -0,0 +1,126 @@
...
+    compile("org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version")
+    compile("io.github.openfeign:feign-core:9.5.1")
+    compile("io.github.openfeign:feign-jackson:9.5.1")
+    compile("org.zalando.stups.build:tokens-k8s-spring-boot-starter:0.1.1") {
+        exclude group: "org.slf4j", module: "slf4j-simple"
+    }
+    compile('uk.co.datumedge:hamcrest-json:0.2')
+    compile('org.postgresql:postgresql:42.1.4')
+
+    testCompile('ru.yandex.qatools.embed:postgresql-embedded:2.4')
```
oleg: `hsqldb` doesn't support `jsonb`

github-integration/src/main/java/de/zalando/zally/integration/config/JacksonObjectMapperConfiguration.kt
oleg: I need to check this. would be nice simplification if ture! :+1: json/yaml configs were just copied from zally :smile_cat:

oleg: @maxim-tschumak `ObjectMapper` with `YamlFactory` can deserialize _yaml_ and _json_ but it will always serialize to _yaml_. In some cases we need to serialize to _json_
that's why we need both `ObjectMapper`s

github-integration/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM registry.opensource.zalan.do/stups/openjdk:latest
+
+LABEL MAINTAINER "http://zalando.github.io/"
+
+COPY build/libs/bark.jar /
+COPY scm-source.json /scm-source.json
```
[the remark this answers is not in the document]
vecmezoni: Are you sure? I've seen a check while it's deploying with the Jenkins pipeline to k8s

github-integration/src/main/java/de/zalando/zally/integration/zally/Configuration.kt
vecmezoni: Will be fixed in https://github.com/vecmezoni/zally/pull/19

github-integration/src/main/java/de/zalando/zally/integration/zally/Configuration.kt
vecmezoni: Will be fixed in https://github.com/vecmezoni/zally/pull/19

github-integration/src/main/resources/api/zally-github-api.yaml
vecmezoni: https://bark-test.playground.zalan.do/reports/28

netme:
👍

vecmezoni:
👍

maxim-tschumak:
:+1:

maxim-tschumak:
:+1:

netme:
👍

Issue #497, opened 2017-08-23
Switch to Rule Ids defined in the Zalando RESTful API Guidelines 


maxim-tschumak:
**As a** user of Zally and [Zalando RESTful API Guidelines](http://zalando.github.io/restful-api-guidelines/)
**I want** to have a proper linking between rule implementation (Zally) and specification (Guidelines)
**So that** there is consistency and referential integrity

**Acceptance Criteria**
* Rule IDs are in sync with [Zalando RESTful API Guidelines](http://zalando.github.io/restful-api-guidelines/)
* Links in rule descriptions and violations refer to the correct rule in the Guidelines
* Review statistics refer to the Guideline rules by the new id
* An overview of automated rules is created (simple mapping `RuleClassName` <-> `RuleId`)

Depends on https://github.com/zalando/restful-api-guidelines/issues/260 (already live).

netme:
Checked the current Zally implementation. We need to find a bit better approach here: sometimes we have few Zally rules behind 1 guidelines rule. Good example is [Rule 104](http://zalando.github.io/restful-api-guidelines/#104), it's checked in Zally in M004, M010 and M017. 

I will still open a preparation pull request which will fix the URLs to REST API guidelines and add the `guidelineCode` parameter and later we'll try to find a good approach here.

netme:
Also, some Zally rules don't have any REST API Guidelines rules behind. We need to find a good approach for that.

maschleg:
I would say one api guideline rule is represented as one ally rule. Maybe there are some situations where it make sense to split the implementation in two rules for better readability. For your example I would say M004 is actually rule [#105](http://zalando.github.io/restful-api-guidelines/#105) and M010 and M017 can be merged.

maxim-tschumak:
Hm.. What is the relation between zally rule implementation and guideline rules? Is it `1-to-n` or `n-to-m`?

netme:
@maxim-tschumak , so far I've seen `1-to-n` only

maxim-tschumak:
What do you think about going through the rules and try to merge them, so that the Zally rules are `1-to-1` to the guideline rules?

netme:
Discussed with @maxim-tschumak . Concept should be defined first.

roxspring:
[...]

PR #504, opened 2017-09-01
497: Switch to Rule IDs from Guidelines, Part 1

netme:
Related to #497


maschleg:
what about AC: "An overview of automated rules is created (simple mapping RuleClassName <-> RuleId)"?

server/src/main/java/de/zalando/zally/rule/EverySecondPathLevelParameterRule.kt
```diff
@@ -12,6 +12,8 @@ class EverySecondPathLevelParameterRule : SwaggerRule() {
     override val url = ""
     override val violationType = ViolationType.MUST
     override val code = "M005"
+    // TODO: Provide guidelines code
+    override val guidelinesCode = ""
```
maschleg: Maybe /#143 but it doesn't state that every second path level must be a path parameter.

server/src/main/java/de/zalando/zally/rule/AvoidTrailingSlashesRule.kt
```diff
@@ -8,9 +8,10 @@ import org.springframework.stereotype.Component
 @Component
 class AvoidTrailingSlashesRule : SwaggerRule() {
     override val title = "Avoid Trailing Slashes"
-    override val url = "/naming/Naming.html"
+    override val url = "/#136"
```
maschleg: url = "/#" + guidelinesCode

server/src/main/java/de/zalando/zally/rule/ExtractBasePathRule.kt
```diff
@@ -11,6 +11,8 @@ class ExtractBasePathRule : SwaggerRule() {
     override val url = "/naming/Naming.html"
     override val violationType = ViolationType.HINT
     override val code = "H001"
+    // TODO: Provide guidelines code
+    override val guidelinesCode = ""
```
maschleg: That's not a guideline rule. As discussed in https://github.com/zalando-incubator/zally/issues/459 we should label the rules as guideline and technical rules.

netme:
Hey @maschleg . This is only the first part of the ticket. More changes will come, including other ACs

server/src/main/java/de/zalando/zally/rule/ExtractBasePathRule.kt
netme: Oh, seems to be I've missed that part of conversation. I will try to find a good solution for that.

server/src/main/java/de/zalando/zally/rule/AvoidTrailingSlashesRule.kt
netme: What should we do in case of technical rules then?

server/src/main/java/de/zalando/zally/rule/EverySecondPathLevelParameterRule.kt
netme: Yes, sounds like a technical rule

server/src/main/java/de/zalando/zally/rule/AvoidTrailingSlashesRule.kt
maschleg: Maybe technical rules do not have an url. But let's discuss this when we do the rules classification.

maschleg:
👍

netme:
👍

Issue #507, opened 2017-09-06
Increase Test Coverage for CLI


netme:
Unfortunately we still have some dark parts of the projects, e.g. commands definition.

### Acceptance criteria
1. CLI test coverage is increased to 90%

PR #518, opened 2017-09-14
507: Increase Test Coverage for CLI tool, Part 1

netme:
Related to #507

### DIFF
--- a/cli/zally/utils/http_client_test.go
+++ b/cli/zally/utils/http_client_test.go
@@ -1,0 +1,52 @@
+package utils
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/zalando-incubator/zally/cli/zally/tests"
+)
+
+func TestListRules(t *testing.T) {
+
+	t.Run("succeed_when_server_responds_fast", func(t *testing.T) {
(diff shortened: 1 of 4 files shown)


netme:
Awesome, `codecov` already reports +4% :-)

cli/zally/utils/http_client_test.go
```diff
@@ -0,0 +1,52 @@
...
+		tests.AssertEquals(t, nil, err)
+
+		responseBody, _ := ioutil.ReadAll(response.Body)
+		tests.AssertEquals(t, "200 OK", response.Status)
+		tests.AssertEquals(t, "Hello", string(responseBody))
+	})
+
+	t.Run("fails_when_timeout_is_reached", func(t *testing.T) {
+		handler := func(w http.ResponseWriter, r *http.Request) {
+			time.Sleep(6 * time.Second)
```
maxim-tschumak: Does it make sense to add a function `http_client:SetTimeout(t)` to make it configurable? (and adding the possibility to make this test faster)

maxim-tschumak:
:+1:

cli/zally/utils/http_client_test.go
netme: Yeah, maybe we can move it to CLI properties, would be useful. I will create another PR later

netme:
👍

PR #549, opened 2017-10-05
Creating an overview of currently implemented checks

### LINKED ISSUE #548: Create an Overview of Zally Checks

maxim-tschumak:
Closes #548.


roxspring:
Doesn't this just end up creating yet another place for these cross references to be maintained and become out of sync?

I fully support the overview existing but wouldn't it be easy enough to auto-generate it from the rules?

maxim-tschumak:
@roxspring you are right! We have actually the `/supported-rules` endpoint

PR #552, opened 2017-10-05
Imposed stable ordering on /supported-rules

roxspring:
... stable and predictable but not user configurable.

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/SupportedRulesController.java
+++ b/server/src/main/java/de/zalando/zally/rule/SupportedRulesController.java
@@ -16,6 +16,7 @@ import org.springframework.web.bind.annotation.RestController;
 
 import java.util.List;
 
+import static java.util.Comparator.comparing;
 import static java.util.stream.Collectors.toList;
 
 @CrossOrigin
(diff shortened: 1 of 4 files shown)


server/src/main/java/de/zalando/zally/rule/SupportedRulesController.java
```diff
@@ -50,6 +51,9 @@ public RulesListDTO listSupportedRules(
             .filter(r -> filterByIsActive(r, isActiveFilter))
             .filter(r -> filterByType(r, typeFilter))
             .map(this::toDto)
+            .sorted(comparing(RuleDTO::getType)
```
maxim-tschumak: The ordering of the `ViolationType`s depends on the declaration order of the items in the Enum.
If it changes, the test will be still successful (production code and test code use the same `compareTo` method).
I think it makes sense to **make the ordering explicit** (`MUST`->`SHOULD`->`MAY`->`HINT`) since we want to order the types by their importance. An option would be to override `compareTo` method.
What do you think, @roxspring?

roxspring: Hmmm, I see your point @maxim-tschumak and only now notice that the deprecated COULD is out of order from the others!

How about I:

1. Fix the ordering of `COULD`
2. Add comments to ViolationType that the order is significant
3. Add a ViolationTypeTest to verify that the `MUST` comes first and `HINT` comes last

That way we get to build on the standard enum `compareTo()` implementation and don't have to change the tests as the enums get changed over time (unless unlucky enough to change first and last items).

maxim-tschumak:
:+1:

netme:
👍

maxim-tschumak:
:+1:

maxim-tschumak:
@netme please re-approve and merge

netme:
:+1:

netme:
Thanks @roxspring for your contributions!

Issue #569, opened 2017-10-26
Rule implementation & infrastructure mixed up in src tree


roxspring:
Currently the rule implementations and infrastructure classes are all mixed up in the same source package. This makes it a little difficult to reason about infrastructure changes such as the Rule Set work. Ultimately I expect the Rule Set work to require separating the implementation from infrastructure anyway into who know's which packages and source tree, but I'm wondering about moving the implementations away from infrastructure as a first step.

Any objections before I create the pull request?

Unless I hear better suggestions I'll put the implementations into a new de.zalando.zally.rule.impl sub-package.

roxspring:
Ended up splitting by ruleset in a separate PR

Issue #560, opened 2017-10-12
[Rule Sets] 3. Package Discovery


maxim-tschumak:
## Epic: [Rule Sets](https://github.com/zalando-incubator/zally/issues/551)

**Acceptance Criteria**
- "Zalando API Guidelines" and "Zally" Rule Sets are split up
- Server discovers packages from a configured directory
- Build is adjusted to include "Zalando API Guidelines" and "Zally" Rule Sets
- Only rules from "Zalando API Guidelines" Rule Set are active per default
- Ignore Rules by `rule-set-id`-`rule-id`

roxspring:
Any thoughts on how to implement this?

Do people implementing custom guidelines just implement a bunch of Rules/Checks and the java package is identified as the RuleSet? Would make it difficult to add further stuff to the RuleSet for example a base url for the guidelines it represents.

Or do people implementing a rule-set need to supply a RuleSet implementation with a list of the rules contained within it? Would be a shame to have to list rules explicitly when Spring does such nice discovery for us.

Or something else entirely?

maxim-tschumak:
From the technical perspective, [Java Platform Module System](http://openjdk.java.net/projects/jigsaw/spec/) (Java 9) can be a great candidate for implementing this feature. [This article](https://www.pluralsight.com/guides/java-and-j2ee/getting-started-with-the-java-platform-module-system#using-services-with-the-serviceloader) contains an example implementation.

Using `ServiceLoader` was initially proposed by @jbellmann. Thanks @maschleg for pointing out this new feature in Java9!

roxspring:
If we're going to achieve this, particularly if Zally and Zalando rulesets are to be examples to follow then I think we need to carve the gradle project up a little. To start the discussion I suggest we break the gradle project into a multi-module one along the following lines:
- zally-rules.jar = Rule APIs, annotations and core rule framework
- zally-ruleset-zally.jar = Zally specific rule implementations (depends on zally-rules.jar)
- zally-ruleset-zalando.jar = Zalando guidelines rule implementations (depends on zally-rules.jar)
- zally-server.jar = Spring boot server aka everything else (depends on all of the above, discovery mechanism to follow)

This seem like a reasonable approach @maxim-tschumak, @netme? Assuming agreement then I guess this is a warning to expect some big pull requests moving most of the source files around!

I've opened up a couple of minor pull requests (#646 and #647) to tidy up the package dependencies ahead of the main work.

beatkyo:
[...]

Issue #559, opened 2017-10-12
[Rule Sets] 2. Core Interfaces


maxim-tschumak:
## Epic: [Rule Sets](https://github.com/zalando-incubator/zally/issues/551)

**Acceptance Criteria**
- Interfaces are designed
  - Violation
  - Check
  - Rule
  - Rule Set
  - Rule Type
- Current Check implementations are refactored to implement new interfaces

netme:
> Current Check implementations are refactored to implement new interfaces

Does this mean that everything should go to Zalando REST API package?

maxim-tschumak:
> Does this mean that everything should go to Zalando REST API package?
I would move active checks only. The other ones can be part of the Zally package.

roxspring:
FYI I have some ideas and proof-of-concept java code around the core interfaces - with a view to also allowing #446 to be addressed too. The ultimate idea is to allow rules to be implemented along the following lines, with violations being automatically tied to the Check, Rule & contextual location within the document:

```java
@Component
@Rule(id=101, title="Use Specific HTTP Status Codes")
public class UseSpecificHTTPStatusCodesRule {

  @Check(severity=MUST, message="Status codes should be in the range 100-599")
  public void noInventedCodes(String status, Response response) {
    if(Integer.parseInt(status)<100 || Integer.parseInt(status)>599) {
      throw new Violation();
    }
  }

  @Check(severity=SHOULD, message="Obscure status codes should be described")
  public void describeObscureCodes(String status, Response response) {
    if(Integer.parseInt(status)/100 > 10) {
      if(response.description==null) {
        throw new Violation();
      }
    }
  }

  @Check(severity=MAY, message="Enforce something else entirely)
  public void somethingAboutPathParameters(PathParameter param) {
    //TODO
  }
}
```

Hopefully by announcing this it'll force me to finally polish up the code and post it tonight!

roxspring:
Definitely think that the existing .rule package should be split into at least abstract vs impl classes - very difficult to find your way around otherwise.

roxspring:
Created POC for a CheckValidator which invokes @Check annotated methods as in my earlier comment. Location within swagger definition is standardised and should easily support x-zally-ignore  at arbitrary points. Ultimately it should be able to serve as a replacement for both JsonRulesValidator and SwaggerRulesValidator. Implemented in Java initially but should translate to Kotlin easily enough.

Feedback welcome!

[...]

PR #570, opened 2017-10-27
Separated rule implementations from infrastructure

roxspring:
As suggested in #569 I've moved rule implementations into a distinct sub-package so that infrastructure interfaces and classes can be easily seen and refactored for the Rule Sets effort.


netme:
@roxspring thanks for your contribution. In general it's not a bad idea to separate the rule implementations, but the same time I'm not super convinced with `impl` package naming, it sounds a bit vintage to me and in general I somehow agree with answers in this Stackoverflow thread: https://stackoverflow.com/questions/1004980/should-interfaces-be-placed-in-a-separate-package 
Let's think a bit more carefully about this topic.

@maxim-tschumak , @maschleg what do you think about it?

roxspring:
I'm not a big fan of the `impl` sub package name myself but I view it as a lesser crime than 48 classes of mixed responsibilities in a single package. Would have probably preferred a `rule` sub package but `rule.rule` seemed ugly. Perhaps splitting down technology lines and using a `swagger` sub package would be better?

As mentioned in #569 I figure it's a transitory move anyway - presumably we'll soon end up with separate packages and/or source trees to achieve the '"Zalando API Guidelines" and "Zally" Rule Sets are split up' acceptance criteria needed for #560.

maxim-tschumak:
I also don't really like the `impl` name of the package. But I see a benefit of separating current rule/checks implementations and zally "core" functionality (interfaces, controller, configuration, etc.).
How do you feel about:
- Moving all non-rule-related stuff from `rule` package to new `core` package (I'm open to other names)
- Moving `util` package into the `rule` package.
This could be a good first step for #559.
@netme At some point, we have to separate interfaces declaration from the actual implementation of the rules.

roxspring:
Happy to split Zally and Zalando rules in to separate sub packages as part of #575 if that's preferred.

PR #571, opened 2017-10-28
Only show active rules

### LINKED ISSUE #566: "zally rules" command line displays all rules without distinguishing active vs inactive
When running "zally rules" on the command line I notice that all rules are displayed the same even when some of them have been disabled at the server. There are a number of obvious ways we could fix this and achieve parity with the GUI:

1. Display "status: active" or "status: inactive" along side each rule
2. Provide additional command line flags "--active" and "--inactive" (or "--active false") so that we can choose which subset to display
3. Just stop showing the inactive rules because the user doesn't need to know about them.

Option (3) would be my preferred choice, ideally doing the same in the GUI for precisely the same reason.

roxspring:
Closes #566 

Modified cli to always request only the active rules, never displaying inactive ones.
Modified web-ui to always request on the active rules and never display the active vs inactive status any more.
Left the server capable of filtering on is_active=true|false in case it's useful to somebody. Although I guess if there's no use case we need to support then maybe we ought to remove that too and simplify the codebase.

### DIFF
--- a/cli/zally/commands/supported_rules.go
+++ b/cli/zally/commands/supported_rules.go
@@ -61,9 +61,9 @@ func validateType(ruleType string) error {
 }
 
 func fetchRules(requestBuilder *utils.RequestBuilder, rulesType string) (*domain.Rules, error) {
-	uri := "/supported-rules"
+	uri := "/supported-rules?is_active=true"
 	if rulesType != "" {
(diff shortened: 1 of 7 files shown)


netme:
👍

maxim-tschumak:
:+1:

PR #574, opened 2017-11-06
Adding references to OpenAPI where appropriate

earth2marsh:
The specification was donated to the Open API Initiative, so even version 2.0 of what had been called the Swagger spec is now version 2.0 of the OpenAPI Specification. See: https://swagger.io/difference-between-swagger-and-openapi/

I didn't bother with any tests, because this change is docs only. LMK if you'd still like an issue filed.


maxim-tschumak:
:+1:

roxspring:
Makes sense to me. 

For bonus marks it would be nice to see `web-ui/src/client/app/containers/violations-tab.jsx` updated to reference the spec too.

maxim-tschumak:
:+1:

mfellner:
👍

PR #575, opened 2017-11-07
Added RuleSet interface

roxspring:
This is a first step in addressing #559. I'm hoping it's uncontroversial so that we can start on the next steps where things get interesting :)
- Added `RuleSet` interface along with `ZalandoRuleSet` and `ZallyRuleSet` implementations.
  - Very tempted to move Zalando and Zally rules into separate sub-packages but held off to avoid churn and controversy. Happy to add that change into the PR if supported.
- Associated existing rules with one of the `RuleSet` implementations
- Skipped merging `code` and `guideline` to `id` since that'll be better done once multiple `Check` implementations are supportable.
- Nothing changed affecting the REST API or it's clients.

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/AbstractRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/AbstractRule.kt
@@ -1,6 +1,8 @@
 package de.zalando.zally.rule
 
-abstract class AbstractRule : Rule {
+abstract class AbstractRule(ruleSet: RuleSet) : Rule {
+
+    override val ruleSet = ruleSet
 
     val zallyIgnoreExtension = "x-zally-ignore"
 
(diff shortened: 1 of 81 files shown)


maxim-tschumak:
@roxspring Great!
Feel free to move Zalando and Zally rules to different packages. Additionally, I would create `api` package for all internal interfaces (`Rule`, `RuleSet`).

roxspring:
No problem. Will make those changes and update the branch.

roxspring:
Changes made as agreed. Also added doc comments to Rule interface.

For now there are some dependencies between packages which don't seem entirely healthy but I'll work on tidying those in upcoming changes.

maxim-tschumak:
:+1:

maxim-tschumak:
:+1:

tkrop:
👍

Issue #558, opened 2017-10-12
[Rule Sets] 1. Ignore Rules by External Ids


maxim-tschumak:
## Epic: [Rule Sets](https://github.com/zalando-incubator/zally/issues/551)

**Acceptance Criteria**
- Ignore functionality is based on the external rule id's (currently `guidelinesCode` field of the `SwaggerRule`)
- Same set of checks are executed per default (as now)

roxspring:
Servers may be running with existing configurations that ignore e.g. the OAuth 2.0 rule referencing the internal id `M004`. Similarly swagger documents might be out there with a `x-zally-ignore: [M004]` declaration. Are we comfortable breaking these configurations as we transition to using the external id of `104` instead?

(we've discussed elsewhere that there are currently limited Zally servers in production making breakages more acceptable - I can confirm that we've not yet made heavy use of x-zally-ignore so a breakage would have very low impact for us in this case)

maxim-tschumak:
> Servers may be running with existing configurations that ignore e.g. the OAuth 2.0 rule referencing the internal id M004. Similarly swagger documents might be out there with a x-zally-ignore: [M004] declaration. Are we comfortable breaking these configurations as we transition to using the external id of 104 instead?

Yes. There is no heavy use of `x-zally-ignore` extension @ Zalando too.

PR #581, opened 2017-11-12
All rule-ignoring is performed by RulesPolicy

roxspring:
* ViolationsApiController now aware of the configured RulesPolicy
* ViolationsApiController passes a modified RulesPolicy to ApiValidator rather than the request-time list of ignores
* ApiValidators use request level RulesPolicy passed into the method rather than config level RulesPolicy in the constructor 
* RulesValidator now extracts x-zally-ignores from documents rather than the rule implementations themselves
* RulesValidator uses a RulesPolicy (modified with any x-zally-ignores additions) to decide which rules to use.

Should make #558 easier to implement and edges my #559 plans forward too.


roxspring:
Hmmm, looks like a cli test is failing... will take a look tomorrow night.

PR #586, opened 2017-11-14
Fixes #584

### LINKED ISSUE #584: Use Hyphenated HTTP Headers does not exclude ID
Zally parser does not ignore ID as specified in  http://zalando.github.io/restful-api-guidelines/#131

mblascoespar:
Fixes #584

### DIFF
--- a/server/src/main/java/de/zalando/zally/util/PatternUtil.java
+++ b/server/src/main/java/de/zalando/zally/util/PatternUtil.java
@@ -20,6 +20,7 @@ public class PatternUtil {
     private static final String PATTERN_APPLICATION_PROBLEM_JSON = "^application/(problem\\+)?json$";
     private static final String PATTERN_CUSTOM_WITH_VERSIONING = "^\\w+/[-+.\\w]+;v(ersion)?=\\d+$";
 
+
     public static boolean hasTrailingSlash(String input) {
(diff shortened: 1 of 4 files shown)


server/src/main/java/de/zalando/zally/util/PatternUtil.java
```diff
@@ -56,8 +57,8 @@ public static boolean isKebabCase(String input) {
         return input.matches(KEBAB_CASE_PATTERN);
     }
 
-    public static boolean isHyphenated(String input) {
-        return Arrays.stream(input.split("-")).allMatch(p -> p.matches("([A-Z][^A-Z ]*)|([^A-Z ]+)"));
+    public static boolean isHyphenatedExcludingID(String input) {
+        return Arrays.stream(input.split("-")).allMatch(p -> p.matches("([A-Z][^A-Z ]*)|([^A-Z ]+)|([I][D])"));
```
maxim-tschumak: `([A-Z][^A-Z ]*)|([^A-Z ]+)|ID` should be enough

mblascoespar: true ill change

mblascoespar:
becaus of sustantial rule changes this is not required anymore

PR #592, opened 2017-11-16
Rules can have multiple @Check methods instead of one validate() method

roxspring:
Done:
* Added a new `@Check` annotation to indicate check implementations
* Removed `SwaggerRule.validate(Swagger)` method
* Moved `SecureWithOAuth2Rule.validate(Swagger)` check implementation to `@Check DefineOAuthScopesRule. secureWithOAuth2(Swagger)`
* Moved `UsePasswordFlowWithOauth2Rule.validate(Swagger)` check implementation to `@Check DefineOAuthScopesRule.usePasswordFlowWithOauth2(Swagger)`
* So now `DefineOAuthScopesRule` implements 1 rule with 3 checks!

To Do (hopefully in this PR):
* Get some eyes on this - it's a fundamental change!
* Rename `DefineOAuthScopesRule` and tests to reflect the larger remit (Avoided for initial review since it might confuse the diff)


server/src/main/java/de/zalando/zally/rule/SwaggerRulesValidator.kt
```diff
@@ -32,8 +33,25 @@ class SwaggerRulesValidator(@Autowired rules: List<SwaggerRule>,
...
+
+            val violations = mutableListOf<Violation>()
+
+            rule::class.java.methods
+                    .filter { it.isAnnotationPresent(Check::class.java) }
+                    .filter { it.parameters.size == 1 }
+                    .filter { it.parameters[0].type.isAssignableFrom(root::class.java) }
+                    .forEach {
+                        it.isAccessible = true
+                        if (it.returnType == Violation::class.java) {
```
tkrop: Would use another filter here.

Maybe it is a good idea to move all filter conditions into a filter method.

server/src/main/java/de/zalando/zally/rule/SwaggerRulesValidator.kt
```diff
@@ -32,8 +33,25 @@ class SwaggerRulesValidator(@Autowired rules: List<SwaggerRule>,
...
+                            val violation = it.invoke(rule, root) as Violation?
+                            if (violation != null) {
+                                violations.add(violation)
+                            }
+                        }
+                    }
+
+            violations
```
tkrop: `flatMap` instead of `forEach` would be idiomatic, but I would stick to this version and wrap the list here **immutable**.

server/src/main/java/de/zalando/zally/rule/api/Check.kt
```diff
@@ -0,0 +1,5 @@
+package de.zalando.zally.rule.api
+
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.FUNCTION)
+annotation class Check
```
tkrop: Should contain description text (see https://github.com/zalando-incubator/zally/issues/551).

server/src/main/java/de/zalando/zally/rule/SwaggerRulesValidator.kt
```diff
@@ -32,8 +33,25 @@ class SwaggerRulesValidator(@Autowired rules: List<SwaggerRule>,
...
+        return { rule: SwaggerRule ->
+
+            val violations = mutableListOf<Violation>()
+
+            rule::class.java.methods
+                    .filter { it.isAnnotationPresent(Check::class.java) }
+                    .filter { it.parameters.size == 1 }
+                    .filter { it.parameters[0].type.isAssignableFrom(root::class.java) }
+                    .forEach {
+                        it.isAccessible = true
```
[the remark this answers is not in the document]
roxspring: Yeah - I was surprised that I needed this too - added it at the last minute in order to allow the public  check method of a private static inner class - one/both of the RestApiTestConfiguration inner classes. I guess making the static inner class public will resolve the issue more cleanly.

server/src/main/java/de/zalando/zally/rule/SwaggerRulesValidator.kt
roxspring: Maybe. As a next (or nearly next) step I was planning to generalize further so that JsonRules can also be processed which would need a different variation of the filter and invoker, e.g:
```
else if (it.returnType isAssignableFrom Iterable<Violation>) { // pseudo code alert!
   violations.addAll(it.invoke(rule, args) as Iterable<Violation>
}
```
But you're right optimizing one step at a time might be best.

server/src/main/java/de/zalando/zally/rule/SwaggerRulesValidator.kt
roxspring: Will take a look.

server/src/main/java/de/zalando/zally/rule/api/Check.kt
roxspring: Agreed but wanted to get the mechanics in place first. You think it's worth delaying this PR until the annotation interface is finalized?

I'd like to also include a ViolationType too since it seems that MUST rules often also contain some SHOULD clauses, or the check implementation isn't as reliable and so needs to be downgraded.

server/src/main/java/de/zalando/zally/rule/SwaggerRule.kt
```diff
@@ -1,10 +1,5 @@
...
 
 import de.zalando.zally.rule.api.RuleSet
-import io.swagger.models.Swagger
 
-abstract class SwaggerRule(ruleSet: RuleSet) : AbstractRule(ruleSet) {
-
-    abstract fun validate(swagger: Swagger): Violation?
-
-}
+abstract class SwaggerRule(ruleSet: RuleSet) : AbstractRule(ruleSet)
```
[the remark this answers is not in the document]
roxspring: > This is now only a marker class/interface

Yup `SwaggerRule` brings little to the party right now, and if we apply the same approach to `JsonRule` then they can both be dropped. Until then SwaggerRule at least gives us a mechanism to request the Rules without the JsonRules.

> Should be substituted against an annotation @Rule([set=<String>,] id=<int>, title=<String>, type=<Type>, link=<String>)

Yes, I like the annotation approach and even suggested it elsewhere. What I'm not totally clear on is how to get Spring to autowire a "list of all components with the Rule annotation".

> I think that it is even not necessary to create rule objects, if methods are static. A rule set is a virtual container of rules, that may even not be reflected in the annotation, but by a context annotation or file provided in package-info.java or higher level file.

I can see this working out longer term but I'm trying to keep to incremental changes at this stage and I think what you suggest here is too big for this step.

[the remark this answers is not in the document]
roxspring: > Yes, I like the annotation approach and even suggested it elsewhere. What I'm not totally clear on is how to get Spring to autowire a "list of all components with the Rule annotation".

OK - actually that looks pretty simple - give the RulesValiatorConfiguration and autowired Application Context and use `applicationContext.getBeansWithAnnotation(Check::class.java)`.

Still, addressing that in this PR introduces further complications in Violation construction (since we can't directly access the Rule annotation properties from within the check implementation where the Violations are created). I suggest we park this for now and then will open subsequent PRs to:
- Manipulate Violation creation such that minimal information is needed within the check method, then the RuleValiators associate information from the check annotation and rule
- Drop the Rule interface in favour of a Rule annotation as discussed and modify the RuleValidators to pull context from the annotation rather than instance.

Reasonable?

server/src/main/java/de/zalando/zally/rule/SwaggerRulesValidator.kt
roxspring: Ended up simplifying this and just invoking the method, then process whatever the result is. See the new `RulesValidator.invoke(..)` method.

server/src/main/java/de/zalando/zally/rule/SwaggerRulesValidator.kt
roxspring: Ended up switching to a flatMap and it's definitely neater.

roxspring:
Done:
- Generalised `@Check` support to RulesValidator and made more idiomatic
- Removed no longer needed JsonRule and SwaggerRule

To Do (this PR):
- More review.
- Rename `DefineOAuthScopesRule` and tests to reflect the larger remit
- Improved reflection error handling / logging?
- Improved test coverage?

To Do (separate PR):
- Switch Rule from implemented interface to annotation
- Finalize RuleSet/Rule/Check properties

server/src/main/java/de/zalando/zally/rule/api/Check.kt
maxim-tschumak: > I'd like to also include a ViolationType too since it seems that MUST rules often also contain some SHOULD clauses [..]

:+1:

server/src/main/java/de/zalando/zally/rule/SwaggerRule.kt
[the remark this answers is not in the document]
tkrop: Very good 👍

server/src/main/java/de/zalando/zally/rule/api/Check.kt
tkrop: Could make sense to have a (violation) *type* on a Check too. You are right, that must rules have sub clauses that contain only should. The big question than is how we represent this duplicate violation types in the user response :).

maxim-tschumak: > The big question than is how we represent this duplicate violation types in the user response :)

I see two valid options:

1. Strong coupling of Violation to a Rule (`Violation.type == Rule.type`)
2. Weaker coupling between Violation and Rule: violation type (or, "severity") is derived from the check types (highest wins)

I prefer the second option.
This also plays well with the separation of internal and external APIs: Rules and Checks implement internal APIs. Each Rule and Check instance and has a _static_ type (as defined in the guidelines). A Violation (internal and external API) is a result of linting which has a reference to a rule and one or more checks. The type of the Violation can be derived from the check types in this case. The reference information to Zally Rule and Check is still provided.

mkulak:
What's the motivation for this change?

roxspring:
@mkulak The motivation is to allow multiple related checks to exist within a single rule implementation. This makes it easier to represent guidelines that consist of some MUST statements along side some SHOULD statements. Similarly it should allow lots of similar clauses to be cleanly separated while staying close to code that they share - e.g. "Use HTTP Methods Correctly" rule might benefit from a separate check per method, or "Use Specific HTTP Status Codes" might benefit from a separate check per status code family.

I think this change is ready. Clearly there are follow up changes to make (noted below) but I'm hoping to get agreement that they can be left for a separate PR. I'm going to try and resist further changes in this PR unless specifically asked to by members. 

Done:
- Added a new `@Check` annotation to indicate check implementations
- Removed `SwaggerRule.validate(Swagger)` method
- Moved `SecureWithOAuth2Rule.validate(Swagger)` check implementation to `@Check DefineOAuthScopesRule. secureWithOAuth2(Swagger)`
- Moved `UsePasswordFlowWithOauth2Rule.validate(Swagger)` check implementation to `@Check DefineOAuthScopesRule.usePasswordFlowWithOauth2(Swagger)`
- So now `DefineOAuthScopesRule` implements 1 rule with 3 checks!
- Generalised `@Check` support to RulesValidator and made more idiomatic
- Removed no longer needed JsonRule and SwaggerRule classes
- Rename `DefineOAuthScopesRule` and tests to reflect the larger remit
- Improved logging

To Do (separate PR):
- Switch Rule from implemented interface to annotation
- Finalize RuleSet/Rule/Check properties
- Remove the random dependencies on `InvalidApiSchemaRule.getGeneralViolation()`

maxim-tschumak:
Great! @netme please also have a look

maxim-tschumak:
:+1:

server/src/main/java/de/zalando/zally/rule/zalando/SecureWithOAuth2Rule.kt
```diff
@@ -1,29 +1,96 @@
...
+    @Check
+    fun checkSecurityDefinitions(swagger: Swagger): Violation? {
         val hasOAuth = swagger.securityDefinitions.orEmpty().values.any { it.type?.toLowerCase() == "oauth2" }
         val containsHttpScheme = swagger.schemes.orEmpty().contains(Scheme.HTTP)
         return if (!hasOAuth)
-            Violation(this, title, "No OAuth2 security definitions found", violationType, url, emptyList())
+            Violation(this, "Secure Endpoints with OAuth 2.0", "No OAuth2 security definitions found", violationType, url, emptyList())
         else if (containsHttpScheme)
-            Violation(this, title, "OAuth2 should be only used together with https", violationType, url, emptyList())
+            Violation(this, "Secure Endpoints with OAuth 2.0", "OAuth2 should be only used together with https", violationType, url, emptyList())
```
netme: Minor: this can be moved to a constant

netme:
👍

server/src/main/java/de/zalando/zally/rule/zalando/SecureWithOAuth2Rule.kt
roxspring: I plan to move these to properties of the `@Check` annotation so that they're no longer floating around the code so much - I suspect that'll remove the need for a constant but happy to discuss further at that point.

(Couldn't make the change yet since Violation construction wouldn't have access to the annotation properties. My plan is to create Violations with just the affected locations and maybe a detailed message, then have the RuleValidators pull in contextual titles/descriptions/severities from the Check / Rule / RuleSet. Felt like I had to cut the changes off somewhere to keep the pull request somewhere near manageable though!)

server/src/main/java/de/zalando/zally/rule/RulesValidator.kt
```diff
@@ -24,5 +32,27 @@ abstract class RulesValidator<RuleT, RootT>(val rules: List<RuleT>, val invalidA
...
     abstract fun ignores(root: RootT): List<String>
 
-    abstract fun validator(root: RootT): (RuleT) -> Iterable<Violation>
+    private fun validator(root: Any): (RuleT) -> Iterable<Violation> {
+        return { rule: RuleT ->
+            log.debug("validating ${rule.name} rule")
+            rule::class.java.methods
+                    .filter { it.isAnnotationPresent(Check::class.java) }
+                    .filter { it.parameters.size == 1 }
+                    .filter { it.parameters[0].type.isAssignableFrom(root::class.java) }
```
tkrop: I still would suggest to refactor all filters to single filter calling a private filter function for the time being.

tkrop:
👍

server/src/main/java/de/zalando/zally/rule/RulesValidator.kt
roxspring: Oops - managed to forget that action before. Should be able to make the change this evening.

Worth delaying this PR for? or can it wait until the next?

tkrop: Both would be okay. Otherwise I would not have given my consent 😸

PR #595, opened 2017-11-22
Simplified Rule Interface

### LINKED ISSUE #558: [Rule Sets] 1. Ignore Rules by External Ids

roxspring:
Simplified Rule interface to remove overlapping and unnecessary properties.

Done:
- Replaced `rule.code` and `rule.guidelinesCode` with unified `rule.id`
  - Updated test expectations to match
  - Should resolve #558
- RuleSets now own url construction rather than each rule own a url fragment duplicating the id
  - `application.yml` no longer contains reference to base urls.
- Dropped `rule.name` from interface, but left `abstractRule.name` since it's useful for implementations.
- `ZallyRuleSet` now references `rules.md` documentation at GitHub.

To Do (separate PR):
- Switch Rule from implemented interface to annotation
- Finalize RuleSet/Rule/Check properties
- Remove the random dependencies on `InvalidApiSchemaRule.getGeneralViolation()`


server/rules.md
```diff
@@ -0,0 +1,23 @@
+# Zally Rule Sets
+
+Zally comes with a number of Rule Sets which can be selected used. The built in rule sets are documented below.
```
maxim-tschumak: `selected used` -> `selected and used`

server/src/main/java/de/zalando/zally/rule/zalando/ZalandoRuleSet.kt
```diff
@@ -1,12 +1,16 @@
...
 import org.springframework.stereotype.Component
 import java.net.URI
 
 @Component
-class ZalandoRuleSet : de.zalando.zally.rule.api.RuleSet {
+class ZalandoRuleSet : RuleSet {
     override val id = javaClass.simpleName
     override val title = "Zalando RESTful API and Event Scheme Guidelines"
-    override val url = URI.create("https://zalando.github.io/restful-api-guidelines/")
+    override val url = URI.create("http://zalando.github.io/restful-api-guidelines/")
```
maxim-tschumak: Please change back to HTTPs link

roxspring: The rule urls used to be constructed using baseurl from application.yaml which happened to be configured with http:// so without the change above lots of cli integration test expectations need to be updated. Clearly I opted for the lazy option of changing the ruleset but happy to Do The Right Thing instead.

maxim-tschumak: > Clearly I opted for the lazy option of changing the ruleset but happy to Do The Right Thing instead.

I know it's Friday, but please do The Right Thing :)

roxspring: Done and done.

maxim-tschumak:
:+1:

roxspring:
Any chance of a review @netme, @maschleg?

Thanks!

maxim-tschumak:
:+1:

server/src/main/java/de/zalando/zally/apireview/ApiViolationsController.java
```diff
@@ -106,7 +102,7 @@ private ViolationDTO toDto(Violation violation) {
             violation.getTitle(),
             violation.getDescription(),
             violation.getViolationType(),
-            baseUrl + violation.getRuleLink(),
+            violation.getRule().getRuleSet().url(violation.getRule()).toString(),
```
maschleg: As AbstractRule has the reference to the RuleSet this interface could be cleaner: 
```
violation.getRule().getUrl().toString()
```

roxspring: It could... which would imply that the Rule interface has a `getUrl()` method declared again since we shouldn't rely on all rules extending AbstractRule. I think this is would be a mistake though, on two grounds:

1. I think the ownership of the urls should be with the RuleSet and not with the Rule. It feels wrong to me that rules should be able to ignore/replace the base url where a collection of rules is documented.
2. I'm intending soon to switch Rule over from an interface to an annotation soon, at which point there's no place to define a `getUrl()` since annotations only provide metadata, not behaviour. 

Perhaps as an alternative we could move this expression into a `violation.url()` method?

maschleg: A resource should be the owner of its url. So I do not have worries that a Rule could overwrite the base url. However, as we do not have a use case for it and you plan to change Rule to an annotation I am fine with that implementation.

maschleg:
👍

server/src/main/java/de/zalando/zally/apireview/ApiViolationsController.java
roxspring: Thanks!

PR #601, opened 2017-11-28
Violation related tests focus on description and paths properties

roxspring:
* Corrected QueryParameterCollectionFormatRule description
* Rule related tests avoid focusing on violation type
* TestRules now construct Violations using rule title

(All in preparation for simplifying Violation creation to just using description and path properties.)

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/zalando/QueryParameterCollectionFormatRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/zalando/QueryParameterCollectionFormatRule.kt
@@ -15,7 +15,7 @@ class QueryParameterCollectionFormatRule(@Autowired ruleSet: ZalandoRuleSet) : A
     override val violationType = ViolationType.SHOULD
     override val id = "154"
     val formatsAllowed = listOf("csv", "multi")
-    val violationDescription = "CollectionFormat should be one of: {formatsAllowed}"
+    val violationDescription = "CollectionFormat should be one of: $formatsAllowed"
 
     @Check
     fun validate(swagger: Swagger): Violation? {
--- a/server/src/test/java/de/zalando/zally/apireview/RestApiIgnoreRulesTest.java
+++ b/server/src/test/java/de/zalando/zally/apireview/RestApiIgnoreRulesTest.java
@@ -2,6 +2,7 @@ package de.zalando.zally.apireview;
 
 import de.zalando.zally.dto.ApiDefinitionResponse;
 import de.zalando.zally.dto.ViolationDTO;
+import org.assertj.core.groups.Tuple;
 import org.junit.Test;
 import org.springframework.test.context.TestPropertySource;
 
@@ -10,6 +11,7 @@ import java.util.List;
 import java.util.Map;
 
 import static de.zalando.zally.util.ResourceUtil.readApiDefinition;
+import static java.util.Collections.emptyList;
 import static org.assertj.core.api.Assertions.assertThat;
 
 @TestPropertySource(properties = "zally.ignoreRules=H999")
@@ -20,9 +22,12 @@ public class RestApiIgnoreRulesTest extends RestApiBaseTest {
         ApiDefinitionResponse response = sendApiDefinition(readApiDefinition("fixtures/api_spp.json"));
 
         List<ViolationDTO> violations = response.getViolations();
-        assertThat(violations).hasSize(2);
-        assertThat(violations.get(0).getTitle()).isEqualTo("dummy1");
-        assertThat(violations.get(1).getTitle()).isEqualTo("schema");
+        assertThat(violations)
+                .extracting("description","paths")
+                .containsExactly(
+                        new Tuple("dummy1", emptyList()),
+                        new Tuple("schema incorrect", emptyList())
+                );
 
         Map<String, Integer> count = response.getViolationsCount();
         assertThat(count.get("must")).isEqualTo(2);
--- a/server/src/test/java/de/zalando/zally/rule/RestSupportedRulesTest.java
+++ b/server/src/test/java/de/zalando/zally/rule/RestSupportedRulesTest.java
@@ -12,7 +12,6 @@ import org.springframework.test.context.TestPropertySource;
 
 import java.util.Arrays;
 import java.util.List;
-import java.util.stream.Collectors;
 
 import static java.util.stream.Collectors.joining;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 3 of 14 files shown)


server/src/main/java/de/zalando/zally/rule/zalando/QueryParameterCollectionFormatRule.kt
```diff
@@ -15,7 +15,7 @@ class QueryParameterCollectionFormatRule(@Autowired ruleSet: ZalandoRuleSet) : A
     override val violationType = ViolationType.SHOULD
     override val id = "154"
     val formatsAllowed = listOf("csv", "multi")
-    val violationDescription = "CollectionFormat should be one of: {formatsAllowed}"
+    val violationDescription = "CollectionFormat should be one of: $formatsAllowed"
```
roxspring: I presume this was supposed to be templated in all along!?

server/src/test/java/de/zalando/zally/rule/RestSupportedRulesTest.java
```diff
@@ -63,10 +62,26 @@ public void testIsActiveFlag() {
...
     public void testFilterByType() {
-        for (ViolationType ruleType : ViolationType.values()) {
-            assertFilteredByRuleType(ruleType.toString());
-            assertFilteredByRuleType(ruleType.toString().toLowerCase());
-        }
+
+        final int expectedCount = implementedRules.size();
+
+        final int mustCount = getSupportedRules("MuSt", null).size();
+        assertThat(mustCount).isLessThan(expectedCount);
```
maxim-tschumak: I don't see a value of such test assertions.

server/src/test/java/de/zalando/zally/rule/zalando/NoVersionInUriRuleTest.kt
```diff
@@ -10,35 +8,36 @@ class NoVersionInUriRuleTest {
...
     fun returnsViolationsWhenVersionIsInTheEndOfBasePath() {
         val swagger = Swagger().apply { basePath = "/api/v1" }
-        assertThat(rule.validate(swagger)).isEqualTo(expectedViolation)
+        assertThat(rule.validate(swagger))
+                .hasFieldOrPropertyWithValue("description", "basePath attribute contains version number")
+                .hasFieldOrPropertyWithValue("paths", listOf<String>())
     }
 
     @Test
     fun returnsViolationsWhenVersionIsBig() {
```
maxim-tschumak: `returnsViolationsWhenVersionIsBig` -> `returnsViolationsWhenBasePathStartsWithVersion`

server/src/test/java/de/zalando/zally/rule/zally/NoProtocolInHostRuleTest.kt
```diff
@@ -30,14 +30,18 @@ class NoProtocolInHostRuleTest {
     fun negativeCaseHttp() {
         val swagger = Swagger().apply { host = "http://google.com" }
         val res = rule.validate(swagger)
-        assertThat(res?.copy(description = "")).isEqualTo(expectedViolation)
+        assertThat(res)
+                .hasFieldOrPropertyWithValue("description", "Information about protocol should be placed in schema. Current host value 'http://google.com' violates this rule")
```
maxim-tschumak: Hm.. Why using `hasFieldOrPropertyWithValue` instead of relying on type system (and not referencing to a field by `string` parameter)?
I would rather have an expected object and (deep) equality check or checking each field separately.

server/src/test/java/de/zalando/zally/rule/RestSupportedRulesTest.java
roxspring: You're right - asserting that the sum of the subsets is equal covers all we need to test. Will update.

server/src/test/java/de/zalando/zally/apireview/RestApiIgnoreRulesTest.java
```diff
@@ -20,9 +22,12 @@ public void shouldIgnoreWithProperty() throws IOException {
         ApiDefinitionResponse response = sendApiDefinition(readApiDefinition("fixtures/api_spp.json"));
 
         List<ViolationDTO> violations = response.getViolations();
-        assertThat(violations).hasSize(2);
-        assertThat(violations.get(0).getTitle()).isEqualTo("dummy1");
-        assertThat(violations.get(1).getTitle()).isEqualTo("schema");
+        assertThat(violations)
+                .extracting("description","paths")
+                .containsExactly(
```
maxim-tschumak: So I don't see a value in refactoring the tests in this way:

- `containsExactly` checks also the ordering of the elements (same as `.get(0)`)
- no explicit size check (harder to reproduce from test output)

Please explain what is your intention

server/src/test/java/de/zalando/zally/rule/zally/NoProtocolInHostRuleTest.kt
roxspring: The rule tests are about asserting the behaviour of the rule, making sure that the correct description and paths are built given the scenario under test. Verifying the violationType and rule title (e.g. via a deep equals) make these behaviour changes brittle and overly sensitive (e.g. would break when reclassifying test from SHOULD to MUST). Additionally asserting with a deep equals can clearly indicate _that_ things are different but does a poor job indicating _what_ has changed.  Admittedly it might be nicer to use a type safe method reference rather than name the field but I don't see that option in Assert4j...

server/src/test/java/de/zalando/zally/apireview/RestApiIgnoreRulesTest.java
roxspring: Primary goal for changing is to remove the dependence on the title property which I indent to remove next (change already waiting in the wings). Secondly I find it much easier to understand tests, and reason about failing tests, when there's a single failing assertion which clearly documents all the reasons for test to fail.

Don't quite get your "no explicit size check (harder to reproduce from test output)" statement. Admittedly the  size check is now implicit but I don't see the downside of that. It's much more important to show in the expectations what N items you expected to see rather than that there were N of them. Additionally checking the size explicitly just adds room for the expectations to become out of sync - e.g. it would have been trivial to make a change to the previous version of the test such we now expected 3 violations but forget to explicitly test the 3rd violations content; if the implementation happened to return the wrong 3rd violation then the test would not notice and provide a false sense of security.

server/src/test/java/de/zalando/zally/rule/zally/NoProtocolInHostRuleTest.kt
maxim-tschumak: We test a Rule here. A Rule has a well-defined interface: it returns a Violation. A violation has typed properties. We assert the values of the properties to test the right behavior of the Rule. Type system adds an additional explicit test: type test. Why not taking this gift? :)

> The rule tests are about asserting the behaviour of the rule, making sure that the correct description and paths are built given the scenario under test.

Those are types fields/properties of a `Violation` (part of the interface)

> Verifying the violationType and rule title (e.g. via a deep equals) make these behaviour changes brittle and overly sensitive (e.g. would break when reclassifying test from SHOULD to MUST).

Agree, we shouldn't test any static fields of the rule (which don't change during the runtime). But it's OK to test if the rule's `violationType` was used to construct a `Violation` (we test the logic which takes the violation type from the right source).

> Additionally asserting with a deep equals can clearly indicate that things are different but does a poor job indicating what has changed.

Agree, IMO testing libraries/frameworks can be better in this.

> Admittedly it might be nicer to use a type safe method reference rather than name the field but I don't see that option in Assert4j...

why not simply:
```
assertThat(res?.description).isEqualTo("Information about protocol should be placed in schema. Current host value 'http://google.com' violates this rule")
assertThat(res?.paths).isEqualTo(listOf<String>())
```
instead of
```
assertThat(res)
                .hasFieldOrPropertyWithValue("description", "Information about protocol should be placed in schema. Current host value 'http://google.com' violates this rule")
                .hasFieldOrPropertyWithValue("paths", listOf<String>())
```

I'm not religious about this :) Let's hear some other opinions. @tkrop @maschleg @mfellner @netme @mkulak what do you think?

roxspring: > But it's OK to test if the rule's violationType was used to construct a Violation (we test the logic which takes the violation type from the right source).

For now :) ViolationType should be coming from the Check annotation very shortly and supplied via the RuleValidator rather than at construction, so conceptually should be treated much like the rule and title which technically are still supplied at construction too, for now.

> why not simply:

As stated elsewhere I find a single assertion much clearer to work with and therefore "simpler" than your suggestion. But like you, I'm really not religious about it, so if you all feel strongly then we can drop this PR and I'll move onto the next one :)

server/src/test/java/de/zalando/zally/apireview/RestApiIgnoreRulesTest.java
maxim-tschumak: > Additionally checking the size explicitly just adds room for the expectations to become out of sync

Good point

roxspring:
Dropping this in favour of #606

PR #602, opened 2017-11-30
#600 bump kotlin version

### LINKED ISSUE #600: Update to Kotlin 1.2
Since Kotlin 1.2 [is out](https://blog.jetbrains.com/kotlin/2017/11/kotlin-1-2-released/), it's probably a good time to update. 

## Preliminary steps:
- Check if all our dependencies / tools are compatible with Kotlin

mkulak:
Closes #600 

I don't really get why we excluded kotlin-reflect from jackson and declare explicit dependency on it.
But apparently it doesn't work with this setup so I had to change it.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -8,7 +8,7 @@ def versions = [
 
 buildscript {
 
-    ext.kotlin_version = '1.1.51'
+    ext.kotlin_version = '1.2.0'
 
     repositories {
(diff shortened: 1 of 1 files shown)


netme:
Awesome!

netme:
:+1:

mkulak:
👍

Issue #607, opened 2017-12-11
HyphenateHttpHeadersRule implements non-existent rule 131


roxspring:
[HyphenateHttpHeadersRule ](https://github.com/zalando-incubator/zally/blob/master/server/src/main/java/de/zalando/zally/rule/zalando/HyphenateHttpHeadersRule.kt) references a [guideline](https://zalando.github.io/restful-api-guidelines/#131) which doesn't exist.

Unless somebody is busy writing up the guideline then I imagine we should be deleting the rule.

maxim-tschumak:
Good catch! Let's move this rule to the Zally Rule Set ;)

roxspring:
Sounds reasonable. Will probably wait for #606 before making further changes though.

PR #618, opened 2017-12-15
Move HyphenateHttpHeadersRule from .zalando to .zally package

roxspring:
Moved as discussed in #607.

We could change the id 131 to ZallyRuleSet style (M131??) but that would break any existing "ignore" configurations. I've left it alone for now but am happy to be directed otherwise.


maxim-tschumak:
I think we can ignore "ignore configurations"
:+1:

maxim-tschumak:
:+1:

maschleg:
👍

PR #621, opened 2017-12-24
Test rules have unique names and a distinct RuleSet

roxspring:
While working on #559 I hit problems due to having both a production rule 166 and separate test rule 166. Updated the test rules and rule ignoring tests so that they use distinct ids and a separate RuleSet for good measure.

### DIFF
--- a/server/src/test/java/de/zalando/zally/apireview/RestApiTestConfiguration.java
+++ b/server/src/test/java/de/zalando/zally/apireview/RestApiTestConfiguration.java
@@ -1,20 +1,18 @@
 package de.zalando.zally.apireview;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import de.zalando.zally.rule.api.Severity;
+import de.zalando.zally.TestRuleSet;
 import de.zalando.zally.rule.AbstractRule;
 import de.zalando.zally.rule.ApiValidator;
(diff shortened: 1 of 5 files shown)


maxim-tschumak:
:+1:

server/src/test/java/de/zalando/zally/apireview/RestApiTestConfiguration.java
```diff
@@ -68,7 +68,7 @@ public String getTitle() {
 
         @Override
         public String getId() {
-            return "166";
+            return "T" + getClass().getSimpleName();
```
netme: Suggestion: why don't we just rename the rule to `TestCheckApiNameIsPresentJsonRule`, so that we can use a class name? `T` prefix looks a bit ugly and adds more confusion.

roxspring:
Can do. 

Thanks,

Rob

netme:
@roxspring btw, please also fix the commit messages. We have activated commit messages check in Zappr and now every commits requires a reference to the GH issue (e.g. `GH-559` or `#559`). Sorry for all these compliance topics. 

Please find an example in  #622

roxspring:
Made the requested changes on a clean branch and therefore opened a fresh PR #627

netme:
@roxspring btw, `git rebase -i` might help in this case, no need to close existing PRs :-)

PR #627, opened 2017-12-28
Test rules have unique names and a distinct RuleSet (Replacement PR)

roxspring:
While working on #559 I hit problems due to having both a production rule 166 and separate test rule 166. Updated the test rules and rule ignoring tests so that they use distinct ids and a separate RuleSet for good measure.


maxim-tschumak:
:+1:

netme:
👍

PR #628, opened 2017-12-31
559: Rule is now an annotation rather than basic interface

roxspring:
Related to #559 

`Rule` interface replaced with `@Rule` annotation.
Added `RulesManager` to manage `RuleDetails` and `CheckDetails` which link the rule/check instances with their metadata and allow reflective calls.


roxspring:
Codacy report seems a little cruel - can add javadocs to various classes if I absolutely must, but they weren't there before so it doesn't really seem right that they need to appear now!

netme:
@roxspring that's something new from Codacy. Let's ignore it since many old parts are affected.

netme:
Still need to take a closer look on the PR

roxspring:
Sure. There didn't seem to be sensible incremental steps left so it's a big change, but hopefully a positive one. I expected it to take some time to review.

Assuming all goes well then it should close off the work for #559 and (hopefully) be the end of me making huge sweeping changes to every rule!

netme:
👍

netme:
@roxspring sounds like it was a bug on Codacy side, the checks passed now.

netme:
@maxim-tschumak could you please take another look on that PR?

maxim-tschumak:
:+1:

Issue #631, opened 2018-01-08
OAuth2.0 Integration for Web UI


maxim-tschumak:
**Background information:**
Currently, we @ zalando have a "zally-web-ui appliance" project. It is mainly used to integrate OAuth2.0 by implementing OAuth flow on top of `@zalando/zally-web-ui` package. The result is deployed within the Zalando organization.
If I remember correctly, we initially decided to split up these projects to hide internal details about Zalando's infrastructure. But there is probably a way to make it configurable ;)

**User Story**
**As** a Zally maintainer
**I want** to include OAuth2.0 integration
**In order to:**
- enable OAuth2.0 for others
- simplify build and release process
- enable continuous delivery with CDP (Zalando's continuous delivery platform)
- simplify the project setup

**Acceptance criteria**
- `@zalando/zally-web-ui` can still be published as NPM package
- CDP builds `zally-web-ui-alpha` docker image on every change of `/web-ui`
- OAuth2.0 integration is disabled per default
- docker-compose script doesn't break

maxim-tschumak:
PR is on the way.

WDYT @netme, @ruiaraujo, @tkrop, @rbarilani ?

@roxspring We are aware that you may also have a production instance of Zally (Web UI). These changes should not break anything on your side. But please have a look at the PR once it is there.

rbarilani:
@maxim-tschumak Authentication can be already integrated, see https://github.com/zalando/zally/tree/master/web-ui#add-authentication

roxspring:
Thanks for the heads-up! FWIW we do have the web-ui running internally but actively disable the "production" profile since we don't feel the need to involve authz/n for Zally. I'll keep an eye out for any further pull requests related to this issue just in case.

rbarilani:
@maxim-tschumak As I already said in private: 

> I disagree to "support OAuth2" in a software that is meant to validate swagger specifications

I disagreed when the basic Aauth2 support was introduced in the server code and I still disagree.

In the web-ui I tried to be open also to other auth-mechanism (eg. you can even have an hardcoded list of username and passwords on your nodejs server or basic auth, or whatever your zally server will accept 😄 , take in mind that also the AJAX calls to zally server can be proxied, that's how we do in our zalando instance to add OAuth2 tokens stored in the encrypted session) . It requires a little more work for the user but can be more flexible than TRYING to support every OAuth2 provider out there (in my experience just a few are following "plain" RFC standards, but please correct me if I'm wrong).

As always just my 2 cents.

[...]

PR #632, opened 2018-01-08
Configurabel OAuth2.0 Integration

maxim-tschumak:
Related to #631.


web-ui/src/client/app/containers/app.jsx
```diff
@@ -70,7 +70,7 @@ export function App(props) {
         >
           Github Project
         </a>{' '}
-        - Copyright ZALANDO SE 2016
+        - Copyright ZALANDO SE {new Date().getUTCFullYear()}
```
netme: Nice catch

web-ui/package.json
```diff
@@ -9,7 +9,9 @@
   "scripts": {
     "start": "NODE_ENV=production node server.js",
     "dev":
-      "webpack-dev-server --config ./webpack/dev.js & NODE_ENV=development nodemon server.js",
+      "concurrently \"webpack-dev-server --config ./webpack/dev.js\" \"npm run prebuild && NODE_ENV=development nodemon server.js\"",
+    "prebuild":
+      "cd packages/passport-oauth2-zalando && npm install && npm run build",
```
netme: Why don't we fetch `passport-oauth2-zalando` directly from NPM repo?

netme:
+ general comment: could you please try to create smaller commits, so it will be easier to review and track back the changes later?

web-ui/src/oauth/auth/add-passport.js
```diff
@@ -0,0 +1,82 @@
...
+    if (Array.isArray(scopes)) {
+      return scopes;
+    }
+    if (typeof scopes === 'string') {
+      return scopes.split(',');
+    }
+    return [];
+  })(env.OAUTH_SCOPES || []);
+
+  const strategy = new ZalandoOAuthStrategy(
```
ruiaraujo: I dislike having these Zalando specific modules here in the open source repo.

rbarilani: @maxim-tschumak I dislike too, this was the reason why OAUTH2 Zalando is plugged in just in our internal project.

web-ui/package.json
maxim-tschumak: It's published in Zalando internal repo only.

web-ui/src/oauth/auth/add-passport.js
maxim-tschumak: I understand your concern. But,
1. We already have OAuth2.0 integration in Zally server.
2. This is the only way for us (team operator/administrator of Zally installation) to enable continuous delivery/deployment.
3. It's deactivated per default and should not harm any other installation.
4. Example for other OAuth2.0 integrations.

Was I able to convince you? :)

rbarilani: @maxim-tschumak I understand your concern but:

1. Well that's IMHO must be removed, since it's not a concern of "Rest Api Validator" to handle authentication

2. I think I can help on that.

3. Ok, fine

4. Can be solved by an open source application example

So if you really want to proceed, you are the maintainer, so.... do it 😄

But nope... you didn't convince me, sorry.

PR #635, opened 2018-01-09
Removed AbstractRule since it was adding no value

roxspring:
Removed AbstractRule since it was adding little value and no longer needed as everything of value is encoded in the Rule annotation.

Addresses #559.


maxim-tschumak:
Problem solved :)

maxim-tschumak:
:+1:

netme:
👍

Issue #640, opened 2018-02-05
Validate results against non-openapi URL responds with misleading schema validation violation [101]


roxspring:
Our Zally installation is often used by users submitting urls from GitHub or GitLab, and if you submit the link to the raw view then all works well: 
https://raw.githubusercontent.com/zalando/zally/master/server/src/main/resources/api/zally-api.yaml

However I've seen several users accidentally link to the html rendered view which unhelpfully announces schema violations (because it's trying to validate the html):
https://github.com/zalando/zally/blob/master/server/src/main/resources/api/zally-api.yaml

Additionally, in GitLab at least, some projects may be protected and so the links result in "401 Unauthorized" or "302 Found" (probably should be a 403 Forbidden), either resulting in the html being processed and a schema violation being displayed.

The error is misleading and so I think we ought to do something more informative.

I'm pretty certain that we should be displaying an error for non 200 response codes and not performing the validation. I think we should also be validating the Content-Type of the response, I'd like to think we should white-list the json and yaml mime-types but note that the github example above comes through as text/plain so wonder whether it's better to black-list text/html as the likely problem case. Thoughts welcome.

roxspring:
(Of course one partial answer here might be to provide Zally with an OAuth token it can use against our GitLab installation - I'll explore that separately)

tkrop:
Even authorized users may provide inadequate links. The communication between frontend and backend should provide enough insights to allow a sufficient error message about what is wrong.

netme:
That should be also implemented in #641 . Closing

PR #641, opened 2018-02-06
Validate HTTP response when retrieving a definition from a url

roxspring:
When retrieving a users API definition from a url we now:
* Verify that the response was in the HTTP 2xx success family.
* Verify that the response has a JSON or YAML looking Content-Type.

Failing either verification results in an error in the UI and avoids processing against validation rules.

Addresses #640.

### DIFF
--- a/server/src/main/java/de/zalando/zally/apireview/ApiDefinitionReader.java
+++ b/server/src/main/java/de/zalando/zally/apireview/ApiDefinitionReader.java
@@ -13,11 +13,8 @@ import org.springframework.web.client.ResourceAccessException;
 import org.springframework.web.client.RestTemplate;
 
 import java.util.List;
-import java.util.stream.Stream;
 
-import static java.lang.String.format;
 import static java.util.Arrays.asList;
-import static java.util.Collections.unmodifiableList;
 import static org.springframework.http.HttpStatus.BAD_REQUEST;
 import static org.springframework.http.MediaType.parseMediaTypes;
 
@@ -29,6 +26,7 @@ public class ApiDefinitionReader {
 
     // a whitelist of mime-types to accept when expecting JSON or YAML
     private static final List<MediaType> MEDIA_TYPE_WHITELIST = parseMediaTypes(asList(
+        // standard YAML mime-type plus variants
         "application/yaml",
         "application/x-yaml",
         "application/vnd.yaml",
@@ -36,12 +34,14 @@ public class ApiDefinitionReader {
         "text/x-yaml",
         "text/vnd.yaml",
 
+        // standard JSON mime-type plus variants
         "application/json",
         "application/javascript",
         "text/javascript",
         "text/x-javascript",
         "text/x-json",
 
+        // github.com raw content pages issue text/plain content type for YAML
         "text/plain"
     ));
 
--- a/server/src/main/java/de/zalando/zally/apireview/ApiDefinitionReader.java
+++ b/server/src/main/java/de/zalando/zally/apireview/ApiDefinitionReader.java
@@ -4,6 +4,9 @@ import de.zalando.zally.dto.ApiDefinitionRequest;
 import de.zalando.zally.exception.MissingApiDefinitionException;
 import de.zalando.zally.exception.UnaccessibleResourceUrlException;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
@@ -15,7 +18,7 @@ import org.springframework.web.client.RestTemplate;
 import java.util.List;
 
 import static java.util.Arrays.asList;
-import static org.springframework.http.HttpStatus.BAD_REQUEST;
+import static org.springframework.http.HttpStatus.UNSUPPORTED_MEDIA_TYPE;
 import static org.springframework.http.MediaType.parseMediaTypes;
 
 @Component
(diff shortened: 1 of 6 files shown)


server/src/main/java/de/zalando/zally/apireview/ApiDefinitionReader.java
```diff
@@ -5,17 +5,46 @@
...
+        "text/x-yaml",
+        "text/vnd.yaml",
+
+        "application/json",
+        "application/javascript",
+        "text/javascript",
+        "text/x-javascript",
+        "text/x-json",
+
+        "text/plain"
```
netme: Do we really need `text/plain`?

netme:
@roxspring looks good, could you please also add a test?

server/src/main/java/de/zalando/zally/apireview/ApiDefinitionReader.java
roxspring: Meant to comment about that. Ideally not needed but the one github example I tried out was Zally's own swagger.yaml which comes back as text/plain:

```
$ curl -s -i https://raw.githubusercontent.com/zalando/zally/master/server/src/main/resources/api/zally-api.yaml | grep Content-Type
X-Content-Type-Options: nosniff
Content-Type: text/plain; charset=utf-8
```

netme: Oh, I see, yes, maybe it's worth a comment.

netme:
👍

server/src/main/java/de/zalando/zally/apireview/ApiDefinitionReader.java
```diff
@@ -35,7 +64,19 @@ public String read(ApiDefinitionRequest request) throws MissingApiDefinitionExce
...
+            final ResponseEntity<String> response = client.getForEntity(removeSpecialCharactersSuffix(url), String.class);
+
+            final HttpStatus status = response.getStatusCode();
+            if (!status.is2xxSuccessful()) {
+                throw new UnaccessibleResourceUrlException("Unexpected response " + status.value() + " " + status.getReasonPhrase(), BAD_REQUEST);
+            }
+
+            final MediaType contentType = response.getHeaders().getContentType();
+            if (MEDIA_TYPE_WHITELIST.stream().noneMatch(contentType::isCompatibleWith)) {
+                throw new UnaccessibleResourceUrlException("Unexpected content type in response: " + contentType, BAD_REQUEST);
```
maxim-tschumak: Maybe `415 Unsupported Media Type` instead of the generic `400`?

netme: Valid point

roxspring: Is it? The request to Zally was in application/json and was parsed and understood correctly, responding with 415 Unsupported Media Type seems wrong in this circumstance. The request incorrectly references a 3rd party URL not supplying a definition and so I figure 400 Bad Request is right.

(As long as we're in the 4xx territory it all has the same effect so I don't really care and am happy to go in whatever agreed direction)

maxim-tschumak: I understand your point. Still, I would prefer `415` + documenting this behavior in the API specification.

You are absolutely right if we are talking about HTTP protocol only (since the content type of the original request is OK). But if we think about the REST interface, `403` provides more specific information to the API consumer.

The question is how important to us to distinguish between the wrong content type of the original request and the wrong content type of the "payload". I would say more specific status code (if documented correctly) has more value for the consumer. But this is a trade-off, of course.

server/src/main/java/de/zalando/zally/apireview/ApiDefinitionReader.java
```diff
@@ -35,7 +64,19 @@ public String read(ApiDefinitionRequest request) throws MissingApiDefinitionExce
 
     private String readFromUrl(String url) throws UnaccessibleResourceUrlException {
         try {
-            return client.getForEntity(removeSpecialCharactersSuffix(url), String.class).getBody();
+            final ResponseEntity<String> response = client.getForEntity(removeSpecialCharactersSuffix(url), String.class);
+
+            final HttpStatus status = response.getStatusCode();
+            if (!status.is2xxSuccessful()) {
+                throw new UnaccessibleResourceUrlException("Unexpected response " + status.value() + " " + status.getReasonPhrase(), BAD_REQUEST);
```
roxspring: Actually turns out the code coverage is right and we never actually get here - getForEntity() throws an exception for 4xx responses, presumably also for 5xx (and appears to follow redirects so never returns 3xx). Fixing.

server/src/main/resources/api/zally-api.yaml
```diff
@@ -79,8 +79,15 @@ paths:
...
-        includes the list of violations grouped by the API Guidelines rules.
+        (http://zalando.github.io/restful-api-guidelines/).
+
+        A successful response includes the list of violations grouped by
+        the API Guidelines rules.
+
+        If an api definition is supplied via url then any non-successful
+        responses from that will be passed on. For example you may be
+        using Zally without authentication but supply a password
+        protected url and still get a `401 Unauthorized` response.
```
roxspring: Documented existing behaviour of passing on downstream error responses.

server/src/main/resources/api/zally-api.yaml
```diff
@@ -97,6 +104,10 @@ paths:
           description: Input file not parsable
           schema:
             $ref: 'https://zalando.github.io/problem/schema.yaml#/Problem'
+        415:
+          description: API definition url responded with content-type which clearly isn't JSON or YAML.
+          schema:
+            $ref: 'https://zalando.github.io/problem/schema.yaml#/Problem'
```
roxspring: Documented new behaviour of rejecting downstream error responses with the wrong content-type.

netme:
Now Travis seems to be unhappy :-( Maybe integration test?

netme:
👍

maxim-tschumak:
@roxspring one last thing: please increase the (minor) api version in the api spec. Thanks!

maxim-tschumak:
:+1:

netme:
👍

maxim-tschumak:
@netme never sleeps :)

Issue #645, opened 2018-02-18
Remove UseOpenApiRule.getGeneralViolation()


roxspring:
`UseOpenApiRule.getGeneralViolation()` is used by `RulesValidator` to signal parse failures that really ought to be picked up by `UseOpenApiRule` used normally. 

As a side effect it means that `UseOpenApiRule` needs to get passed around in the tests and introduces unnecessary dependencies from core packages into the zalando rules package.

`UseOpenApiRule` should be a standard rule treated as all the others.

PR #646, opened 2018-02-18
UseOpenApiRule is not treated specially

roxspring:
As long as `UseOpenApiRule` is in the active list of rules then `UseOpenApiRule.getGeneralViolation()` can be removed.

Addresses #645.


maxim-tschumak:
:+1:

netme:
Nice change!

netme:
👍

Issue #648, opened 2018-02-18
CLI output JSON formatted


roxspring:
For internal CI purposes we're currently running a variation of the CLI tool which outputs in JSON format so that the script running it can do further processing using jq etc. 

Opening this issue to allow a PR to be opened to make the change available upstream.

netme:
Hey @roxspring. I'm really happy that your team is using CLI, however I don't fully agree with this approach. If you are interested in JSON output, why not to trigger API directly and post-process it with jq?

netme:
Maybe you can provide some example and we will see how to make CLI more useful

maxim-tschumak:
I agree with @netme.
@roxspring, please describe your use case

roxspring:
We're using GitLab CI and have a job configured with each API project to check it's swagger.yaml file against Zally. Until recently that would just use the classic cli to fail the build on MUST violations, but consequently SHOULD and lesser violations were going unnoticed. So we decided that the job should be updated so that it additionally uses the GitLab API to add markdown formatted comments to the commit in question. The job consists of a small bash script running within a docker image.

We were already using the freshly built CLI, which itself already knew how to make the requests. I knew that the API required the yaml to be wrapped in an ApiDefinitionRequest and wasn't sure how to easily/reliably do that from a curl command line. So hacking the Zally CLI seemed the simplest way forward.

On reflection, I can understand that it would be ideal for such integrations to use the API directly but I was "put off" by the complexity of forming a suitable request. Perhaps this suggests we should have a 2nd simpler end point that allows direct POSTing of the json/yaml in question to ensure that barrier is kept low.

On the other hand, I'm a little surprised at making the CLI easier to process by a script - automation seems to be the primary use case of such command line tools these days.
We're using GitLab CI and have a job configured with each API project to check it's swagger.yaml file against Zally. Until recently that would just use the classic cli to fail the build on MUST violations, but consequently SHOULD and lesser violations were going unnoticed. So we decided that the job should be updated so that it additionally uses the GitLab API to add markdown formatted comments to the commit in question. The job consists of a small bash script running within a docker image.

We were already using the freshly built CLI, which itself already knew how to make the requests. I knew that the API required the yaml to be wrapped in an ApiDefinitionRequest and wasn't sure how to easily/reliably do that from a curl command line. So hacking the Zally CLI seemed the simplest way forward.

[...]

PR #649, opened 2018-02-23
cli now supports --json option switching all output to JSON

roxspring:
As discussed in #648 - was useful to us internally and figured it might be good for others too.


netme:
As mentioned in #648, let's discuss how to make it better. This solution can be too general and lead to the tool misuse.

Issue #446, opened 2017-07-25
Implement support for disabling rules as local as possible.


maxim-tschumak:
**As** a user of Zally
**I want** to be able to disable specific rules on different levels of the Swagger hierarchy.
**So that** the integration of Zally into the build process is more flexible

Related to / Triggered by: #406

Example by @tfrauenstein:
> For instance, rule "Should: Limit number of Sub-Resource Levels" only applies to resource paths and disabling should be for a specific path -- technically an OpenAPI "path item object" instance -- and not for the complete API. Hence disabling should only become effective, if it is done within the smallest context of the OpenAPI element relevant for the rule. Is this requirement agreed?

innokenty:
I don't see any other way of implementing this, rather than hardcoding checks of the `x-zally-ignore` extension checks in the code of every single rule on each level of swagger doc hierarchy. That's a quite time-consuming approach, also taking into consideration all the possible bugs and combinations of different inner tags in there.

This is doable though, but I would first wait and make sure this is actually a highly demanded feature. I doubt there are many people that really want to disable some checks on some level and not globally.

maxim-tschumak:
I see two options, besides the one proposed by @innokenty :

1. A dynamic **RuleRunConfiguration** would describe against which _parts_ of the API (meta information, concrete endpoints, concrete models, concrete parameters) the rule should check. Each rule has to implement an interface to check specific parts of the API (e.g. `Rule#checkEndpoint(e string): List<Violations>`).

2. **Exclude** paths marked with `x-zally-ignore` **after the check**.

@netme @maschleg @tkrop @mfellner What do you think?

tfrauenstein:
yes, implementation is more involved. 
However, I see it as a must feature already in scope of the original feature idea:

"When drafting an API, the API designer usually runs Zally multiple times.
He then checks Zallys violation output and decides how to handle each violation.
Sometimes API is changed as to avoid the violation, sometimes API is not changed on purpose -- often the case for SHOULD and MAY violations.
Improve API designer convenience via support of pragma directives inside API definitions that disable checking of specific rules for specific OpenAPI elements.
By this, API designers can avoid efforts for redundant checking of violations for multiple Zally runs during API evolution. It is also a must-feature for integration of Zally in automated CD or operation monitoring tooling."

I'm sceptical that 1st delivery step of only global disabling really make sense: 
[...]

Issue #204, opened 2017-03-22
Be compatible with OpenAPI 3.0


netme:
It's more an investigation task. Since OpenAPI 3.0 is on the way, we should be able to validate it as well.

## Possible steps
1. Check with API guild if we are ready with OpenAPI 3.0
2. Check if the existing swagger parsing libraries are compatible
3. Check the existing rules

Visual guide: https://blog.readme.io/an-example-filled-guide-to-swagger-3-2/

## Related Issues
- [ ] #714 Refactor Checks to Operate on the Context Object

fmueller:
This depends on `swagger-parser`library I guess. So, we can contribute to this as well.

netme:
Yes, also can be an option

netme:
And now it's officially announced https://swagger.io/announcing-openapi-3-0/

 🎉

innokenty:
I'm afraid we gonna have a dependency clash, in case we wanna import `swagger-models` maven artifact with the new api v3. It's not released yet, but it looks like the packages and class names are mostly the same: https://github.com/swagger-api/swagger-core/tree/feature/3.0.0-rc0/modules/swagger-models

Any ideas, how we wanna support v2 and v3 simultaneously in one process? Any **_r e p a c k a g i n g_** fans in the thread? 😃

innokenty:
Wouldn't it be easier and nicer to split server into 3 modules: core, v2-validator and v3-validator? Running two instances of the server and having a simple switch in the UI, which validator to use.

tkrop:
Shouldn't validation be based on the `swagger` header in the API specification? I would expect the parser to create a similar (or even same logical) model for both versions. Can we check whether we can update to the latest parser version that is able of parsing v2 and v3 and test some APIs?

I would like to announce that our tools are OpenAPI 3.0 compatible as soon as possible, and Zally seems to be the only unchecked application.

netme:
@tkrop yes, we should do something like that. I will put this ticket as a priority then to check the status of the swagger parser we are using and keep you updated.

fatroom:
@netme any news on this topic?

netme:
@fatroom , sorry, we have abandoned this topic a bit. I will take a look on the latest `swagger-parser` implementation

netme:
Looks like there is an official support of 3.0 which will be coming in `2.0` release. I will try to update my local branch to `2.0.0-rc1` to check if it's good enough, however we need to wait until the official release will be published.

[...]

PR #668, opened 2018-03-11
Improved support for x-zally-ignore

roxspring:
* Dropped recently added `SwaggerIgnoresExtension` (adding a second parameter to check methods was pretty ugly).
* `RuleValidator` uses a new `Context` to contain everything a check might need to know:
  * `SwaggerRuleValidator` uses new `SwaggerContext`.
  * `JsonRuleValidator` uses new `JsonContext`.
* Check methods can optionally take a `Context` parameter instead of model root.
* Context implements `x-zally-ignore` logic applicable to vendor extension locations.
* SwaggerContext offers `validatePaths(..)` utility to implement `x-zally-ignore` aware path based validation.
* See `LimitNumberOfSubresourcesRule` for demonstrates use of new utility method.
* More utility methods could be added to iterate over operations, parameters, responses etc but I suggest they evolve as needed and allow #446 to be closed.
* Added `AvoidXZallyIgnoreRule` to explicitly highlight and discourage `x-zally-ignore` usage.


roxspring:
@maxim-tschumak, @netme: Would appreciate some review on this. Personally I think the new utility methods should make most rules much simpler to implement and also buys `x-zally-ignores` support into the bargain. Additionally I think `AvoidXZallyIgnoresRule` is the perfect way to inform users of `x-zally-ignores` usage.

maxim-tschumak:
Thanks for this proposal @roxspring! I find it elegant, but I do have some concerns:

1. Users don't know which of the rules support the "ignore on path level" functionality. 
1. If we go this direction, we should avoid having multiple APIs for the Check (in this case we should stick to `Context` as the only possible parameter of the Check function). This also means we have to adjust all Checks asap (which would fix 1.).
1. Maybe we can simplify the Check interface. (important since our customers have to implement this interface)
1. (minor): `toMessages` method should be renamed, `check(Path)`, `validate`, etc. would be better)

`AvoidXZallyIgnoresRule` is a great idea!

@mfellner @tkrop @netme Please share your thoughts!

roxspring:
Glad you like it @maxim-tschumak - hope others do too!

I had intended (1) to be addressed by (2) in time and should have stated that somewhere - certainly wanted to avoid any more merge requests where I'm forced to touch every single rule! `validate(..)` sounds better for (4) - think I tried it and found it confusing but will have another go at it today and maybe fall back to `check(..)` if that seems clearer.

Are you able to expand on (3) though? I'm not clear what you're wanting to simplify.

maxim-tschumak:
> Are you able to expand on (3) though? I'm not clear what you're wanting to simplify.

I don't have a concrete proposal so far. The implementer of a check has to:
1. know the signature of the `validate` method (`SwaggerContext -> Violation?`), and
1. know the signature of the `validatePaths` method (`String, (Swagger, String, Path) -> List<String> -> Violation`)

This is a bit too much to implement "just" a check on a swagger object. I hope we can simplify this for the check implementers.

> [..] move the description into the `@Check` annotation [..]

I was also thinking about this. For now, we use Check descriptions as Violation texts. We can separate these things:
* Static Check description in the annotation (answering the question "What's wrong?")
* Violation text is used to motivate a change (answering the question "How to fix?")

This should not be a part of this particular change though.

roxspring:
That makes sense. We've been using some very similar extension methods for writing rules in-house for a couple of months now and I guess we've been valuing the clarity of the result over the ease of getting started without copying another.

An alternative approach I mentioned before (Oct/Nov timeframe?) is to move the model traversal into `SwaggerRuleValidator` and then use reflection to invoke checks matching a handful of signatures along the lines below.

- validateSwagger: (Swagger) -> Violation?
- validatePath: (Swagger,String,Path) -> Violation?
- validateOperation: (Swagger,String,Path,HttpMethod,Operation) -> Violation?
- validateParameter: (Swagger,String,Path,HttpMethod,Operation,Parameter) -> Violation?
- validateResponse: (Swagger,String,Path,HttpMethod,Operation,String,Response) -> Violation?

Implementers would still have a few signatures to remember but only one set, and maybe those could be provided in interfaces to help people get started. Perhaps I should prepare an alternate PR based on this approach and see which we prefer?

maxim-tschumak:
I remember your proposal @roxspring :) It's probably the best way to go. Please let's wait for others' opinions.

mfellner:
I agree with @maxim-tschumak that the API for rule authors should be as simple as possible. However the approach of creating multiple adaptor methods inside `Context` (`validatePath`, `validateOperation`, etc.) makes the API more complicated imho.

I like the idea of having a `Context` to abstract away some common logic for each rule and to (eventually) only pass this context to each rule. In fact, we wanted to do the same thing to implement https://github.com/zalando/zally/issues/204 and support both Swagger 2 and OpenAPI 3.

But I also think that giving rules the raw `Swagger` object is cleaner and more maintainable than providing special accessor methods inside `Context`. This and the introduction of a callback-based API also makes implementing rules more complicated than before, imo.

In https://github.com/zalando/zally/issues/689 we suggest traversing `Swagger` ahead of time to create a lookup map to get a JSON pointer for any child node of `Swagger`. This traversal can also be used to check if a child node has the method `vendorExtension`, check wether `x-zally-ignore` is set, and **completely remove** the respective sub-tree of the `Swagger` object.

That way ignored parts of `Swagger` would already be removed/nulled ahead of time, before running the checks.

@roxspring, what do you think about this alternate approach?

roxspring:
> I agree with @maxim-tschumak that the API for rule authors should be as simple as possible. However the approach of creating multiple adaptor methods inside Context (validatePath, validateOperation, etc.) makes the API more complicated imho.

Totally agree that the new rule author should need to understand minimal API contracts. I think we're also all in agreement that x-zally-ignore processing should be happening centrally; outside of the rules if at all possible. 

I think the bit where we perhaps disagree is that my team have found that repeating big chunks of traversal logic from rule to rule is boring, error-prone and makes reading the code harder. To resolve this I'd like to abstract away the traversal logic of "for each response of each operation of each path" into a single readable expression, and the (possibly more advanced) rule author to focus directly on their rule logic.

My suggestion to address this is to allow a set of API check signatures, starting with `validate(Swagger):Violation?`, but potentially expanding to:

- `validate(Swagger):Violation?`
- `validate(Swagger,String,Path):Violation?`, -
- `validate(Swagger,String,Path,HttpMethod,Operation):Violation?`, 0 -
- `validate(Swagger,String,Path,HttpMethod,Operation,String,Response):Violation?` etc. 

Admittedly the final parameter will often be all that's of interest, but sometimes one/more of the other parameters is needed for context, and when implementing any unused parameter can simply be named `_` for clarity. A side benefit of the abstracted traversal is that the paths can be prefixed in a uniform manner, such that all path related violations have paths beginning "paths: /the/path: " or whatever is agreed upon.

I guess we could try and query the model using JSON pointer expressions to hide some of the traversal but I'm not clear how to include the benefits of contextual model elements or uniform paths, and it has the cost of developers having to learn JSON pointer (which I failed at with a 5min attempt earlier today).

> I like the idea of having a Context to abstract away some common logic for each rule and to (eventually) only pass this context to each rule. In fact, we wanted to do the same thing to implement #204 and support both Swagger 2 and OpenAPI 3.
> 
> But I also think that giving rules the raw Swagger object is cleaner and more maintainable than providing special accessor methods inside Context. This and the introduction of a callback-based API also makes implementing rules more complicated than before, imo.

Agree that the callback approach over-complicates things, and direct access to the Swagger root object is much easier for rule developers to use. 

> In #689 we suggest traversing Swagger ahead of time to create a lookup map to get a JSON pointer for any child node of Swagger. This traversal can also be used to check if a child node has the method vendorExtension, check whether x-zally-ignore is set, and completely remove the respective sub-tree of the Swagger object.
> 
> That way ignored parts of Swagger would already be removed/nulled ahead of time, before running the checks.

Now that's an idea that hadn't occurred to me! You'd have to take a deep copy of the entire model before pruning (or re-parse), so that the original model can be used for the next rule. I'm not too clear how a JSON pointer make this easier but I'm sure you have a plan. My initial reaction is that perhaps additional copying and pruning might burn resources unnecessarily - but it's definitely worth exploring!!

mfellner:
@roxspring thank you for your feedback on this matter, it is really helpful! Let me try to list the different concers that we've started to discuss across multiple issues:

1. cleaner implementation of `x-zally-ignore`
2. the notion of `Context` as a wrapper/adapter/decorator for `Swagger`
3. abstracting away redundant traversal of `Swagger` in each rule
4. OpenAPI 3 support with fallback support for Swagger 2
5. generating violation paths automatically instead of manually in each rule
6. using only valid RFC-standard JSON pointers as violation paths
7. overloading/abusing Violations attributes to store additional information

**(1.)** So far we've said that it can either be realised by:
* having abstract adapter methods to access nodes inside `Swagger` that filter out ignored nodes
* traversing and pruning the `Swagger` object ahead of time

**(2.)** We agree it's useful for both (3)., (4.) and maybe (5.).

**(3.)** A helpful change than can help making rule implementations much simpler and also support (4.)! However it might break backwards compatibility or we need to update all rules?

**(4.)** We definitely want OpenAPI 3 support asap, but still want to support Swagger 2 as well. (2.) and (3.) can help with this feature.

**(5.)** Another helpful change to make rule implementations easier. It can be either achieved by
* an implementation for (3.) where each adapter method also constructs the correct path
* an ahead-of-time traversal as explained in https://github.com/zalando/zally/issues/689

**(6.)** This is currently not enforced in any way but (5.) could implement it that way. Valid JSON pointers would be more useful to clients, e.g. the web client.

**(7.)** We need to rethink the `Violation` interface or what a rule should return.

About your last comment:

> Now that's an idea that hadn't occurred to me! You'd have to take a deep copy of the entire model before pruning (or re-parse), so that the original model can be used for the next rule. I'm not too clear how a JSON pointer make this easier but I'm sure you have a plan. My initial reaction is that perhaps additional copying and pruning might burn resources unnecessarily - but it's definitely worth exploring!!

* My expectation would be that a `x-zally-ignore` should disable *all* rules and checks so we would need to prune `Swagger` only once and not keep an unmodified copy around.
* JSON pointers won't be useful for the zally-ignore feature, I hope the listing above better explains the different concerns.

mfellner:
A potential issue with the suggested solution for (3.) of having multiple different `validate` method signatures is that some rules still need to get different properties from different places inside `Swagger`. For example, `SecureWithOAuth2Rule`.

In those cases the method signature would have to be `validate(Swagger):Violation?` and the rule would still need to implement all the necessary traversal. @vadeg and I thought that a better solution to solve (3.) could be either a plain old adepter or a fancy [Kotlin DSL](https://kotlinlang.org/docs/reference/type-safe-builders.html).

roxspring:
Great summary of where the discussion is at @mfellner 👍 - I agree with all the points you numbered. Let me add (8.) and respond to your comments regarding (3.):

**(8.)**
> My expectation would be that a x-zally-ignore should disable all rules and checks so we would need to prune Swagger only once and not keep an unmodified copy around.

My expectation is different. I was expecting to be able to ignore specific rules at lower locations. I want my team to be obeying as many of the rules as possible, and am ok with them temporarily using `x-zally-ignore: [SomeNewRule]` on a subtree that can't meet the new rule yet for legacy reasons, but I don't want them to then accidentally fall out of compliance with all other rules.

**(3.)**
> A potential issue with the suggested solution for (3.) of having multiple different validate method signatures is that some rules still need to get different properties from different places inside Swagger. For example, SecureWithOAuth2Rule.

That's precisely the point of the various signatures, you provide the node in the model that needs validating preceeded by it's ancestry which might be needed for context. e.g. `checkUsedScopesAreDefined(..)` would operate at the `Operation` level but would need the root `Swagger` context to lookup defined scopes.

```
    fun checkSecurityDefinitions(_: Swagger, securityDefinition: SecuritySchemeDefinition): Violation? {
        // validate specific definition
    }

    @Check(severity = Severity.SHOULD)
    fun checkPasswordFlow(_: Swagger, securityDefinition: SecuritySchemeDefinition): Violation? {
        // validate specific definition
    }

    fun checkUsedScopesAreDefined(swagger: Swagger, _: String, _: Path, _: HttpMethod, op: Operation): Violation? {
        val definedScopes = getDefinedScopes(swagger)
        // validate specific operation
    }
```

**(3.)** 
> @vadeg and I thought that a better solution to solve (3.) could be either a plain old adepter or a fancy Kotlin DSL.
I haven't played with Kotlin DSLs so can't really comment, other than to say I'm curious how it would work out!

vadeg:
I have just played a little bit with Kotlin and I have an idea to implement simple and flexible API for `Swagger` and `OpenApi`. I think by simplifying traversal logic we would not need several `validate` methods for all rules and can leave it up to specific rule.
For example:
```
        val parseResult = OpenAPIParser().readLocation(url, null, ParseOptions())
        val root = parseResult.openAPI

        // Get all operations having tag "super-pets" from all paths equals "/pets"
        val allOperations = root.paths { (name, _) ->
            name == "/pets"
        }.operations { it ->
            it.tags.contains("super-pets")
        }

        println("All /pets operations: ${allOperations.all().size}")

        // Get all responses from all operations above
        val allPetsResponses = allOperations.responses()
                .flatMap { it -> it.entries }
        println("All responses for all /pets operations: ${allPetsResponses.size}")
```
I have just slightly updated `OpenApi (former Swagger class)` with additional methods (thank you, Kotlin) and now I can traverse more flexible rather than before. 
Also, I can join methods into chains, implement filtering, grouping and etc. 
I think if we can simplify traversal logic by implementing extensions/delegates and etc. then we can keep one `validate` method and traverse much easier.

This 
```
    fun checkSecurityDefinitions(_: Swagger, securityDefinition: SecuritySchemeDefinition): Violation? {
        // validate specific definition
    }
```
can be rewritten like this
```
   val root = //Swagger object init
   val violations = root.securitySchemes { name, scheme ->
        //validate specific definition
        Violation("Something wrong with security")
   }
    // Do something with violations...
```
**Pros**
+ All traversable logic is located "close" to the source.
+ More traversable callbacks can be added.
+ Everything can be checked at compile time.
+ No "magic" with multiple `validate` methods.

**Cons**
- Need carefully implement callbacks.
- Rule remains simple.
- Less `map -> flatMap(map -> map -> flatMap)` inside rules.

mfellner:
Thank you for your answers, @roxspring!

> My expectation is different. I was expecting to be able to ignore specific rules at lower locations. 

Yes, I can see that now. It makes sense to have a more granular ignore-logic. In that case it would be inefficient to traverse and prune `Swagger` for every rule. However we then need to be able to detect wether a certain rule operates on a sub-tree of `Swagger` with a `x-zally-ignore` attribute matching that specific rule at its root.

I could imagine doing that by creating another data structure during an initial traversal pass that maps `Swagger` node objects to (already parsed) `x-zally-ignore` values. Then we can use this map to check wether a node should be ignored. The map would probably need to have entries not only for the root node of sub-tree (the one with `x-zally-ignore`) but also all of its children...

> That's precisely the point of the various signatures, you provide the node in the model that needs validating preceeded by it's ancestry which might be needed for context. e.g. checkUsedScopesAreDefined(..) would operate at the Operation level but would need the root Swagger context to lookup defined scopes.

```kotlin
fun checkUsedScopesAreDefined(swagger: Swagger, _: String, _: Path, _: HttpMethod, op: Operation): Violation? {
    val definedScopes = getDefinedScopes(swagger)
    // validate specific operation
}
```

I am not convinced that we would gain much by this though. Although we can directly access `op` we still need to have some method `getDefinedScopes` to get the defined scopes from the root. And the other parameters need to be ignored with `_`.

This would also not be very intuitive and discoverable to write as opposed to the type-system assisted way of accessing properties on the `Swagger` object directly. However the problem of verbose and cumbersome traversal with `flatMap` etc. remains. Another question that I have: would the RulesValidator call a method like `checkUsedScopesAreDefined(swagger: Swagger, _: String, _: Path, _: HttpMethod, op: Operation)` multiple times, essentially swa * str * pat * meth * op times (those variables being the number of possible attribute combinations)?

That's why I think @vadeg's approach of simplifying the access and traversal of `Swagger` properties is more promising. Given that with the approach of special `validate` methods we would still need "helper" methods for accessing some properties, why not reduce the whole solution to just making this access easier and more concise?

roxspring:
I know this PR has ended up mostly in discussion, but have you looked at the changes proposed? I find them to be very similar to those proposed by @vadeg, with just a couple of differences:

- The methods operate on a Context object rather the model root, thereby allowing the additional metadata essential for implementing `x-zally-ignore` functionality (i.e. the id of the current rule)
- The methods also handle the tedious construction of Violations.
- Possibly @vadeg might be advocating dropping support for multiple checks within a rule which would be a shame IMHO.

In-house we've been using similar extension methods (on Swagger, no ignore support, constructing Violations) for a couple of months and find rules much cleaner to read/write/maintain with them. 

I note also that @mfellner's first comment on this thread objected to complicating things with a "callback-based API" and yet this is exactly what's being approved of from @vadeg. I don't call this out to complain but to highlight that I think we've now all traveled the same mental journey around the problem-space and seem to be approaching full circle! Hopefully this means we are closer to identifying the best (aka least-worst) route forward! I think the options on the table, applied to a more interesting deep iteration (worst case?) are as follows:

I think @vadeg proposes **(1.)**:
```
@Check(severity=Severity.MUST)
fun validate(root: Swagger) = 
    root.responses { pattern, path, method, operation, status, response ->
        when {
            status != "200" -> null // some filtering based on context
            response.schema == null -> Violation("Schemas should be specified", /* path construction */)
            else -> null
        }
   }
```
vs possibly the similar **(1b.)?**:
```
@Check(severity=Severity.MUST)
fun validate(root: Swagger) = 
    root.paths { pattern, path ->
        path.operations { method, operation ->
            operation.responses { status, response ->
                when {
                    status != "200" -> null // some filtering based on context
                    response.schema == null -> Violation("Schemas should be specified", /* path construction */)
                    else -> null
                }
            }
        }
   }
```
vs as in my PR **(2.)**
```
@Check(severity=Severity.MUST)
fun validate(context: SwaggerContext) =
    context.validateResponses("Schemas should be specified") { pattern, path, method, operation, status, response ->
        when {
            status != "200" -> null // some filtering based on context
            response.schema == null -> /* path construction */
            else -> null
        }
   }
```
vs as previously proposed **(3.)**
```
@Check(severity=Severity.MUST)
fun validate(root: Swagger, pattern: String, path: Path, method: HttpMethod, operation: Operation, status: String, response: Response) =
    when {
        status != "200" -> null // some filtering based on context
        response.schema == null -> Violation("Schemas should be specified", /* path construction */)
        else -> null
    }
```

Additionally @maxim-tschumak and I also pondered variations where the violation description is static and pulled out to the annotation too, e.g.: **(4.)**
```
@Check(severity=Severity.MUST, description="Schemas should be specified")
fun validate(root: Swagger, pattern: String, path: Path, method: HttpMethod, operation: Operation, status: String, response: Response) =
    when {
        status != "200" -> null // some filtering based on context
        response.schema == null -> Violation(/* path construction */)
        else -> null
    }
```

Personally I like (1.) but find it viable only if we pre-process the model for `x-zally-ignore`, and so probably lean towards (2.) as the most readable (especially if we move the description to the annotation as in (4.)). 

I do also recognize though that (3.) might be the easiest to understand for new rule authors not starting from a cut & paste, but I question how many won't start from a cut & paste and whether we really ought to optimize for them!?

roxspring:
Sorry - failed to address a couple of questions:

> Another question that I have: would the RulesValidator call a method like `checkUsedScopesAreDefined(swagger: Swagger, _: String, _: Path, _: HttpMethod, op: Operation)` multiple times, essentially swa * str * pat * meth * op times (those variables being the number of possible attribute combinations)?

Yes, that's would be my intention, probably doing some aggregations of Violations where rule/severity/description match.

> That's why I think @vadeg's approach of simplifying the access and traversal of Swagger properties is more promising. Given that with the approach of special validate methods we would still need "helper" methods for accessing some properties, why not reduce the whole solution to just making this access easier and more concise?

Totally agree that "easier and more concise" is achieved but it doesn't address the underlying problem of local `x-zally-ignore` processing.

vadeg:
Thank you, @roxspring for detailed response. 

Actually, I am not advocating. I am totally fine with your solution. It works for `paths` and it is fine. The problem is that now every rule author should remember that `path` should be retrieved only using `Context.validatePaths` methods. But there are still `Swagger` object accessible and no one prevents from accessing `swagger.paths` and get unfiltered paths. I think that is error-prone because:
* authors should know about this behaviour and prevent using Swagger object for paths directly.
* reviewers should know about this behaviour and prevent authors to use Swagger object for paths directly.

Actually, we were thinking how to apply `x-zally-ignore` not only to path but in general. For example there is a rule `SnakeCaseInPropNameRule` and it works with
* definitions
* paths
* schemas
* parameters in definitions
* parameters in path
* responses

To all of them `x-zally-ignore` can be applied as well. That means I might have several more methods on context like `validateDefinitions`, `validateSchemas`, `validateSomethingElse`. All of these method has only one goal - filter out all entities marked with `x-zally-ignore` for specific rule. 
We are thinking are there some other options how it can be done. 

> I think the bit where we perhaps disagree is that my team have found that repeating big chunks of traversal logic from rule to rule is boring, error-prone and makes reading the code harder. To resolve this I'd like to abstract away the traversal logic of "for each response of each operation of each path" into a single readable expression, and the (possibly more advanced) rule author to focus directly on their rule logic.

How many of these methods we would need to define? Of course a lot of rules working with paths but they all work in a different way. For example `MediaTypesRule` currently works with `produces`/`consumes` but in API 3.0 it will work with `path/responses` or `components/responses`,

Rule `SnakeCaseInPropNameRule` and `CommonFieldTypesRule` works with all objects.
I have an impression that in the end we will end up with one method per model entity. 

Our ideas, your ideas - all of them have pros and cons. Currently we are trying different ways how we can solve these problems (or can it be solved in general). I think together we can implement proper solution.

roxspring:
> The problem is that now every rule author should remember that path should be retrieved only using Context.validatePaths methods. But there are still Swagger object accessible and no one prevents from accessing swagger.paths and get unfiltered paths

Yes, this is a fair point, and one which I'd largely been ignoring.

> How many of these methods we would need to define? ... I have an impression that in the end we will end up with one method per model entity.

Yes, this is a fair accusation too.

OK, so I suspect the only real way to truly enforce that rule XYZ properly ignores elements with `x-zally-ignore: [XYZ]` is to prune the object model before handing it to the rule to process. Each pruned variation of the tree nodes would have different hashCodes throughout and so any Map<Object,JSONPointer> would likely need to be rebuilt for that pruned variation too. Our assumption is that this would be wasteful of time/memory.

Clearly the pathological case here is that somebody submits a swagger definition with every enabled rule being ignored for different subtrees of the model - and perhaps this burns more time/memory than we'd like - but if it still completes then maybe that's not too bad? I would hope that the normal case would be to be ignoring a handful of rules at worst, likely clustered around similar subtrees - in which case there should be lots of opportunities to reuse trees and caches.

vadeg:
@roxspring @mfellner , some more ideas came to my mind except those that @roxspring already suggested:

## Option 1: Decorators
Wrap every object class with some kind of decorator which can filter object based on extensions. 

**Pros**
+ Clear logic

**Cons**
- Original hierarchy will be duplicated with decorators hierarchy.
- Very time consuming and error-prone

## Option 2: AOP
Use AOP (f.e. AspectJ) to pre-process Swagger data model and attach pointcuts to specific places (like Spring is doing with beans).

**Pros**
+ Whole model and be processed with few lines of code.

**Cons**
- Unexpected behaviour. It would be hard to understand why it works like that only looking at the code.

## Option 3: Filter violations

Build a map of ignored paths for each rule. Pass whole Swagger model to each rule (like it works right now) because only rule knows what it needs for checks. Receive violations from rules like JSON pointers and exclude violations which should be ignored using ignored paths.

**Pros**
+ Easy to implement because we will have JSON pointers map for another reasons.

**Cons**
- Rules will be executed even if it is not needed.

roxspring:
In `SecureWithOAuth2Rule.checkUsedScopes()` we validate Path objects conform with what's found in `SecuritySchemeDefinition`s. One thing I like about filtering the input model the rule sees (whether pre-pruning, decorating, aspects) is that you could use `x-zally-ignore` in the `SecuritySchemeDefintion`s and effect the results. As far as I can think through, filtering output Violations is unable to achieve the same.

Not sure that this theoretical benefit translates into any practical benefit though, so perhaps filtering the results is the sane way forward. It's certainly seems a lot more practical with the JSON pointers than it is currently with unstructured strings.

vadeg:
@roxspring @mfellner I have created #692 as one of the possible solutions how to ignore model instances based on vendor extensions.

mfellner:
Thank you everybody for the detailed inputs! It can feel cumbersome to discuss these topics for so long however it is hopefully going to pay off if we get it right.

Personally, I would prefer solutions that do not need more 3rd party libraries or that introduce yet another programming concept to Zally (e.g. aspects). I'm even skeptical using things from Spring because imo it should only be used on the web service layer of the application.

Ideally we find a solution that only uses **vanilla Java and Kotlin features**.

In the meantime I was able to implement the Swagger "model tree walker" (https://github.com/zalando/zally/blob/openapi3/server/src/test/java/de/zalando/zally/util/OpenApiWalkerTest.kt) that makes it easy to get **JSON pointers** and `x-zally-ignore` information for any node in the Swagger (or OpenAPI) object.

To me, this in combination with the "telescoping" validate methods approach that @roxspring has suggested and/or a light-weight and *native* variant of @vadeg's proxies looks the most promising to me. @tkrop also chimed in and said that we wouldn't be opposed to the "telescoping" validate methods either.

But to move forward I believe we need a **small incremental improvement**. The `OpenApiWalker` would allow for an intermediate solution, evaluating `x-zally-ignore` after the fact:

* change the `Violation` interface to only have two properties
  * `description: String`
  * `node: Object` (the violating node in Swagger/OpenApi)
* `RulesValidator` reads the violation and handles it
  * looks up the JSON pointers (paths) in the node map based on the node in the `Violation`
  * checks wether the node is ignored by looking into the node map as well
  * discards the Violation or transforms it and passes it on

**pros**
* small API change
* rules become immediately simpler: no more paths/pointers generation, no more "ignore" checking

**cons**
* doesn't solve the Swagger traversal issue (yet)
* rules are executed even if they should be ignored

Personally I think this could be a very promising approach because if we at least believe that changing the `Violation` interface in this way is right (because it allows us to make rules simpler since `RulesValidator` has more information to play with after a `Violation` is generated). It would give us some immediate benefits while being a relatively small code change and would also be forwards compatible with future improvements, even ignoring parts of Swagger ahead of time and simpler traversal.

What do you think, @roxspring, @vadeg, @tkrop, @maxim-tschumak, @netme?

roxspring:
> Thank you everybody for the detailed inputs! It can feel cumbersome to discuss these topics for so long however it is hopefully going to pay off if we get it right.

Agreed - thanks all!❤️ 

Definitely agree that small incremental improvements are ideal wherever possible, and the proposed plan seems to deliver that: Formalizing the Violations/Results around something like JSON pointer seem to be a positive step forward, independent of `x-zally-ignore` support. 

A few minor questions/issues with the plan:

- If each Violation is to be associated with exactly one node, then presumably an unstated part of your plan is to let all the rules return a list of Violations rather than just one? (RulesValidator should already support this)
- Have you looked at getting the JsonNode based rules to play with this approach?
- I have some questions/comments around the OpenApiWalker but I guess they should wait for a WIP PR from that branch (hint hint).

mfellner:
Thanks @roxspring! In the end our next steps also depend on wether we want to prioritise OpenApi 3 or the other topics first (personally I would prefer the former).

> If each Violation is to be associated with exactly one node, then presumably an unstated part of your plan is to let all the rules return a list of Violations rather than just one?

You are right! Thanks for pointing that out.

> Have you looked at getting the JsonNode based rules to play with this approach?

The `OpenApiWalker` should work with JsonNodes too in principle but I haven't tried it yet.

> I have some questions/comments around the OpenApiWalker but I guess they should wait for a WIP PR from that branch (hint hint).

OK! Let's discuss here: https://github.com/zalando/zally/pull/693

vadeg:
@mfellner nice plan!

> * checks wether the node is ignored by looking into the node map as well
> * discards the Violation or transforms it and passes it on

I am not sure that it will work in all cases. @roxspring provided good example when it would not work because filtering should be done before executing rule but not after.

> Personally, I would prefer solutions that do not need more 3rd party libraries or that introduce yet another programming concept to Zally (e.g. aspects). I'm even skeptical using things from Spring because imo it should only be used on the web service layer of the application.

Yes, but this is very small and narrowed aspect :) `Spring` dependency is optional. I did it because it was faster to show. `Spring` can be safely replaced by `cglib` dependency. And yes, this is another 3rd party library. I think in compare to effort/value it brings more value with less effort rather than other solutions.

> Ideally we find a solution that only uses vanilla Java and Kotlin features.
Ideally yes and I also vote for this. But, unfortunately, I have not found anything useful in Kotlin. Java has Proxies (as you mentioned in #692 ) but it does not work with classes. Only with interfaces (that is one of the major reasons that `cglib` exists actually).

I would be happy if someone will find better solution for this except `cglib`.

maxim-tschumak:
Thanks for this important discussion @roxspring @vadeg @mfellner !

I agree with the plan described by @mfellner and think this is the best way to go. It's maybe not the nicest solution from the technical perspective, but this trade-off is very reasonable in our situation: this is a nice-to-have feature and we also want to discourage (#686) its extensive usage. Let's not over-engineer the solution :)

netme:
@maxim-tschumak , @mfellner , @vadeg thanks a lot for your comments.
@roxspring should we close this PR?

roxspring:
Yup. Long been clear that this PR wasn't going in but wanted to keep it open while the conversation was useful.

PR #686, opened 2018-03-16
Added AvoidXZallyIgnoreRule to explicitly highlight and discourage x-zally-ignore usage

roxspring:
Figure we might as well get the uncontroversial bit in - and work on the local `x-zally-ignore` support separately.

#446


maxim-tschumak:
:+1:

netme:
👍

Issue #656, opened 2018-03-05
Secure With OAuth2 Rule issues


roxspring:
SecureWithOAuth2Rule (104) is broken into 3 checks, arguably verifying 5 different things:
1. checkSecurityDefinitions (MUST)
  a. Ensure "oauth2" security definition exists
  b. Ensure schemes does not contain "HTTP"
2. checkPasswordFlow (SHOULD)
  a. Ensure that only application flow is used
3. checkUsedScopes (MUST)
  a. Ensure that all scoped operations refer to a defined scope
  b. Ensure that all operations are scoped.

(Aside: 3 above seems to leave an obvious gaping hole which may be worth checking: Ensure that all defined scopes are used by some operation)

Initially we disabled the rule for our installation since all of our services are defined as "schemes: [http, https]" and therefore everything was failing. We've recently hit situations where 3a would have avoided mistakes and so would like to enable the rule. Fixing our usage of schemes and therefore complying with 1b seems very reasonable BUT we disagree with 3b because we have some operations which need to be accessed by authenticated users but not with any particular scope. This leaves us with a puzzle because we'd like to enable part, but not all of the rule.

I can see a number of options available to resolve this and so wanted to consult the team:

1. Just add some "authenticated" scope to our infrastructure and run the original Zally rule untouched (Actively costing this option up in case it's viable)
2. Refactor 3 into two separate rules so that 3a can be used while 3b is ignored.
3. Split 3 into two separate checks and add support for ignoring specific checks.
4. Implement #495 or #446 so that rule 104 can be ignored for specific paths/operations where anonymous usage is needed (Pretty sure it would be welcome but not sure of the easiest route there!)

Particularly interested to hear what @maxim-tschumak and @netme would recommend, and whether the changes would be welcome upstream.

roxspring:
Chatting with a colleague I'm reminded that actually 3b above is implementing rule "Define and Assign Access Rights (Scopes) [105]" and isn't at all necessary to implement "Secure Endpoints with OAuth 2.0 [104]". Based on that I'm very tempted to split 3b out into a new rule implementation for 105 which we can ignore internally while having 104 enabled.

Thoughts?

maxim-tschumak:
> Based on that I'm very tempted to split 3b out into a new rule implementation for 105 which we can ignore internally while having 104 enabled.

I agree and support the proposed change.

PS. @roxspring to your original question: I would favor solution **3** (and eventually 4) to make Zally usage more convenient for your and any potential installations.

Issue #105, opened 2017-01-13
Adjust CLI tool to Handle New Errors Format


netme:
See https://travis-ci.org/zalando-incubator/zally/builds/191632853 for more details

netme:
@fmueller, I've cross-checked, everything works like before. Sorry for a false panic. Closing this issue.

fmueller:
You care, which is great. 👍

PR #687, opened 2018-03-16
Split SecureAllEndpointsWithScopesRule out from SecureWithOAuth2Rule

roxspring:
- Stopped `SecureWithOAuth2Rule` from checking that all end points are secured since that's defined as guideline 105 not 104
- Added `SecureAllEndpointsWithScopesRule` to represent guideline 104 and house the code removed above
- Added `SecureAllEndpointsWithScopesRule.checkDefinedScopeFormats(..)` to verify that scopes match expected formats to complete the implementation of guideline 104.

Discussed in #656

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/zalando/SecureAllEndpointsWithScopesRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/zalando/SecureAllEndpointsWithScopesRule.kt
@@ -1,0 +1,96 @@
+package de.zalando.zally.rule.zalando
+
+import com.google.common.collect.Sets
+import com.typesafe.config.Config
+import de.zalando.zally.rule.api.Check
+import de.zalando.zally.rule.api.Rule
+import de.zalando.zally.rule.api.Severity
+import de.zalando.zally.rule.api.Violation
+import io.swagger.models.Operation
+import io.swagger.models.Swagger
+import io.swagger.models.auth.OAuth2Definition
+import org.springframework.beans.factory.annotation.Autowired
+
+@Rule(
+        ruleSet = ZalandoRuleSet::class,
+        id = "105",
+        severity = Severity.MUST,
+        title = "Secure All Endpoints With Scopes"
+)
+class SecureAllEndpointsWithScopesRule(@Autowired rulesConfig: Config) {
+
+    private val pseudoScopes = rulesConfig.getStringList("${SecureAllEndpointsWithScopesRule::class.java.simpleName}.psuedo-scopes")
+
+    private val accessTypes = rulesConfig.getStringList("${SecureAllEndpointsWithScopesRule::class.java.simpleName}.access-types")
+
+    @Check(severity = Severity.MUST)
+    fun checkDefinedScopeFormats(swagger: Swagger): Violation? {
+        return swagger.securityDefinitions.orEmpty().flatMap { (schemeKey, scheme) ->
+            when (scheme) {
+                is OAuth2Definition -> {
+                    scheme.scopes.orEmpty().flatMap { (scope, description) ->
+                        checkDefinedScopeFormat(scope)?.let {
+                            listOf("securityDefinitions $schemeKey $scope: $it")
+                        } ?: emptyList()
+                    }
+                }
+                else -> emptyList()
+            }
+        }.takeIf { it.isNotEmpty() }?.let { Violation("Defined scopes should match an expected format", it) }
+    }
+
+    fun checkDefinedScopeFormat(scope: String): String? {
+        val components = scope.split('.')
+        return when {
+            components.size == 1 ->
+                if (components[0] in pseudoScopes) null
+                else "pseudo-scope '${components[0]}' should be one of $pseudoScopes"
+            components.size > 3 -> "scopes should have no more than 3 dot-separated components"
+            components.last() !in accessTypes -> "access-type '${components.last()}' should be one of $accessTypes"
+            else -> null
+        }
+    }
+
+    @Check(severity = Severity.MUST)
+    fun checkOperationsAreScoped(swagger: Swagger): Violation? {
+        val definedScopes = getDefinedScopes(swagger)
+        val hasTopLevelScope = hasTopLevelScope(swagger, definedScopes)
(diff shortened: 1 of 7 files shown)


server/src/main/java/de/zalando/zally/rule/zalando/SecureAllEndpointsWithScopesRule.kt
```diff
@@ -0,0 +1,96 @@
...
+                            listOf("securityDefinitions $schemeKey $scope: $it")
+                        } ?: emptyList()
+                    }
+                }
+                else -> emptyList()
+            }
+        }.takeIf { it.isNotEmpty() }?.let { Violation("Defined scopes should match an expected format", it) }
+    }
+
+    fun checkDefinedScopeFormat(scope: String): String? {
```
mfellner: Maybe it would make sense to only define a RegExp pattern for scopes in rules-config, e.g.

```regex
^([a-z-]+\.){1,2}(read|write)$
```

This would allow greater flexibility and wouldn't be tied to the quite Zalando-specific pattern described in the guidelines. On the other hand, writing a RegExp is more complicated.

server/src/test/resources/fixtures/no_must_violations.yaml
```diff
@@ -39,10 +39,10 @@ securityDefinitions:
     authorizationUrl: https://example.com/oauth2/dialog
     scopes:
       uid: Unique identifier of the user accessing the service.
-      api-is.api-repository.api-definition.write: Scope allowing the user to write API definitions.
+      api-is.api-repository.write: Scope allowing the user to write API definitions.
```
mfellner: The original scope was pretty weird to begin with but maybe `api-repository.api-definition.write` would be better? This would match the pattern of `<application-id>.<resource-id>.<access-type>`.

roxspring: Fair enough. Couldn't really make much sense of this specific example but suspect your interpretation and suggestion are the right thing to do.

server/src/main/java/de/zalando/zally/rule/zalando/SecureAllEndpointsWithScopesRule.kt
roxspring: I think if you want a single then you're going to have to make it even more complicated to cope with the `uid` pseudo-role:
```regex
^(uid)|(([a-z-]+\.){1,2}(read|write))$
```
If others think the added flexibility is worth it then I'm happy to make that change. Ironically I'm adding this rule but it still won't be flexible enough for me to use it though. We use Keycloak as our OAuth2 server and I don't think we get anything equivalent to a `uid` role to represent anybody logged in, instead it's perfectly valid to be protected by OAuth2 but require no scopes.

mfellner: Yes, the regex is slightly complicated, that's true... 😒 But on the other hand, having `psuedo-scopes` and `access-types` in the configuration is confusing if you haven't read rule http://zalando.github.io/restful-api-guidelines/#105 in detail.

@tkrop, what do you think?

As for the issue of endpoints without scopes (even `uid`), this sounds like rule `#105` simply doesn't apply to you then? Or would a separate "styling" rule for scopes be beneficial to you?

mfellner: Also calling @netme on the question of the regex :)

roxspring: We currently run with rule 104 disabled but would like to start using it, so my motivation for this change was to split out 105 which doesn't really work for us then disable 105 but enable 104.

mfellner: Makes sense. I think that we could introduce a separate naming rule for scope names to the API guidelines: http://zalando.github.io/restful-api-guidelines/#api-naming

Then rule 105 could only be about *having* scopes for an endpoint at all. But that probably goes too far at this time.

tkrop: @mfellner I think a regular expression would provide maximum flexibility in customizing this rule. The name should make very clear what is configured. :+1: from my side for a regexp.

mfellner:
@roxspring ok, so we've discussed this in the team and everyone agrees we would prefer a RegExp. Would you support this approach as well and update the PR accordingly? Other than that we're all set for this PR to be merged. 😀 Thanks!

roxspring:
All done and ready for review.

server/src/main/java/de/zalando/zally/rule/zalando/SecureAllEndpointsWithScopesRule.kt
```diff
@@ -39,20 +38,10 @@ class SecureAllEndpointsWithScopesRule(@Autowired rulesConfig: Config) {
...
+    private fun checkDefinedScopeFormat(scope: String): String? {
         return when {
-            components.size == 1 ->
-                if (components[0] in pseudoScopes) null
-                else "pseudo-scope '${components[0]}' should be one of $pseudoScopes"
-            components.size > 3 -> "scopes should have no more than 3 dot-separated components"
-            components.last() !in accessTypes -> "access-type '${components.last()}' should be one of $accessTypes"
-            else -> null
+            scopeRegex.matches(scope) -> null
+            else -> "scope '$scope' does not match regex '^(uid)|(([a-z-]+\\\\.){1,2}(read|write))\$'"
```
mfellner: We should probably use the string of `scopeRegex`! `"$scopeRegex"` should work.

roxspring: Doh! Bitten by a last minute rename of the field. Fix coming shortly...

roxspring: All done.

maxim-tschumak:
:+1:

netme:
👍

PR #692, opened 2018-03-23
Proxy for Swagger object

vadeg:
`SwaggerProxy` class which allows to filter Swagger model entities using provided predicate.

## How it works

`SwaggerProxy` creates a proxy for `Swagger` model instance using [cglib](https://github.com/cglib/cglib) library. Proxy intercepts all method calls, redirects them to origin and get result. 
Call result is checked whether is had `getVendorExtensions` method or not. If yes then vendorExtensions are extracted and predicate is applied to them. Predicate returns `Boolean` value. 
If predicate returns `true` then result is returned back to the caller. 
If predicate returns `false` then result will be:
* `null` in case of POJO
* filtered `HashMap` in case of `Map`
* filtered `ArrayList` in case of `List`
* filtered `HashSet` in case of `Set`

### Pros
+ `SwaggerProxy` can be easily used with `OpenAPI` object (OpenAPI v3) as well. The only thing that should be changed is vendor extensions method name from `getVendorExtensions` to `getExtensions`.
+ Predicate can be applied to any part of object model regardless where it is located in object graph.
+ Proxies are build on demand only. That means if there are big model but rules uses only part of it then proxies will be built only for used parts of model.

### Cons
- It uses reflection what can slowdown process a bit. It can be easily solved with caching because Swagger model is immutable from rules perspective and rules do not change the model.

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/model/SwaggerProxy.java
+++ b/server/src/main/java/de/zalando/zally/rule/model/SwaggerProxy.java
@@ -1,0 +1,137 @@
+package de.zalando.zally.rule.model;
+
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import io.swagger.models.Swagger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.cglib.proxy.Enhancer;
+import org.springframework.cglib.proxy.MethodInterceptor;
+import org.springframework.cglib.proxy.MethodProxy;
+
+/**
+ * CGLib proxy for filtering objects using <code>getVendorExtensions</code> method and provided predicate.
+ */
+public class SwaggerProxy {
+
+
+    private static final Logger LOG = LoggerFactory.getLogger(SwaggerProxy.class);
+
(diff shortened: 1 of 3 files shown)


mfellner:
Interesting. Thanks for the idea! However, I would be strongly opposed to using something like cglib in Zally. Vanilla Java and Kotlin features really ought to be enough. It would make the project a lot more complicated.

Although I've become a lot more skeptical of the whole proxying approach in general, have you looked at native Java proxies already? https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Proxy.html

roxspring:
Nice POC!

vadeg:
@mfellner native Java proxies can not work with classes. It can work only with interfaces. `Swagger` object is a class. To use java proxy you have to define interfaces for all classes in model. That is why cglib is used. 
I have not found anything in Kotlin what can solve this issue in more a elegant way. All possible solutions like property/class delegates look even more complicated. If someone can point me that would be nice. 
Also, I have used `cglib` embedded in `spring`. It can be changed easily to use external `cglib` library (`spring` just merged it inside in a special package) but replacing imported packages. I would say spring dependency is optional here.

netme:
@vadeg nice idea, are you planning to proceed with this implementation?

vadeg:
@netme yes if it will be accepted. We had a discussion with @mfellner and there was no clear whether it is acceptable to use proxy or not.

mfellner:
@vadeg I'm in the process of preparing a revised PR for OpenAPI 3 support that includes an alternative implementation for a Proxy. https://github.com/zalando/zally/pull/706 I would appreciate your feedback. Maybe we can close this PR in favour of the other one.

Issue #651, opened 2018-02-26
Examples validated against schema


roxspring:
API definitions allow schema objects to have an example, and response objects to have a list of examples. I think OpenAPI 3.0 allows examples in even more places. It seems obvious that Zally should be able to validate that these example match the defined schema. (Obviously a few preconditions might need to be applied - that we're dealing with json/xml mime types, maybe it only makes sense to validate the successful responses??)

Seemingly OpenAPI just builds up a json-schema for the structures, and Zally already does some schema validation, so all the pieces appear to be present. Has this been attempted and turns out to be difficult for reasons I haven't spotted? Or should this be easy to achieve and worth working on?

PR #693, opened 2018-03-23
(WIP) OpenApi 3 support

mfellner:
- [x] translate `Swagger` into `OpenApi` objects
- [x] update all rules to use `OpenApi` (or `Swagger` as fallback)
- [x] add `OpenApiWalker` to generate node-to-meta map
- [ ] change `Violation` to include violating `Object` nodes
- [ ] update `RulesValidator`s to handle updated `Violation`s


server/src/main/java/de/zalando/zally/util/OpenApiWalker.java
```diff
@@ -0,0 +1,228 @@
...
+import javax.annotation.Nullable;
+import javax.validation.constraints.NotNull;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.*;
+
+/**
+ * OpenApiWalker traverses a Swagger or OpenApi object.
+ */
+public class OpenApiWalker {
```
roxspring: Both the name and the javadoc show how this class works, but neither give much clue as to what the class is or achieves. I suggest that we rename to something like `JsonPointerCache` because that seems to be the focus (and it's not directly related to anything OpenAPI anyway).

mfellner: I agree the current name is not great. Although it's definitely more than a cache for JSON pointers already since it can hold meta information like the zally-ignore marker. It's more like a `ReverseAST` or something.

server/src/main/java/de/zalando/zally/util/OpenApiWalker.java
```diff
@@ -0,0 +1,228 @@
...
+  private OpenApiWalker() {
+  }
+
+  /**
+   * Meta information for an object node in a Swagger or OpenApi object.
+   */
+  public static class Meta {
+    private final String pointer;
```
roxspring: Commenting on the first occurrence, but point applies throughout...

I'm quite excited at the prospect of having a formalized structured value to pass around and reason about rather than free-form values that can be abused... so it's disappointing to see that the pointer here is represented as an absusable `String` rather than something strongly typed. Looks like we already have jackson's `JsonPointer` on the class-path, could we just reuse that instead?

server/src/main/java/de/zalando/zally/util/OpenApiWalker.java
```diff
@@ -0,0 +1,228 @@
...
+
+  /**
+   * Traverse a Swagger or OpenApi object tree and construct
+   * a map of object nodes to meta information objects.
+   *
+   * @param object Swagger or OpenApi instance.
+   * @param ignore List of classes in the traversable object that should be ignored.
+   * @return IdentityHashMap of object nodes to meta information objects.
+   */
+  public static Map<Object, Meta> walk(Object object, Collection<Class<?>> ignore) {
```
roxspring: Rather than a static utility method building a map, I think it would be neater to encapsulate the map in the object itself and then provide lookup methods so that users don't have to understand the Meta / Node objects. This method might still be called `walk` or be renamed to `add` or `populate` but likely return either void or this rather than the internal map.

roxspring: What's the use-case for the ignore parameter? I see that in tests its populated with ArrayModel, ComposedModel, ModelImpl but couldn't really get to understand why.

server/src/main/java/de/zalando/zally/util/OpenApiWalker.java
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  private static String getterNameToPathName(String name) {
+    String s = name.substring(3); // `get` is first 3 characters
+    return s.substring(0, 1).toLowerCase().concat(s.substring(1));
+  }
+
+  private static String getVendorExtension(Object object, String extension) {
+    try {
+      Method m = object.getClass().getMethod("getVendorExtensions");
```
roxspring: Think "getVendorExtensions" is right for Swagger but for OpenAPI this should be "getExtensions".

server/src/main/java/de/zalando/zally/util/OpenApiWalker.java
mfellner: This tells OpenApiWalker not to go down into sub-trees of `Swagger` that use those classes. In this case we want to ignore everything inside `#/definitions` (basically all the JSON schemas).

I guess this could be simplified, more intuitive?

roxspring: But why do we want to ignore those?

I have another branch working on #651 where validation within those model objects is precisely what I'm doing, and I'd hope that violations would be referencing something suitably deep in the tree!

server/src/main/java/de/zalando/zally/util/ReverseAst.java
```diff
@@ -0,0 +1,257 @@
...
+          if (value != null) {
+            String pointer = p.concat("/").concat(String.valueOf(i));
+            nodes.push(new Node(value, pointer));
+          }
+        }
+      } else {
+        // Check if an ignore extension exists at this position.
+        String ignoreExtension = getVendorExtension(o, Marker.X_ZALLY_IGNORE.key);
+        if (ignoreExtension != null) {
+          marker = new Node(ignoreExtension, ignoreExtension, Marker.X_ZALLY_IGNORE); // Set the current marker.
```
roxspring: Shouldn't this be creating a pointer relative to `p` like the other pointer creations?

(Admittedly haven't thought about this in detail, just noticed when experimenting with `JsonPointer` instead of `String`)

server/src/main/java/de/zalando/zally/util/ReverseAst.java
```diff
@@ -0,0 +1,257 @@
...
+        markers.pop(); // remove the marker
+        if (nodes.isEmpty()) {
+          continue;
+        } else {
+          node = nodes.pop(); // get the next node
+        }
+      }
+
+      final Object o = node.object;
+      final String p = node.pointer;
```
vadeg: Using `StringBuilder` instead of `String` here and replacing `String.concat` with `StringBuilder.append` will eliminate useless `String` object creation done by `String.concat`.

server/src/main/java/de/zalando/zally/util/ReverseAst.java
```diff
@@ -0,0 +1,257 @@
...
+      if (!node.shouldSkip()) {
+        map.put(o, new Meta(p, marker));
+      }
+
+      if (ignore.contains(o.getClass())) {
+        continue;
+      }
+      if (o instanceof String) {
+        continue;
+      } else if (o instanceof Integer) {
```
vadeg: All these checks for `Integer`, `Float`, `Double`, `String`, `Enum` and `Boolean` can be removed if all these classes will be added to `ignore` collection. For example
```
Set<Class> ignoreSet = new HashSet<>(Arrays.asList(
        String.class, Integer.class, Float.class, Double.class, Boolean.class, Enum.class
));
ignoreSet.addAll(ignore);
/* .... */
/* Some code */
/* .... */
if (ignoreSet.contains(o.getClass())) {
    continue;
}

```

server/src/main/java/de/zalando/zally/util/ReverseAst.java
```diff
@@ -0,0 +1,257 @@
...
+          Object key = entry.getKey();
+          Object value = entry.getValue();
+          if (key instanceof String && value != null) {
+            String pointer = p.concat("/").concat(rfc6901Encode((String) key));
+            nodes.push(new Node(value, pointer));
+          }
+        }
+      } else if (o instanceof List) {
+        List<?> l = (List<?>) o;
+        for (int i = 0; i < l.size(); i += 1) {
```
vadeg: Why not just `i++`?

server/src/main/java/de/zalando/zally/util/ReverseAst.java
```diff
@@ -0,0 +1,257 @@
...
+  private static String getVendorExtension(Object object, String extension) {
+    for (Method m : object.getClass().getDeclaredMethods()) {
+      if (EXTENSION_METHOD_NAMES.contains(m.getName())) {
+        try {
+          Object extensions = m.invoke(object);
+          if (extensions instanceof Map) {
+            return (String) ((Map) extensions).get(extension);
+          }
+        } catch (ReflectiveOperationException e) {
+          return null;
```
vadeg: I would not avoid exception. In this case exception happened after method was found and invoked. This means something really went wrong. It would be better to know about that.

server/src/main/java/de/zalando/zally/util/ReverseAst.java
```diff
@@ -0,0 +1,257 @@
...
+    }
+    return null;
+  }
+
+  public boolean isIgnored(Object key) {
+    Meta meta = this.map.get(key);
+    return meta != null && Marker.X_ZALLY_IGNORE.equals(meta.marker);
+  }
+
+  @Nullable
```
vadeg: I think the `Optional` is a better solution rather than `@Nullable` annotation. I can be rewritten like this
```
  public Optional<String> getIgnoreValue(Object key) {
    return Optional
            .ofNullable(this.map.get(key))
            .filter(meta -> Marker.X_ZALLY_IGNORE.equals(meta.marker))
            .map(meta -> meta.markerValue);
  }
```

server/src/main/java/de/zalando/zally/util/ReverseAst.java
```diff
@@ -0,0 +1,257 @@
...
+public class ReverseAst {
+  private static Collection<String> EXTENSION_METHOD_NAMES = new HashSet<>(Arrays.asList(
+      "getVendorExtensions",
+      "getExtensions"
+  ));
+
+  /**
+   * Meta information for an object node in a Swagger or OpenApi object.
+   */
+  private static class Meta {
```
vadeg: IDE shows that class has two spaces indentation instead of four.

server/src/main/java/de/zalando/zally/util/ReverseAst.java
```diff
@@ -0,0 +1,257 @@
...
+  /**
+   * Creates a new instance of ReverseAst. Traverses a Swagger or OpenApi object tree
+   * and constructs a map of object nodes to meta information objects.
+   *
+   * @param object Swagger or OpenApi instance.
+   * @param ignore List of classes in the traversable object that should be ignored.
+   * @return ReverseAst instance.
+   * @throws ReverseAstException If an error occurs during reflection.
+   */
+  public static ReverseAst create(Object object, Collection<Class<?>> ignore) throws ReverseAstException {
```
tkrop: As you already have a dependency to OpenAPI 3.0 and Swagger 2.0 I would suggest to make the ignore class collection to an internal feature of the `ReverseAstBuilder`.

Rule of thumb: do not distribute common concerns into different places.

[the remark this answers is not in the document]
tkrop: Absolutely agreed this builder method has way to much responsibilities to be understandable. You should refactor the method into a `ReverseAstBuilder`, that clearly distributes different aspects into methods. This would allow much cleaner unit testing of the functionality. Let me ask one question: How good is your branch and code coverage here?

server/src/main/java/de/zalando/zally/util/ReverseAst.java
tkrop: In my opinion merging is only an acceptable approach, if `ignore` is a builder private concept. Else I would suggest to at least check against to different sets.

server/src/main/java/de/zalando/zally/util/ReverseAst.java
```diff
@@ -0,0 +1,257 @@
...
+      } else if (o instanceof Float) {
+        continue;
+      } else if (o instanceof Double) {
+        continue;
+      } else if (o instanceof Boolean) {
+        continue;
+      } else if (o instanceof Enum) {
+        continue;
+      } else if (o instanceof Map) {
+        for (Map.Entry<?, ?> entry : ((Map<?, ?>) o).entrySet()) {
```
tkrop: For me it sounds as the map should be of type `Map<String, ?>`.

server/src/main/java/de/zalando/zally/util/ReverseAst.java
tkrop: This sounds like a wired hack.

server/src/main/java/de/zalando/zally/util/ReverseAst.java
mfellner: Java Optionals are not great for Kotlin interop, that's why I chose `null`.

server/src/main/java/de/zalando/zally/util/ReverseAst.java
mfellner: It's just a convention I've been following because `++` can be ambiguous in some contexts (https://youtu.be/_EANG8ZZbRs?t=42m23s).

server/src/main/java/de/zalando/zally/util/ReverseAst.java
mfellner: Fixed this with refactoring.

server/src/main/java/de/zalando/zally/util/ReverseAst.java
mfellner: OK.

server/src/main/java/de/zalando/zally/util/ReverseAst.java
mfellner: OK.

server/src/main/java/de/zalando/zally/util/ReverseAst.java
mfellner: I refactored it in terms of a "Builder" pattern now...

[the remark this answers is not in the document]
mfellner: I agree some refactoring can reduce complexity and increase testability and maintainability of this code here. 

However in general I think that fighting code complexity by breaking up code into separate functions/methods can be a mistake because code coverage often is a lot more complex: https://en.wikipedia.org/wiki/Code_coverage

* I would only support refactoring into  **pure functions** because shared state in all forms - especially within an object, mutated by API methods - is always making code worse.
* Even with pure functions code complexity can actually increase and certain kinds of condition coverage and parameter coverage can decrease.

server/src/main/java/de/zalando/zally/util/ReverseAst.java
mfellner: I wouldn't cast an Object that was just proven to only be instance of Map to Map<String, ?>. The assertion wether the key is a String is done three lines below.

server/src/test/java/de/zalando/zally/util/ast/ReverseAstTest.kt
```diff
@@ -0,0 +1,95 @@
+package de.zalando.zally.util.ast
+
+import de.zalando.zally.util.ResourceUtil.resourceToString
+import io.swagger.parser.SwaggerParser
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import java.util.*
+
+class ReverseAstTest {
```
tkrop: Would be nice to have some more tests for the extensive branch logic to guide the expected behavior more carefully.

mfellner:
After hitting some roadblocks while working on this branch of the code I decided to start over. Closing this PR in favour of https://github.com/zalando/zally/pull/706.

PR #703, opened 2018-04-07
PR-636: Implement Meta information rule.

vadeg:
Implement [rule 218](http://zalando.github.io/restful-api-guidelines/#218) from Zalando RESTful Guideline.

CLI integration tests for remote files fixed. Before files were downloaded from Github (probably master branch) what provides inconsistency between tests and current codebase. Simple `http` server was added for integration tests which emulates remote source.


server/src/main/java/de/zalando/zally/rule/zalando/ApiMetainformationRule.kt
```diff
@@ -0,0 +1,69 @@
...
+    severity = Severity.MUST,
+    title = "Contain API Meta Information"
+)
+class ApiMetainformationRule {
+
+    private val versionRegex = """^\d+.\d+.\d+-?[\w+]*""".toRegex()
+    private val basePath = "/info"
+    private val noInfoBlock = "Info block should be provided"
+
+    @Check(severity = Severity.SHOULD)
```
maxim-tschumak: I would prefer to have multiple checks in place, one per each info field. This has some benefits
* The number of violations is always correct (in the current implementation we return max 1 violation, even when there are multiple fields missing)
* User becomes more information about what she has to fix, and don't have to run linter multiple times.

vadeg:
@maxim-tschumak @netme don't you think that Codacity checks like this is too much? :)
violation

BTW I have found almost no comments in previously added code.

roxspring:
GitHub not allowing me to reply to @vadeg's comment for some reason but wanted to qualify my 👍: 

I too have been a little annoyed at having to add doc comments when so few existed before in the main code tree, but I can see that it helps in the transition from the main codebase being largely undocumented towards being fully documented, which seems worthwhile so I'm happy to put up with that annoyance.

The really frustration though, is being forced to add such comments to the test code - IMHO the test target, conditions and expected result should all be clear from the method name; which leaves very little value in adding a comment. Good clear test code ends up getting littered with unhelpfully duplicate doc comments cluttering the source and making readability worse:

```kotlin
/** Tests for Blah. */
class BlahTest {
    @Test
    /** Test that blah with invalid parameters throws exception */
    fun `blah with invalid parameters throws exception` {
        // the test
    }
}
```

Would be very happy if we could turn some of these rules down, particularly for the test tree!

maxim-tschumak:
@vadeg @roxspring We definitely have to improve our Codacy configuration - it should help us and not disturb the work. Feel free to make a proposal.

maxim-tschumak:
:+1:

vadeg:
I have an impression that Codacity just picks random checks each time when code changes.

literal-duplication

I wonder why this check was green before?

roxspring:
This rule has also bugged me: It spots multiple occurrences of a string literal but doesn't tell you what it was!!!

Again, unlike main code, I find that string duplication is often healthy in test sources. Guess I ought to read up on Codacy configuration and make a proposal 😄 

(looks to me like "$basePath/version" is the/an offending duplicate literal in this case)

server/src/main/java/de/zalando/zally/rule/zalando/ApiMetainformationRule.kt
```diff
@@ -0,0 +1,108 @@
...
+            Violation("Description is not defined", listOf("$basePath/description"))
+        } else null
+
+    /**
+     * Check version property
+     */
+    @Check(severity = Severity.MUST)
+    fun validateInfoVersion(swagger: Swagger): Violation? {
+        val version = swagger.info?.version
+        return when (version) {
```
roxspring: Odd to see a when devolve into an else if else if!? How about removing the version parameter:

```kotlin
return when {
    version==null, version.isBlank() -> Violation("Version is not defined", listOf("$basePath/version"))
    !versionRegex.matches(version) -> Violation("Version is not following Semver rules", listOf("$basePath/version"))
    else -> null
}
```

server/src/main/java/de/zalando/zally/rule/zalando/ApiMetainformationRule.kt
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+    /**
+     * Check contact name
+     */
+    @Check(severity = Severity.MUST)
+    fun validateContactName(swagger: Swagger): Violation? =
+        if (swagger.info?.contact?.name.isNullOrBlank()) {
+            Violation("Contact name is empty", listOf("$basePath/contact/name"))
+        } else null
```
roxspring: Each of these uses of `?.` effectively hides an if/else, I'm pretty sure we could hide this last one two with a `?.let`:
```kotlin
swagger.info?.contact?.name.isNullOrBlank()?.let{
  Violation("Contact name is empty", listOf("$basePath/contact/name"))
}
```

server/src/main/java/de/zalando/zally/rule/zalando/ApiMetainformationRule.kt
vadeg: I "like" Kotlin here. This is would not compile
```
version.isNullOrBlank() ->
    Violation("Version is not defined", listOf("$basePath/version"))
!versionRegex.matches(version) ->
    Violation("Version is not following Semver rules", listOf("$basePath/version"))
else -> null
```
but has the same meaning.

server/src/main/java/de/zalando/zally/rule/zalando/ApiMetainformationRule.kt
vadeg: It would not work like that. `let` will be called only if value is not null. But in this case violation should be when it is null. 
And IDE is complaining about useless `?` in `?.let`.

roxspring: Sorry - think I was getting sleepy and not thinking clearly when adding this comment.

maxim-tschumak:
@vadeg Do you want to resolve the conflict so that we can merge? Or should I take over?

vadeg:
@maxim-tschumak I will fix it.

vadeg:
@maxim-tschumak I have merged PR but Codacy is complaining about comments in tests again.

server/src/main/resources/api/zally-api.yaml
```diff
@@ -20,6 +20,7 @@ info:
  contact:
    name: Team API Management
    email: team-api-management@zalando.de
+   url: "http://tech.zalando.de"
```
maxim-tschumak: Let's also increase the patch version of the specification

server/src/test/java/de/zalando/zally/rule/zalando/ApiMetainformationRuleTest.kt
```diff
@@ -0,0 +1,113 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+/**
+ * Tests for Rule 218
+ */
+class ApiMetainformationRuleTest {
+
+    val rule = ApiMetainformationRule()
+    val swagger = getFixture("218.yaml")
```
maxim-tschumak: We try to use minimal specifications for the tests (since recently :)) and avoid using test fixtures. Please have a look at [this example](https://github.com/zalando/zally/blob/master/server/src/test/java/de/zalando/zally/rule/zalando/ApiAudienceRuleTest.kt#L42). How do you feel about switching to the new pattern?

maxim-tschumak:
@vadeg Great! We can ignore Codacy - it's not mandatory. But please have a look at my comments.

server/src/test/java/de/zalando/zally/rule/zalando/ApiMetainformationRuleTest.kt
vadeg: > We try to use minimal specifications for the tests

Minimal specifications are good. 

> avoid using test fixtures
I do not think that writing specs directly in the code will help a lot. Writing spec in file is much easier (formatting for example) and easy to read.
In some cases it make sense to avoid files but in some cases it is better to use files instead.

maxim-tschumak:
:+1:

tkrop:
:+1:

PR #706, opened 2018-04-16
OpenAPI 3 Support (WIP)

mfellner:
Related to #204

- [x] Support OpenAPI 3 with a fallback to Swagger 2.
- [x] Reverse AST to convert OpenApi/Swagger nodes to JSON pointers and to look up "x-zally-ignore" information for sub-trees of nodes.
- [x]  MethodCallRecorder to automatically generate JSON pointers while accessing (potentially null) properties of OpenApi/Swagger objects.
- [x] Context and ContextRulesValidator for simpler and more robust rules.


server/src/main/java/de/zalando/zally/util/ast/ReverseAst.java
```diff
@@ -0,0 +1,98 @@
...
+     * Find a pointer for a given Node (typically from another AST).
+     * <p>
+     * A pointer can be found if the object value of the given root node or
+     * of any of its children exists in this AST.
+     *
+     * @param key Node and object value from another AST.
+     * @return JSON pointer or null.
+     */
+    @Nullable
+    public String getPointer(Node key) {
```
roxspring: I'm still uncomfortable with all these json pointers being represented as Strings rather than being strongly typed. Jackson core, already on the class path, already has a JsonPointer class which looks suitable. If for some reason it's not then implementing our own would be better than throwing all these Strings around which might get appended or manipulated in unsafe ways.

mfellner:
@netme I'm still in the process of changing some things and porting a few rules over to OpenAPI 3. Once that is done it should be possible to gradually migrate rules to OpenAPI 3 and at that point we can probably think about merging the PR! ETA Thursday this week.

server/src/main/java/de/zalando/zally/util/ast/ReverseAst.java
mfellner: I'm going to reduce the reliance on strings to a minimum however I don't see why JsonPointer in Jackson would be helpful. The point of the string JSON pointers in this new code is simply to be inserted into the Violations returned by each rule. We don't actually read them or perform any logic on them.
(There is one exception where we need to manually translate OpenAPI 3 JSON pointers to Swagger JSON pointers for backwards compatibility. This is done with RegExes right now.)

server/src/main/java/de/zalando/zally/apireview/ApiReview.java
```diff
@@ -81,7 +81,11 @@ public ApiReview(ApiDefinitionRequest request, String userAgent, String apiDefin
 
         this.name = ApiNameParser.extractApiName(apiDefinition);
         this.ruleViolations = violations.stream()
-            .map(v -> new RuleViolation(this, v.getRule().getClass().getSimpleName(), v.getViolationType(), v.getPaths().size()))
+            .map(v -> new RuleViolation(
+                this,
+                String.format("%s (%s)", v.getRule().title(), v.getRule().id()),
+                v.getViolationType(),
+                1))
```
vadeg: Looks like a magic number. Maybe it is better to create constant here?

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.java
```diff
@@ -0,0 +1,186 @@
...
+    static boolean isGenericContainer(Object o) {
+        return o instanceof Collection || o instanceof Map;
+    }
+
+    @SuppressWarnings("unchecked")
+    static <T> T createInstance(Class<T> c) throws MethodCallRecorderException {
+        if (c.isAssignableFrom(Map.class)) {
+            return (T) new HashMap<>();
+        }
+        if (c.isAssignableFrom(Collection.class)) {
```
vadeg: In this case `Set` will be converted to `ArrayList`.

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.java
```diff
@@ -0,0 +1,186 @@
...
+            return (T) new ArrayList<>();
+        }
+        try {
+            Constructor<T> constructor = c.getConstructor();
```
vadeg: This method will fail if there are no default constructor in class.

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.java
```diff
@@ -0,0 +1,186 @@
...
+        throw new MethodCallRecorderException(m.getReturnType().toString());
+    }
+
+    static String toPointer(String s) {
+        return "/".concat(rfc6901Encode(getterNameToPointer(s)));
+    }
+
+    static String toPointer(Method m, Object... arguments) {
+        String s = m.getName();
+        if (arguments.length > 0) {
```
vadeg: `NPE` might be if `arguments` will be `null`.

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.java
```diff
@@ -0,0 +1,186 @@
...
+    private final T proxy;
+    private String currentPointer;
+    private final Set<String> skipMethods;
+    private final IdentityHashMap<Object, String> objectPointerCache;
+    private final IdentityHashMap<Object, IdentityHashMap<Method, String>> methodPointerCache;
+
+    public MethodCallRecorder(T object) {
+        this.currentPointer = "#";
+        this.skipMethods = new HashSet<>();
+        this.objectPointerCache = new IdentityHashMap<>();
```
vadeg: Initialization can be moved into declaration.

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.java
```diff
@@ -0,0 +1,186 @@
...
+        this.methodPointerCache = new IdentityHashMap<>();
```
vadeg: Initialization can be moved into declaration.

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.java
```diff
@@ -0,0 +1,186 @@
...
+
+    private final T proxy;
+    private String currentPointer;
+    private final Set<String> skipMethods;
+    private final IdentityHashMap<Object, String> objectPointerCache;
+    private final IdentityHashMap<Object, IdentityHashMap<Method, String>> methodPointerCache;
+
+    public MethodCallRecorder(T object) {
+        this.currentPointer = "#";
+        this.skipMethods = new HashSet<>();
```
vadeg: Initialization can be moved into declaration.

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.java
```diff
@@ -0,0 +1,186 @@
...
+
+    @Nonnull
+    public T getProxy() {
+        return this.proxy;
+    }
+
+    @SuppressWarnings("unchecked")
+    private <U> U createProxy(U object, Method parent) {
+        MethodInterceptor interceptor = createMethodInterceptor(object, parent);
+        ProxyFactory factory = new ProxyFactory();
```
vadeg: I would say having dependency from `cglib` is more cleaner rather than having dependency from `spring-aop` which in the end uses the same `cglib` proxy as I used in #692.

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.java
```diff
@@ -0,0 +1,186 @@
...
+                    return null;
+                }
+                if (isGenericContainer(object)) {
+                    Class<?> genericReturnValueType = getGenericReturnValueType(parent);
+                    if (genericReturnValueType.equals(Object.class)) {
+                        return null;
+                    }
+                    return createProxy(createInstance(genericReturnValueType), m);
+                }
+                return createProxy(createInstance(returnType), m);
```
vadeg: 1. Why proxy gets created if `result == null`? This is looks like undeclared behaviour. If original instance returns `null` then proxy should also return `null`.
2. Creating instance of method return type is a very dangerous and error-prone because proxy does not have information which if there a public constructor exists / does constructor has an arguments and etc.

server/src/main/java/de/zalando/zally/util/ast/Node.java
```diff
@@ -0,0 +1,52 @@
...
+        return this;
+    }
+
+    Node setChildren(Collection<Node> children) {
+        this.children = children;
+        return this;
+    }
+
+    @Nullable
+    Node getParent() {
```
vadeg: Maybe `Optional`?

server/src/main/java/de/zalando/zally/util/ast/ReverseAst.java
```diff
@@ -0,0 +1,83 @@
...
+    private final Map<String, Node> pointersToNodes;
+
+    ReverseAst(Map<Object, Node> objectsToNodes, Map<String, Node> pointersToNodes) {
+        this.objectsToNodes = objectsToNodes;
+        this.pointersToNodes = pointersToNodes;
+    }
+
+    @Nullable
+    public String getPointer(Object key) {
+        Node node = this.objectsToNodes.get(key);
```
vadeg: Just a matter of taste :)

```
        return Optional.ofNullable(this.objectsToNodes.get(key))
                .map(v -> v.pointer)
                .orElse(null);
```

server/src/main/java/de/zalando/zally/util/ast/ReverseAstBuilder.java
```diff
@@ -0,0 +1,193 @@
...
+     * @throws ReverseAstException If an error occurs during reflection.
+     */
+    @Nonnull
+    public ReverseAst<T> build() throws ReverseAstException {
+        while (!nodes.isEmpty()) {
+            Node node = nodes.pop();
+
+            if (!ignore.contains(node.object.getClass())) {
+                Collection<Node> children;
+                if (node.object instanceof Map) {
```
vadeg: `Set` also should be supported?

server/src/main/java/de/zalando/zally/util/ast/Util.java
```diff
@@ -0,0 +1,31 @@
+package de.zalando.zally.util.ast;
+
+import java.util.Arrays;
+import java.util.List;
+
+final class Util {
+    static final List<Class<?>> PRIMITIVES = Arrays.asList(
```
vadeg: `Set` is enough here I guess.

server/src/main/java/de/zalando/zally/apireview/ApiReview.java
mfellner: `occurrence` is now deprecated because there should be only 1 pointer/path per violation. I changed this line back to use the paths size but marked the property as deprecated inside `RuleViolation`.

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.java
mfellner: Yes. In this case we cannot create a new instance.

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.java
mfellner: I don't know, using the high-level interface abstraction seems cleaner to me. Plus, ProxyFactory is really simple to use.

server/src/main/java/de/zalando/zally/util/ast/ReverseAst.java
roxspring: But throughout this class you end up doing json pointer specific string manipulation which may not be clearly understood by all maintainers, and similarly people working on the main guts of Zally might well benefit from clearly understanding that Violations reference json pointers, and not just random string content. Seems crazy to me to have such a specifically structured string, with an appropriate type already in the classpath, but choose to model it using a generic String.

vadeg:
@mfellner will you continue to work on this PR?

maxim-tschumak:
This is a big piece of work and it brings us closer to full support of OpenAPI 3. Let's merge this one!
@roxspring Are you OK with merging this first and refactoring to `JsonPointer` afterward?

server/src/main/java/de/zalando/zally/util/ast/Node.java
roxspring: Rereading this comment reminds me to ask... is there a reason for this class and several others to be written in Java? Seems to me that we're heading towards all of Zally server being written in Kotlin for consistency, is there a reason to hold back?

roxspring:
@maxim-tschumak, no objection from me.

Let's get this PR in soon and make any further improvements in future (hopefully smaller!?) PRs.

maxim-tschumak:
:+1:

maxim-tschumak:
@netme @mfellner @vadeg let's merge?

vadeg:
@maxim-tschumak fine for me 👍

maxim-tschumak:
Thanks for having a look!

tfrauenstein:
👍

tfrauenstein:
👍

Issue #318, opened 2017-04-28
Try to remove jbossdna.inflector dependency


netme:
We are currently using `jbossdna.inflector` in 2 rules:
- PluralizeNamesForArraysRule
- PluralizeResourceNamesRule

We need to search for an alternative due to the following reasons:
- We are keeping dependency inside current repository
- It violates MIT license and we need to keep an exception in our license.

Proposal: find another library which can pluralize / singularize english words.

cogitor:
I've noticed that the current implementation of `WordUtils.isPlural` would return false if a plural form of a word is the same as the singular form (due to the `plural.equals(word) && !singular.equals(word)` check).

So to refactor this, I've added additional unit tests to check if `jbossdna.inflector` covers invariant plurals like "deer", "sheep", "fish", "salmon", "trout", "hovercraft", "spacecraft" and it's not. A set of tests that's failing is:

```kotlin
 @Test
 fun invariableWordsPluralized() {
     assertTrue(isPlural("sheep"))
     assertTrue(isPlural("deer"))
     assertTrue(isPlural("fish"))
     assertTrue(isPlural("cod"))
     assertTrue(isPlural("salmon"))
     assertTrue(isPlural("information"))
     assertTrue(isPlural("aircraft"))
 }
```

As one alternative to `jbossdna.inflector`, I've tested also https://github.com/atteo/evo-inflector, but it was failing generating even some simple plurals like "dogs"

roxspring:
How about PlingStemmer from http://mpii.de/yago-naga/javatools licensed under Creative Commons Attribution License 3.0?

https://github.com/roxspring/zally/tree/issue-318-isplural if interested.

PR #713, opened 2018-05-17
Replaced jbossdna.inflector with javatools.parsers.PlingStemmer

roxspring:
`PlingStemmer` appears to function similarly to `Inflector` but with a Creative Commons rather than LGPLv2 license - not totally clear to me whether that solves the concerns raised in #318 or just shifts them - but opening a PR forces the discussion!

I've not modified the LICENSE file but am happy to do so with some guidance.


server/build.gradle
```diff
@@ -48,6 +48,7 @@ jar {
 repositories {
     jcenter()
     mavenCentral()
+    maven { url "https://raw.github.com/antoine-tran/maven-repo/master/releases" }
```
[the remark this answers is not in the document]
roxspring: Yeah - wondered about that when reviving this experimental branch - will see whether 1.0.1 is good enough.

maxim-tschumak:
Great!

maxim-tschumak:
:+1:

tfrauenstein:
👍

maxim-tschumak:
:+1:

server/src/test/java/de/zalando/zally/util/WordUtilTest.java
```diff
@@ -48,5 +48,6 @@ public void negativeCasePluralized() {
     @Test
     public void specialCasePluralized() {
         assertTrue(isPlural("vat")); //whitelisted
+        assertTrue(isPlural("apis")); //whitelisted
```
mkulak: should be `api`?

roxspring: I know that I hear both "API" and "APIs" when talking about API in plural around here (in a UK office of mostly English developers), so I think it might be one of those odd cases where both are valid and therefore this is a reasonable change to make and test. In fact the plural initialisms here suggest that "APIs" would be preferred: https://en.oxforddictionaries.com/spelling/initialisms

My primary purpose here though, was to allow `no_must_violations.yaml` continue to raise no MUST violations. Without "apis" added to the white-list then `PluralizeResourceNamesRule` kicks in and objects to the various resources starting `/apis/`, which the old `isPlural(..)` method was already happy about.

(Aside: Why is "vat" listed as a special case plural here? I can't think of a usage where it's considered plural!?)

maxim-tschumak: > Why is "vat" listed as a special case plural here?

No idea - there was probably a use-case. Let's keep it as is.

>  In fact the plural initialisms here suggest that "APIs" would be preferred: https://en.oxforddictionaries.com/spelling/initialisms

Good to know :)

I'm fine with the current implementation since it's backward compatible (at least for the test cases we have)

tfrauenstein:
👍

maxim-tschumak:
:+1:

tfrauenstein:
👍

PR #721, opened 2018-05-26
Use Jackson's `JsonPointer` to represent violation locations 

roxspring:
- Represent paths and pointers using String rather than URI Fragment representation (Basically stripped "#" prefix from pointers, knowing that `JsonPointer` doesn't use the prefix)
- Stop assuming that first path might be a pointer, only set pointer explicitly
- Switched from `String` to `JsonPointer` throughout
  - Also switched `JsonSchemaValidator` to report `List<Violation>` rather than it's own `ValidationResult`
- Simplified Violation creation: 
  - `context.violation("API Audience must be provided")` creates a `Violation` pointing at the current location
  - `context.violation("Operations should use specific HTTP status codes", response)` creates a `Violation` pointing at `response`, falling back to the current location.


roxspring:
Right. Bored with fighting Codacy. Declaring this ready for review.

maxim-tschumak:
Great! Special thanks for updating the CLI and Web-UI!

maxim-tschumak:
:+1:

netme:
👍

PR #729, opened 2018-06-12
User friendly pointer display

### LINKED ISSUE #727: JsonPointers should be displayed in user friendly format
Currently when a pointer based violation is raised the CLI and Web-UI both display the raw JsonPointer (my fault). Given that API definitions involve a lot of paths with slashes in them, which get encoded, then the pointers tend to be encoded and not intuitive to understand. 

We should decode the pointer components and display with a clearer separator for easier comprehension. For example a violation referring to `/paths/~1things/get/responses/200/content/*~1*` could be displayed more clearly as one of:
- `paths: /things: get: responses: 200: content: */*` 
- `paths > /things > get > responses > 200 > content > */*`

Assuming agreement then is there any strong preference on the display separator? I'm assuming something ASCII to be consistent across CLI and web-ui but I guess something like "»" or even more exotic might be viable??

roxspring:
CLI and Web-UI display JsonPointer decoded and separated by ` > ` instead of `/`

For example the pointer `/paths/~1~01things/get/responses/200/content/*~1*` gets displayed as `paths > /~1things > get > responses > 200 > content > */*`

Closes #727


cli/zally/domain/violation.go
```diff
@@ -9,3 +13,8 @@ type Violation struct {
 	Pointer       string   `json:"pointer"`
 	Paths         []string `json:"paths"`
 }
+
+// ToPointerDisplayString returns the violation's Pointer in user friendly display format
+func (v *Violation) ToPointerDisplayString() string {
+	return strings.Replace(strings.Replace(strings.Replace(strings.TrimPrefix(v.Pointer, "/"), "/", " > ", -1), "~1", "/", -1), "~0", "~", -1)
```
maxim-tschumak: This is hard to read. Is there a way to express it with a regex (e.g. https://golang.org/pkg/regexp/)?

maxim-tschumak:
:+1:

cli/zally/domain/violation.go
roxspring: Fair point. I haven't spotted a way that regex really helps here, but we could split it into a sequence of string replacements:

```go
// ToPointerDisplayString returns the violation's Pointer in user friendly display format
func (v *Violation) ToPointerDisplayString() string {
	display := v.Pointer
	display = strings.TrimPrefix(display, "/")
	display = strings.Replace(display, "/", " > ", -1)
	display = strings.Replace(display, "~1", "/", -1)
	display = strings.Replace(display, "~0", "~", -1)
	return display
}
```

That look better?

ddufourboivin-zalando: Or if not, reformat with named intermediary states. Makes it easier to follow the logic and see intermediary results in trace-debug.

```go
func (v *Violation) ToPointerDisplayString() string {
	noPrefix := strings.TrimPrefix(v.Pointer, "/")
	splitBySlash := strings.Replace(noPrefix, "/", " > ", -1)
	tildeOneToSlash := strings.Replace(splitBySlash, "~1", "/", -1)
	tildeZeroToSlash := strings.Replace(tildeOneToSlash, "~0", "~", -1)
	return tildeZeroToSlash
}
```

ddufourboivin-zalando: Actually, like @roxspring 's version better :P I had a nice Kotlin/Scala immutable reflex approach to it.

roxspring: Yeah - like the immutable approach but also liked the closer alignment that variable reuse allowed. Have gone with my version based on later comment.

maxim-tschumak:
:+1:

netme:
Could you please fix codecov issues?

netme:
@roxspring hard to say without going deep into details , I will try to add probably more tests

maxim-tschumak:
:+1:

roxspring:
@netme codecov issues fixed with extra explicit `violation_test.go`

roxspring:
I'm guessing this is waiting for @ruiaraujo for web-ui changes to be reviewed? Or can @maxim-tschumak find somebody else to review??

(chasing up some aging PRs)

maxim-tschumak:
:+1:

maxim-tschumak:
@roxspring we will find someone :)

tkrop:
@roxspring @maxim-tschumak I must admit I would prefer, paths to stick to standard presentation as JsonPointers, even if the escaping is a bit unusual. Is this escaping an often happening effect?

On the other side, this is nothing I would fight for.

tkrop:
:+1:

roxspring:
@tkrop regarding your question, yes the escaping does end up happening often and I'm finding it a bit of a usability issue with our developers here.

Writing extensive swagger/oas specs does involve using JsonPointer references to shared definitions/schemas/properties/responses etc but conveniently none of these involve any encoding/decoding and so the pointers are fairly intuitive to read and write.

Admittedly I've not run into an API definition which required the `~` to be encoded yet, but the vast majority of violations are in reference to a paths and/or media types which both implicitly involve `/` characters, often several of them in a path. Having developers mentally decode those pointers in order to work out where in their spec they need to update may be a small cognitive load in the grand scheme of things, but it's real and unnecessary IMHO.

tkrop:
@roxspring thanks for pointing out. I didn't had the constant escaping of `#/paths/...` on my radar.

Could be helpful to have tooling support for `yq`/`jq` (would look like `.paths."/~1things".get.responses.200.content."*/*"`).

Issue #550, opened 2017-10-05
Improve UseSpecificHttpStatusCodes Rule


maxim-tschumak:
Check implementation should be aligned more to Guideline Rule [150](http://zalando.github.io/restful-api-guidelines/#150).

**Acceptance Criteria**
* List of standardized HTTP response codes is compiled (research needed)
* Both aspects are still covered:
  1. Response code is a standardized HTTP code
  1. Response code is used in the specified HTTP method

netme:
Is it a duplicate of #630 ?

roxspring:
#630 was tackling a very specific case where I'd noticed a gap. I've not done the research implied here.

roxspring:
It's painful to compare the guideline here with the rules-config.conf since the former describes a mapping from status -> methods but the latter describes method -> statuses. We should converge on a single form for easier comparison.

PR #732, opened 2018-06-19
Improve UseSpecificHttpStatusCodes Rule

roxspring:
- Inverted configuration from `method -> statuses` to `status -> methods` to align with documentation and more easily compare
- Compared configuration with documentation and allowed `POST` to respond `409` to align them
- Renamed `UseSpecificHttpStatusCodes` to `UseStandardHttpStatusCodes` to match the guideline title
- Added and used `Context.validateOperations(..)` convenience method

Relates to #550


maxim-tschumak:
@roxspring Thanks for another contribution. I hope I can review this PR by the end of the week

server/src/main/resources/rules-config.conf
```diff
@@ -35,14 +35,43 @@ CommonFieldTypesRule {
...
+    412: [PUT, DELETE, PATCH]
+    415: [POST, PUT, DELETE, PATCH]
+    423: [PUT, DELETE, PATCH]
+    428: [ALL]
+    429: [ALL]
+
+    // Server Side Error Codes
+    500: [ALL]
+    501: [ALL]
+    503: [ALL]
```
maxim-tschumak: We should make sure that all standard response codes are accepted. Is it a complete list? e.g. `407` is missing ([RFC](https://tools.ietf.org/html/rfc7235#section-3.2))

roxspring: I've explicitly not attempted to research what the method/status combinations should be allowed or changed the implementation as you suggest. All I'm doing here is to bring the Zally implementation into line with what the Zalando Guideline already states (which doesn't mention `407`) and reworking the configuration so that this comparison is much easier to do going forward. 

If somebody is to do the research into the complete list of acceptable HTTP response code and have the rule check for those, then I suggest this ought to be done in https://github.com/zalando/restful-api-guidelines first and then reflected in Zally afterwards.

Similarly having the guideline document a whitelist, but then implement using a blacklist is likely to lead to inconsistency and confusion. Again I'd have thought this should be thrashed out in https://github.com/zalando/restful-api-guidelines and implemented in Zaly afterwards.

Without this change the guideline and rule configuration are expressed differently to such an extent that it's difficult to spot what the impliciations are - IMHO we should be striving for rules to very clearly match the guidelines such that differences can be easily identified and avoided.

maxim-tschumak: The rule states: `[..] must only use standardized HTTP status codes [..]`. And the overview is just a hand-picked set of popular response codes: `Below we listed the most commonly used and best understood [..]`

roxspring: As a "hand-picked set of popular response codes" it's quite extensive and includes some relatively obscure codes (such as 207, 423) so it ends up reading as a very prescriptive list, even if not intended as such 😄 

Medium to long term I really don't object to the suggested redesign of the rule implementation. Personally think it also ought to involve revision of the guideline to avoid the confusion above, and to allow the guideline and rule to be very closely aligned. By opening the PR I didn't think I was signing up to this longer term change.

In the short term, the aim of the PR was to bring Zally into alignment with the explicit stated examples. To me this seems a clearly positive minor change in itself. Blocking the change because it's not implementing the desired medium/long term design seems a little unreasonable.

ddufourboivin-zalando: I am 👍 with splitting this in 2 task (this one and a follow up) if the discussion out-scopes the proposed change.

tkrop: I would stick with the list in the guideline. For other status codes I would like to see a good use case before we add them to both lists: Zally and the guideline.

tkrop:
:+1:

maxim-tschumak:
:-1:

maxim-tschumak:
The rule in the guidelines implies 2 checks:

1. A whitelist of standard HTTP response codes: a MUST check
2. A whitelist of a well-understood set of response codes (list in the guidelines): SHOULD check

Currently, the second rule is a MAY, but @tkrop is going to change this to a SHOULD rule today

maxim-tschumak:
@roxspring @tkrop @ddufourboivin-zalando let's merge this PR and I'll do the research and implement the rest in https://github.com/zalando/zally/issues/752

maxim-tschumak:
:+1:

maxim-tschumak:
:+1:

tkrop:
:+1:

PR #742, opened 2018-07-02
741 fix checks for zalando rule 143

### LINKED ISSUE #741: Checks for Zalando Rule 143 are too Strict
The Rule implementation has to check:
- [x] If first path segment is a parameter (is not stated explicitly in the rule, it's common sense )

maxim-tschumak:
Fixes #741


maxim-tschumak:
@tkrop To summarize this: you propose to forbid two (or more) successive path parameters, right?

BTW, we haven't defined the `:` separator anywhere.

tkrop:
@maxim-tschumak Yes, this interpretation is only guided by the examples.

However thinking about it, this interpretation is still incomplete, as we could even allow ´/resource/{id-part-1}/{id-part-2}´. The important aspect of the rule is (excuse the imprecise math usage behind the i.e.):

∀x∈P ⇔ x + "/{resource-y-id}"∉P where P:={resource-paths} i.e [/prefix]/resource-name/({resource-x-id})+.

With other words there may not exists two paths with different resources, where the resource identifiers are not separated by an additional (sub-)resource name. But checking this will be quite complex, as one has to take into account all other existing paths.

roxspring:
For what it's worth, in-house we use a couple of rules defined in this area:

- PathParamIsWholePathComponent (name still not snappy enough though!)
  - `/component1` 👍 (no path param at all)
  - `/component1/{component2}` 👍
  - `/component1/{component2}/component3` 👍 
  - `/component1/{component2}-with-suffix` ⛔️ (no suffixes)
  - `/component1/with-prefix-{component2}` ⛔️ (no prefixes)
  - `/component1/{component2a}:{component2b}` ⛔️ (no composites)

- PathParamProceededByPlural
  - `/things` 👍 (no path param at all)
  - `/things/{id}` 👍 ("things" is plural)
  - `/things/{id}/others/{otherid}` 👍 ("things" and "others" are plural)
  - `/{id}` ⛔️ (id has no proceeding plural)
  - `/things/{id}/{otherid}` ⛔️ (otherid has no proceeding plural)
  - `/thing/{id}` ⛔️ ("thing" is proceeding but not not plural)

I _think_ these two in combination achieve the result you're after (if you ignore the bonus plurality check). Happy to donate the code if of interest.

maxim-tschumak:
@tkrop @roxspring thanks for your input! I ended up having 3 checks:
- A parameter is not allowed in the first path segment
- Successive parameters are not allowed
- prefixed or suffixed parameters are not allowed (we believe it's a good idea, @roxspring)

Please re-review

tkrop:
:+1:

maxim-tschumak:
:+1:

PR #744, opened 2018-07-02
Remove generic type in `ReverseAst` (#743)

### LINKED ISSUE #743: ReverseAst type parameter is useless
Class `ReverseAst<T>` never uses the `T` type parameter ever. It should be removed.

ddufourboivin-zalando:
fix #743


tkrop:
:+1:

PR #748, opened 2018-07-05
PR-560: Extract rule API into subproject.

beatkyo:
I tried to move on #560. 

I moved Rules API to a gradle submodule `zally-rules`.

I also managed to remove Spring dependency from `zally-rules`. Using `BeanDefinitionRegistryPostProcessor` we can register BeanDefinition at runtime prior to any Bean instantiation. This may contribute to #122 .

From what I've seen, to be fully indenpendant from Spring, we would need a way to pass the `Config` object without `@Autowired` to rule creation. I probably missed other dependencies.

Also I have no prior experience with gradle so I probably overlooked the build.


maxim-tschumak:
I believe your a trying to address #560 (part of Rule Set epic). The idea behind it is to separate the rule implementations from the `server` component. Please have a look into the issue #560.

beatkyo:
Oops my bad, I meant #560 not 551. I Fixed the PR

maxim-tschumak:
Some comments from my side:
* Let's rename the subpackage to `zally-rule-api` or `zally-rule-interface`
* We should also include `Context` interface

roxspring:
This seems like a healthy step forward to me!

FWIW the `zally-rule-api` gets my vote for the subproject name. Short and sweet, and doesn't make me stop and think about the interface concept vs keyword. (Although I guess the argument against is that it might make some people think about JVM API vs REST API).

Clearly a `Context` interface should be extracted into the subproject too since most/all rules shouldn't be implemented without reference to a `Context` going forward, but the current `Context` class has lots of dependencies into murky depths of `.util.ast` so there's some extraction work to do (perhaps a `Context` interface and `ReverseAstContext` implementation???). Not sure if it's worth blocking the PR for.

beatkyo:
Without `Context` the module cannot be used independently of the server code. 

This raise an issue to me, the `Context` dependency is not clearly expressed with the `Rule` and `Check` annotations. I get it that reflection and `Check` are used to execute the rules. But this leaves just annotations with no implementation contract.

I don't really know were to go from this though.

roxspring:
> Without `Context` the module cannot be used independently of the server code.

That's what I was trying to get at above. As things stand the `Context` is very much part of the implementation, but it's interface is key to writing rules (which take a `Context` argument). I had a go at the extraction and it wasn't as painful as I feared: https://github.com/roxspring/zally/commit/7327ae00659ba01c6c26ff89279b2f1b4a335cd3. I'm not totally convinced by the `ReverseAstContext` name but I don't have a better idea right now - lazy options like `DefaultContext` or `ContextImpl` don't seem very zallyish.

> This raise an issue to me, the `Context` dependency is not clearly expressed with the `@Rule` and `@Check` annotations. I get it that reflection and `@Check` are used to execute the rules. But this leaves just annotations with no implementation contract.

Yes, this is a downside of the annotation approach. Previously there was a simple `Rule` interface with a clear contract defined by the `validate(Swagger): Violation?` method, but we wanted to move to a position where we have multiple validations (i.e. `@Check`s) within a single rule implementation. Annotations seemed the sensible way to achieve that. Note also that the `Context` class is still very young, most rules aren't even using it... yet... the transition is moving fast!

Neither this history nor current rule writing recommendations are clearly documented though. Guess we ought to track an issue for that.

> I don't really know were to go from this though.

- I think it'd be good to rename the subproject: `zally-rule-api` establishes a convention of subprojects matching the sub-package naming.
- If you want to pull my commit (modified or not) into your PR then I'm fine with that.
- If @maxim-tschumak is happy to take this PR as is, then I'm also happy to open a separate PR for the `Context` work.

maxim-tschumak:
I had exactly the same idea in mind with the extracting the interface (@roxspring's branch). I believe this is the right way to go.

> I'm not totally convinced by the ReverseAstContext name

Please don't re-use the AST abbreviation - it is wrongly used in this context (we have to improve this eventually). `DefaultContext` or `ContextImpl` are perfectly fine names for the context implementation!

> If @maxim-tschumak is happy to take this PR as is, then I'm also happy to open a separate PR for the Context work.

great! I'm happy :)

> Oh, and given that the subproject is pure kotlin, perhaps we should move the source to src/main/kotlin/ rather than src/main/java/ - again happy to do it in a separate PR if preferred.

+1

beatkyo:
@roxspring I picked your commit and went for DefaultContext instead of ReverseAstContext

Used rebase & push force to rename module to `zally-rule-api` and source directory to `kotlin`. Also rebase on `master`.

maxim-tschumak:
LGTM!

minor comment: please add \n on the end of the following files:
- server/settings.gradle
- server/zally-rule-api/settings.gradle
- server/zally-rule-api/src/main/kotlin/de/zalando/zally/rule/api/Context.kt

maxim-tschumak:
:+1:

maxim-tschumak:
:+1:

roxspring:
@maxim-tschumak, @netme, others, any objections to this? I'm seeing it as a really good step forward towards modularization so would really like to see this merged but getting 2nd approvals seems difficult at the moment (holiday season??).

@beatkyo assuming no objections from above, are you able to resolve the conflict?

ddufourboivin-zalando:
I like the idea. It's just gonna be interesting to sync (resolve conflicts) since we are actively working on the rules @maxim-tschumak and I at the moment.

Let us know when the conflicts are resolved. We'll try to merge it before new ones appear.

beatkyo:
rebased done

maxim-tschumak:
:+1:

ddufourboivin-zalando:
Sorry @beatkyo , it got out of sync again. @maxim-tschumak and I will try NOT merging today in this repo so you can fix it and we merge it. Sorry about this.

ddufourboivin-zalando:
👍

maxim-tschumak:
:+1:

ddufourboivin-zalando:
Finally, we did it ;) Let's back-merge in our branches now! 😱 ;)

Issue #714, opened 2018-05-23
Refactor Checks to Operate on the Context Object


maxim-tschumak:
After implementing the initial support for OpenAPI3 we have to adjust the check implementations. Instead of operating on the Swagger object, the checks have to use the `Context` object. Doing this, the checks become universal - applicable against Swagger2 and OpenAPI3 specifications.

**Acceptance Criteria:**
1. The following checks are refactored to use the `Context` object
2. Minimal specifications are used in the tests

#### Done
- [x] JsonProblemAsDefaultResponseRule
- [x] UseSpecificHttpStatusCodes
- [x] AvoidTrailingSlashesRule
- [x] CommonFieldTypesRule
- [x] EverySecondPathLevelParameterRule
- [x] MediaTypesRule 🎋 [Branch](https://github.com/zalando/zally/tree/714-rules-to-oas--MediaTypesRule)  🔍 [Pull Request](https://github.com/zalando/zally/pull/740)
- [x] AvoidLinkHeadersRule 🎋 [Branch](https://github.com/zalando/zally/tree/714-rules-to-aos--AvoidLinkHeadersRule) 🔍 [Pull Request](https://github.com/zalando/zally/pull/763)
- [x] PluralizeResourceNamesRule

#### High priority list (please migrate these first)
- [x] FormatForNumbersRule
- [x] LimitNumberOfResourcesRule ([PR](https://github.com/zalando/zally/pull/801))
- [x] PluralizeNamesForArraysRule ([PR](https://github.com/zalando/zally/pull/802))
- [x] QueryParameterCollectionFormatRule ([PR](https://github.com/zalando/zally/pull/805))
- [x] SecureWithOAuth2Rule ([PR](https://github.com/zalando/zally/pull/806))
- [x] SuccessResponseAsJsonObjectRule ([PR](https://github.com/zalando/zally/pull/807))
- [x] ExtensibleEnumRule ([PR](https://github.com/zalando/zally/pull/810))

#### Lower priority list
- [x] Use429HeaderForRateLimitRule ([PR](https://github.com/zalando/zally/pull/812))
- [x] SnakeCaseInPropNameRule ([PR](https://github.com/zalando/zally/pull/815))
- [x] SnakeCaseForQueryParamsRule ([PR](https://github.com/zalando/zally/pull/816))
- [x] ApiMetainformationRule ([PR](https://github.com/zalando/zally/pull/819))
[...]

PR #750, opened 2018-07-10
fix: common fields rule (#747)

### LINKED ISSUE #747: Check for Zalando rule #174 false-positive
Zalando rule 174 checks for proper usage of common fields. The Guideline states the field `type` must have the type `string`.

The rule check doesn't check the type of referenced definition and generates a violation if the type is not set explicitly.

**Example / How to reproduce** The Zally Server API specification:
The `Rule` definition contains a `type` field which references to the definition of the `RuleType`. The type of the `RuleType` is `x-extensible-enum`, which is also a string.
So technically, the type of the `type` field of the `Rule` definition is `string` and the check should not produce any violations

maxim-tschumak:
Related to #714 
Fixes #747


server/src/test/java/de/zalando/zally/rule/zalando/CommonFieldTypesRuleTest.kt
```diff
@@ -1,93 +1,206 @@
...
-        assertThat(rule.checkField("id", StringProperty())).isNull()
-        assertThat(rule.checkField("id", StringProperty("UUID"))).isNull()
-        assertThat(rule.checkField("created", StringProperty("date-time"))).isNull()
-        assertThat(rule.checkField("modified", StringProperty("date-time"))).isNull()
-        assertThat(rule.checkField("type", StringProperty())).isNull()
+    fun `checkField matching should be case insensitive`() {
+        assertThat(rule.checkField("iD", createSchema(INTEGER_TYPE, null))).isNotNull()
+        assertThat(rule.checkField("CREATED", createSchema(INTEGER_TYPE, null))).isNotNull()
+        assertThat(rule.checkField("tYpE", createSchema(null, null))).isNotNull()
+        assertThat(rule.checkField("CREated", createSchema(STRING_TYPE, "time"))).isNotNull()
```
tkrop: By the way, we should introduce some synonyms rule: `created` -> `created_at`, `modified` -> `modified_at`.

server/src/test/java/de/zalando/zally/rule/zalando/CommonFieldTypesRuleTest.kt
```diff
@@ -1,93 +1,206 @@
...
+              /pets:
+                get:
+                  responses:
+                    200:
+                      content:
+                        application/json:
+                          schema:
+                            properties:
+                              id:
+                                "${'$'}ref": "#/components/schemas/CustomId"
```
tkrop: Unfortunately, this is not the only allowed place for a reference.

maxim-tschumak: It's a representative test - we don't want to test the resolve feature of the `swagger-parser`.

maxim-tschumak:
Now, the rule also checks all the nested properties recursively. @tkrop please have another look

maxim-tschumak:
:+1:

PR #751, opened 2018-07-11
feat: functional naming schema checks for hostnames (#749)

### LINKED ISSUE #749: Implement Checks for Functional Naming (Zalando Rules 223, 224, 213)

maxim-tschumak:
Closes #749

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRule.kt
@@ -24,7 +24,7 @@ class FunctionalNamingForHostnamesRule {
     private val functionalDomain = """[a-z][a-z0-9]*"""
     private val functionalComponent = """[a-z][a-z0-9-]*"""
     private val functionHostnameURLRegEx =
-            """(https://)?$functionalDomain-$functionalComponent.zalandoapis.com.*""".toRegex()
+            """(https://)?$functionalDomain-$functionalComponent\.zalandoapis\.com.*""".toRegex()
 
     @Check(severity = Severity.MUST)
     val mustFollowFunctionalNaming = checkHostnames(mustFollow)
--- a/server/src/test/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRuleTest.kt
+++ b/server/src/test/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRuleTest.kt
@@ -62,7 +62,7 @@ class FunctionalNamingForHostnamesRuleTest {
     }
 
     @Test
-    fun `(must|should|may)FollowFunctionalNaming should not return any violations if audience is not set`() {
+    fun `(must|should|may)FollowFunctionalNaming should return no violations if audience is not set`() {
         val context = Context.createOpenApiContext("openapi: 3.0.1")!!
 
         assertThat(rule.mustFollowFunctionalNaming(context)).isEmpty()


server/src/main/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRule.kt
```diff
@@ -0,0 +1,51 @@
...
+    private val description = "hostname has to follow the functional naming schema"
+
+    private val mustFollow = listOf("external-public", "external-partner")
+    private val shouldFollow = listOf("company-internal", "business-unit-internal")
+    private val mayFollow = listOf("component-internal")
+
+    private val functionalDomain = """[a-z][a-z0-9]*"""
+    private val functionalComponent = """[a-z][a-z0-9-]*"""
+    private val functionHostnameURLRegEx =
+            """(https://)?$functionalDomain-$functionalComponent.zalandoapis.com.*""".toRegex()
```
roxspring: Looks like this will currently match `domain-component_zalandoapis_com.evil`.

Suspect it should end more like `\.zalandoapis\.com($|/.*)` to check for literal dots and a definite end to the fqdn.

server/src/test/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRuleTest.kt
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Test
+    fun `shouldFollowFunctionalNaming should return violation for company-internal APIs with invalid hostname`() {
+        val context = withAudienceAndHostname("company-internal", "incorrect url")
+
+        val violations = rule.shouldFollowFunctionalNaming(context)
+
+        assertThat(violations).isNotEmpty
+        assertThat(violations[0].description).containsPattern(".*follow the functional naming schema.*")
```
tkrop: missing `should` condition?

server/src/test/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRuleTest.kt
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Test
+    fun `mustFollowFunctionalNaming should return violation for external-public APIs with invalid hostname`() {
+        val context = withAudienceAndHostname("external-public", "incorrect url")
+
+        val violations = rule.mustFollowFunctionalNaming(context)
+
+        assertThat(violations).isNotEmpty
+        assertThat(violations[0].description).containsPattern(".*follow the functional naming schema.*")
```
tkrop: missing `must` condition?

server/src/test/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRuleTest.kt
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Test
+    fun `mayFollowFunctionalNaming should return violation for component-internal APIs with invalid hostname`() {
+        val context = withAudienceAndHostname("component-internal", "incorrect url")
+
+        val violations = rule.mayFollowFunctionalNaming(context)
+
+        assertThat(violations).isNotEmpty
+        assertThat(violations[0].description).containsPattern(".*follow the functional naming schema.*")
```
tkrop: missing `may` condition?

server/src/test/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRuleTest.kt
```diff
@@ -0,0 +1,84 @@
...
+    @Test
+    fun `(must|should|may)FollowFunctionalNaming should not return any violations if audience is not set`() {
+        val context = Context.createOpenApiContext("openapi: 3.0.1")!!
+
+        assertThat(rule.mustFollowFunctionalNaming(context)).isEmpty()
+        assertThat(rule.shouldFollowFunctionalNaming(context)).isEmpty()
+        assertThat(rule.mayFollowFunctionalNaming(context)).isEmpty()
+    }
+
+    private fun withAudienceAndHostname(audience: String, url: String): Context {
```
tkrop: Test what happens if audience is not in list? Do we have an audience check?

server/src/test/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRuleTest.kt
```diff
@@ -0,0 +1,84 @@
...
+        assertThat(rule.isUrlValid("")).isFalse()
+        assertThat(rule.isUrlValid("random text")).isFalse()
+        assertThat(rule.isUrlValid("maxim@some.host")).isFalse()
+        assertThat(rule.isUrlValid("infrastructure.zalandoapis.com")).isFalse()
+        assertThat(rule.isUrlValid("special+characters.zalandoapis.com")).isFalse()
+        assertThat(rule.isUrlValid("http://infrastructure-api-linter.zalandoapis.com")).isFalse()
+        assertThat(rule.isUrlValid("https://infrastructure-api-linter.othercompanysapis.com")).isFalse()
+    }
+
+    @Test
```
tkrop: I would take a parameterized test here containing all expected *must*, *should*, and *may* violation audiences.

server/src/main/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRule.kt
maxim-tschumak: Good catch! Thanks. I add escaping for the 2 dots, but will not change the part which comes after `zalandoapis.com` - I want to be less restrictive here and just check the host part of the URL.

server/src/test/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRuleTest.kt
maxim-tschumak: Already tested, see test `(must|should|may)FollowFunctionalNaming should not return any violations if audience is not set`

tkrop:
:+1:

server/src/test/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRuleTest.kt
maxim-tschumak: The violation description is static, the severity is handled over the annotation - no need to test it explicitly

maxim-tschumak:
:+1:

Issue #167, opened 2017-03-14
Better Code Coverage Tool for Kotlin code


fmueller:
Our current code coverage solution only measures Java code. Since we decided to migrate more and more code to Kotlin, we need to find a way that this code is also measured for coverage by tests.

netme:
After Kotlin meetup did some investigations:

- We are generating code coverage reports using JaCoCo tool:
   ```bash
   gradle clean build
   open build/reports/jacoco/test/html/index.html 
   ```
- Codacy doesn't accept that part of the report since the file extensions are different than `java`
- We can use codecov.io for code coverage, there is a Kotlin example codecov/example-kotlin

netme:
We can also try https://codebeat.co/

fmueller:
Do these other platforms support multi-language setup with Java/Kotlin? So, can we compare Code Coverage report of Java and Kotlin files?

netme:
I can create an example PR and check how it works.

netme:
Asked Codacy guys via their support channel. Kotlin is not on their short roadmap, but we can build our own Docker image to support it. Here is an example: https://github.com/codacy/codacy-coffeelint 

The question is if it worth it or not.

maxim-tschumak:
I don't think it has very high priority for us. Closing the issue...

felix-mueller-sm:
Just as an information: Codacy does support Kotlin now. :)

maxim-tschumak:
Thanks for the info @fmueller! re-opening the issue.

netme:
Can confirm, got a mail from Codacy yesterday 🎉

netme:
Unfortunately Codacy does not support JaCoCo test coverage reports for Kotlin, already reached their team, they are currently working on that. Moving the ticket back to TODO meanwhile.

vadeg:
Codecov supports test coverage for Kotlin. Fixed in #1223

PR #757, opened 2018-07-13
[WIP] Codacy / Detekt Configuration

roxspring:


roxspring:
Grrr. I hate Codacy. Now I can't generate errors in order to test configuration fixes. Arrggghhh!!!!!

maxim-tschumak:
I've deactivated codecov webhook - it was annoying. Should I re-activate it so that you can test the custom config?

roxspring:
LOL clearly we both hit breaking point around the same time and reached for different solutions!

I'm not sure what the right answer is really - on one hand, static analysis can help identify some stupid mistakes and keeps us honest around coding standards - on the other hand, it's very annoying to have all these nitpicky complaints only pointed out at CI time after a PR has been prepared.

I'm tempted to try running Detekt direct from the gradle build so that complaints can be picked up and addressed earlier - but I don't know how much this will slow the build down.

maxim-tschumak:
Here is the issue regarding improving the code coverage tooling: #167
I like the idea of integrating a code quality tool into the build process - let's give it a try!

PR #763, opened 2018-07-24
714-rules-to-aos--AvoidLinkHeadersRule

ddufourboivin-zalando:
#714 Convert rule AvoidLinkHeadersRule to the next _Context_ object (_OpenAPI 3_).


server/src/test/java/de/zalando/zally/rule/zalando/AvoidLinkHeadersRuleTest.kt
```diff
@@ -11,21 +13,30 @@ class AvoidLinkHeadersRuleTest {
 
     @Test
     fun positiveCaseSpp() {
-        val swagger = getFixture("api_spp.json")
-        assertThat(rule.validate(swagger)).isNull()
+        val context = getContextFromFixture("api_spp.json")
```
maxim-tschumak: We want to use comprehensible minimal specification fixtures in tests instead of full-size API specifications.

server/src/test/java/de/zalando/zally/rule/zalando/MediaTypesRuleTest.kt
```diff
@@ -120,7 +120,7 @@ class MediaTypesRuleTest {
 
     @Test
     fun `the SPP API generates violations`() {
```
maxim-tschumak: I would replace tests based on a complete API specification with test cases describing the behavior of the check functions.

server/src/main/java/de/zalando/zally/rule/HttpHeadersRuleWithContext.kt
```diff
@@ -0,0 +1,51 @@
+package de.zalando.zally.rule
+
+import com.typesafe.config.Config
+import de.zalando.zally.rule.api.Violation
+import io.swagger.v3.oas.models.parameters.Parameter
+import io.swagger.v3.oas.models.responses.ApiResponse
+
+// todo: When all rules deriving from `HttpHeadersRule` are converted to the `Context` object, replace `HttpHeadersRule` with this class (#714).
+abstract class HttpHeadersRuleWithContext(rulesConfig: Config) {
```
maxim-tschumak: Just a comment: In general, I would avoid having a class hierarchy for rules. I would rather move these util functions to some Utility class or to the context's companion object.

ddufourboivin-zalando: I did not want to change what was in place more than necessary. This is simply a copy of the original base class.

I am also OK in changing that. This header-gathering function can be moved to either the context or simply some util object.

maxim-tschumak: > I did not want to change what was in place more than necessary. This is simply a copy of the original base class.

I know - that's why it's only a comment :)

> I am also OK in changing that. This header-gathering function can be moved to either the context or simply some util object.

We don't have to do this in this PR.

server/src/test/java/de/zalando/zally/rule/zalando/AvoidLinkHeadersRuleTest.kt
ddufourboivin-zalando: For the negative cases, I do agree; but for the positive ones too?

server/src/test/java/de/zalando/zally/rule/zalando/MediaTypesRuleTest.kt
ddufourboivin-zalando: This has not changed since the PR for this rule. Only the `!!` have been removed because of a change in `getContextFromFixture`.

server/src/test/java/de/zalando/zally/rule/zalando/AvoidLinkHeadersRuleTest.kt
ddufourboivin-zalando: Nevermind that. We said we wanted to get rid of all of most of them because of internal reasons.

server/src/test/java/de/zalando/zally/rule/zalando/AvoidLinkHeadersRuleTest.kt
```diff
@@ -1,31 +1,143 @@
...
     private val rule = AvoidLinkHeadersRule(testConfig)
 
     @Test
-    fun positiveCaseSpp() {
-        val swagger = getFixture("api_spp.json")
-        assertThat(rule.validate(swagger)).isNull()
+    fun `a Swagger API with no header called Link produces no violation`() {
+        @Language("YAML")
+        val context = Context.createSwaggerContext("""
+            swagger: 2.0
```
maxim-tschumak: You can have an even smaller example, like:

```
swagger: 2.0
paths:
  /foo:
    get:
      parameters:
        FlowId:
          name: X-Flow-Id
          in: header
          type: string
```

BTW, the `parameters` are wrongly intended in your example (should be under `get:`).

maxim-tschumak:
:+1:

server/src/test/java/de/zalando/zally/rule/zalando/AvoidLinkHeadersRuleTest.kt
ddufourboivin-zalando: Those parameters are for all of the API, not just for that path/verb.

ddufourboivin-zalando: I wanted the example to include:
- a header in a path/verb
- a header in the global parameters
- another parameter that is not a header

to make sure the test does not return false negatives.

I can remove one of the header parameter for shortening it.

I can also add one that is called `Link` but is not a header, again to shield against false negatives in the test.

maxim-tschumak:
:+1:

ddufourboivin-zalando:
👍

PR #771, opened 2018-07-30
BUGFIX: Path with trailing slash respond 403

### LINKED ISSUE #770: BUG: Requests with trailing slashes respond 403
How to reproduce:

```
curl -XPOST -v -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" --data "@test.yaml" "$ZALLY_SERVER/api-violations" | jq
```
yields
```json
{
  "title": "Forbidden",
  "status": 403,
  "detail": "Access is denied"
}
```

While without a trailing slash
```
curl -XPOST -v -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" --data "@test.yaml" "$ZALLY_SERVER/api-violations" | jq
```
authentifies correctly.

ddufourboivin-zalando:
closes #770 

Consider that there is no test on the OAUTH configuration at all. Please review carefully, since I am no Spring Security expert.

### DIFF
--- a/server/src/main/java/de/zalando/zally/configuration/OAuthConfiguration.java
+++ b/server/src/main/java/de/zalando/zally/configuration/OAuthConfiguration.java
@@ -45,11 +45,13 @@ class OAuthConfiguration extends ResourceServerConfigurerAdapter {
             .and()
             .authorizeRequests()
             .antMatchers(HttpMethod.OPTIONS, "/**").permitAll()
-            .antMatchers("/health").permitAll()
-            .antMatchers("/metrics",
-                "/api-violations",
-                "/supported-rules",
-                "/review-statistics").access("#oauth2.hasScope('uid')")
+            .regexMatchers("/health/?").permitAll()
+            .regexMatchers(
+                "/metrics/?",
+                "/api-violations/?",
+                "/supported-rules/?",
+                "/review-statistics/?")
+            .access("#oauth2.hasScope('uid')")
             .antMatchers("**").denyAll();
 
         http
--- a/server/src/test/java/de/zalando/zally/configuration/PathMatchersTest.kt
+++ b/server/src/test/java/de/zalando/zally/configuration/PathMatchersTest.kt
@@ -1,0 +1,55 @@
+package de.zalando.zally.configuration
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Ignore
+import org.junit.Test
+import org.mockito.Mockito
+import org.springframework.security.web.servlet.util.matcher.MvcRequestMatcher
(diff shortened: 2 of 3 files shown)


server/src/main/java/de/zalando/zally/configuration/OAuthConfiguration.java
```diff
@@ -45,11 +45,13 @@ public void configure(final HttpSecurity http) throws Exception {
             .and()
             .authorizeRequests()
             .antMatchers(HttpMethod.OPTIONS, "/**").permitAll()
```
maxim-tschumak: let's migrate this one also to `regexMatcher`

server/src/test/java/de/zalando/zally/configuration/PathMatchersTest.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.security.web.util.matcher.RegexRequestMatcher
+import javax.servlet.http.HttpServletRequest
+
+/**
+ * This serves as a reminder/documentation of how the matchers do work.
+ */
+class PathMatchersTest {
+
+    @Test
+    fun `antMatcher without trailing slash matches just requests without trailing slash`() {
```
maxim-tschumak: Why do we need to test the `antMatcher`? (since we migrated all the filters to `regexMatcher`)

tkrop: We should test the effectiveness of the configuration, not the working of configuration patterns :(.

maxim-tschumak: In general, I agree with @tkrop. But I'm also OK with testing the effectiveness of the configuration since we had a problem with this critical part of the application and want to become more confident about the configuration.

ddufourboivin-zalando: This "test" is there to show / explain / remind ourselves on what does what in terms of matchers and to make sure our assumptions are proven true.

As for testing the configuration, I welcome any idea or experience you might have in testing such a thing. In the current state of the application, I do not know how it is possible.

server/src/main/java/de/zalando/zally/configuration/OAuthConfiguration.java
ddufourboivin-zalando: I expect `regexMatchers` to be more runtime-costy than `antMatchers`. Since this one did not need a `regex` logic to work property, I voluntarily left it as is.

maxim-tschumak:
:+1:

ddufourboivin-zalando:
👍

PR #772, opened 2018-07-31
fix(server): accept application/problem+json as content type (#768)

### LINKED ISSUE #768: False-positive results of the ProblemJSON rule
The Rule "[MUST use Problem JSON](https://zalando.github.io/restful-api-guidelines/#176)" produces false-positive violations for API specifications which produce `application/problem+json` in responses.

Violation description:
```
Operations Should Return Problem JSON When Any Problem Occurs During Processing Whether Caused by Client Or Server. Media type must be application/json.
```

The `application/problem+json` content type should be accepted.

maxim-tschumak:
Fixes #768


ddufourboivin-zalando:
:

server/src/main/java/de/zalando/zally/rule/zalando/UseProblemJsonRule.kt
```diff
@@ -24,6 +24,7 @@ import io.swagger.v3.oas.models.responses.ApiResponse
 class UseProblemJsonRule {
     private val description = "Operations Should Return Problem JSON When Any Problem Occurs During Processing " +
```
tkrop: Is this a description sentence or a title? Why do we write every word with capital letters?

server/src/test/java/de/zalando/zally/rule/zalando/UseProblemJsonRuleTest.kt
```diff
@@ -55,6 +55,37 @@ class UseProblemJsonRuleTest {
         )
     }
```
tkrop: Can you please separate the before test into two tests?

server/src/main/java/de/zalando/zally/rule/zalando/UseProblemJsonRule.kt
```diff
@@ -24,6 +24,7 @@ import io.swagger.v3.oas.models.responses.ApiResponse
...
         "Whether Caused by Client Or Server."
+    private val allowedContentTypes = listOf("application/json", "application/problem+json")
```
tkrop: I see the point why `application/json` covers `application/problem+json`, but I think our rule clearly states to only use `application/problem+json`. Is there another problem for OpenAPI 2.0 specs, that prevents this matching? In this case I would suggest to address this separately for this case only.

ddufourboivin-zalando: @tkrop : I do not understand your comment. There does not seem to be here a OAS 3 specific logic.

server/src/main/java/de/zalando/zally/rule/zalando/UseProblemJsonRule.kt
ddufourboivin-zalando: I concur. This should be a normal sentence.

server/src/main/java/de/zalando/zally/rule/zalando/UseProblemJsonRule.kt
maxim-tschumak: @tkrop @ddufourboivin-zalando The guideline doesn't say explicitly that the API designers have to use `application/problem+json`. It only says that the standard also defines it.

Technically, using the `application/json` content type is also correct. I would not disallow the usage of `application/json`.

maxim-tschumak:
:+1:

server/src/test/resources/fixtures/no_must_violations.yaml
```diff
@@ -100,6 +101,9 @@ paths:
         - uid
       parameters:
       - $ref: "#/parameters/LifecycleState"
+      produces:
```
tkrop: Is noise.

PR #776, opened 2018-08-01
Parse Error Causes 500 SERVER ERROR

### LINKED ISSUE #773: Parse and Conversion Problems cause 500
# How to reproduce?

Access [Zally web UI](https://zally.zalando.net/).

Lint this:
```yaml
swagger: "2.0"
```

Result:
> ERROR
> Ooops something went wrong!

# Expected Result

We should get a violation of [rule 101](https://opensource.zalando.com/restful-api-guidelines/#101) _Must: Provide API Specification using OpenAPI_, since the specification is not valid _Open API_.

# Root Cause

This specification makes the converter (from Swagger 2 to OpenAPI 3) throw an exception. Instead of crashing, _Zally_ should detect that an form a proper violation instead of crashing with a 500.

ddufourboivin-zalando:
fixes #773


server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
```diff
@@ -123,45 +123,53 @@ class DefaultContext(openApi: OpenAPI, swagger: Swagger? = null) : Context {
...
                     val sep = "\n  - "
                     val messageBulletList = parseResult.messages.joinToString(sep)
                     throw RuntimeException("Swagger parsing failed with those errors:$sep$messageBulletList")
                 }
+                parseResult?.openAPI?.let {
+                    ResolverFully(true).resolveFully(it) // workaround for NPE bug in swagger-parser
+                    DefaultContext(it)
+                }
+            } catch (t: Throwable) {
+                null
```
maxim-tschumak: Let's log the exception message, at least in the debug mode

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
```diff
@@ -123,45 +123,53 @@ class DefaultContext(openApi: OpenAPI, swagger: Swagger? = null) : Context {
...
                         } catch (e: NullPointerException) {
                             log.warn("Failed to fully resolve Swagger schema.", e)
-                        if (failOnParseErrors) throw e }
+                            if (failOnParseErrors) throw e
+                        }
                         DefaultContext(it, swagger)
                     }
                 }
+            } catch (t: Throwable) {
+                null
```
maxim-tschumak: same here

web-ui/src/server/index.js
```diff
@@ -11,7 +11,7 @@ const DEFAULT_OPTIONS = {
   windowEnv: {
     OAUTH_ENABLED: false,
     DEBUG: true,
-    ZALLY_API_URL: 'http://localhost:8000',
+    ZALLY_API_URL: 'http://localhost:8080',
```
maxim-tschumak: Please double check if `docker-compose` still works.

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
tkrop: I have the feeling this is the wrong place to catch this exception.

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
tkrop: same here.

ddufourboivin-zalando:
Dropping this PR in favor of a cleaner solution. Stay tuned.

PR #777, opened 2018-08-02
feat(server): check for YAML format (#755)

### LINKED ISSUE #755: Check for Zalando Rule 101: must use YAML format
Guideline rule: https://opensource.zalando.com/restful-api-guidelines/#101

**Acceptance criteria**
- [x] Check returns a `MUST` violation if the API specification is provided in YAML format

maxim-tschumak:
Closes #755

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
+++ b/server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
@@ -137,7 +137,7 @@ class DefaultContext(override val source: String, openApi: OpenAPI, swagger: Swa
             }
             return parseResult?.openAPI?.let {
                 ResolverFully(true).resolveFully(it) // workaround for NPE bug in swagger-parser
-                DefaultContext(content, it, null)
+                DefaultContext(content, it)
             }
         }
 
--- a/server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
+++ b/server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
@@ -18,7 +18,7 @@ import io.swagger.v3.parser.core.models.ParseOptions
 import io.swagger.v3.parser.util.ResolverFully
 import org.slf4j.LoggerFactory
 
-class DefaultContext(openApi: OpenAPI, swagger: Swagger? = null) : Context {
+class DefaultContext(override val source: String, openApi: OpenAPI, swagger: Swagger? = null) : Context {
     private val recorder = MethodCallRecorder(openApi).skipMethods(*extensionNames)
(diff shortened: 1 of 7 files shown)


server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
```diff
@@ -18,7 +18,7 @@ import io.swagger.v3.parser.core.models.ParseOptions
 import io.swagger.v3.parser.util.ResolverFully
 import org.slf4j.LoggerFactory
 
-class DefaultContext(openApi: OpenAPI, swagger: Swagger? = null) : Context {
+class DefaultContext(override val source: String, openApi: OpenAPI, swagger: Swagger? = null) : Context {
```
tkrop: Please explain: what is the `override` for?

You can may ` = null`. By the way is this construction public or private?

server/src/main/java/de/zalando/zally/rule/zalando/UseOpenApiRule.kt
```diff
@@ -57,6 +58,19 @@ open class UseOpenApiRule(@Autowired rulesConfig: Config) {
         return violations
     }
 
+    @Check(severity = Severity.MUST)
+    fun checkIfTheFormatIsYAML(context: Context): Violation? {
+        // at this point the api specification has been already parsed successfully
+        // -> the source is either a valid YAML or JSON format
+        // -> JSON must start with '{' and end with '}'
+        val cleanedUpSource = context.source.trim()
```
tkrop: Trim is only omitting whitespace characters. Shouldn't we although trim line breaks and other non-printable characters? How about using `trimIdent()`?

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
maxim-tschumak: > Please explain: what is the override for?

`override` keyword indicates that the member of the `Context` interface is overridden in this class

> You can may = null.

This is not possible - `source` is enforced to be not `null`.

> By the way is this construction public or private?

It's public.

server/src/main/java/de/zalando/zally/rule/zalando/UseOpenApiRule.kt
maxim-tschumak: There is no need in using `trimIdent()` - removing leading and trailing whitespaces is enough since we only check the beginning and the end of the string. It's also faster.

tkrop:
:+1:

maxim-tschumak:
:+1:

PR #779, opened 2018-08-08
fix(server): Short circuit recursive definitions to avoid infinite recursion

### LINKED ISSUE #778: `OutOfMemoryError: Java heap space` processing recusive swagger definition
The following minimal swagger yaml causes an `OutOfMemoryError: Java heap space`. It's a slightly odd case derived from a real definition we have: a tree node structure is redefined in order to add additional fields to each node (the lesser WriteNode is used when POSTing the object, the ReadNode version is augmented with additional read-only fields).

```yaml
swagger: '2.0'
info:
    title: Tree API
    version: 1.0.0
paths:
    '/tree':
    get:
        responses:
        200:
            description: List of nodes.
            schema:
            $ref: '#/definitions/ReadNode'
definitions:
    WriteNode:
    type: object
    properties:
        name:
        type: string
        children:
        type: array
        items:
            $ref: '#/definitions/WriteNode'
    ReadNode:
    allOf:
        - $ref: '#/definitions/WriteNode'
        - type: object
        properties:
            extra: # property that WriteNode doesn't have
            type: string
            children: # children redefined to be ReadNode rather than WriteNode
            type: array
            items:
                $ref: '#/definitions/ReadNode'
```

The `ReverseAstBuilder` ends up recursing indefinitely through:

- /components/schemas/WriteNode/properties/children/items
- /components/schemas/WriteNode/properties/children/items/properties
- /components/schemas/WriteNode/properties/children/items/properties/children
- /components/schemas/WriteNode/properties/children/items/properties/children/items
- /components/schemas/WriteNode/properties/children/items/properties/children/items/properties
- /components/schemas/WriteNode/properties/children/items/properties/children/items/properties/children

roxspring:
Closes #778

### DIFF
--- a/server/src/main/java/de/zalando/zally/util/ast/ReverseAstBuilder.java
+++ b/server/src/main/java/de/zalando/zally/util/ast/ReverseAstBuilder.java
@@ -66,7 +66,7 @@ public class ReverseAstBuilder<T> {
                 continue;
             }
             if (!PRIMITIVES.contains(node.object.getClass())) {
-                List<Node> children;
+                Collection<Node> children;
                 if (node.object instanceof Map) {
                     children = handleMap((Map<?, ?>) node.object, node.pointer, node.marker);
                 } else if (node.object instanceof List) {
@@ -78,8 +78,8 @@ public class ReverseAstBuilder<T> {
                 } else {
                     children = handleObject(node.object, node.pointer, node.marker);
                 }
-                for(final ListIterator<Node> i = children.listIterator(children.size());i.hasPrevious();) {
-                    nodes.push(i.previous());
+                for (Node child : children) {
+                    nodes.push(child);
                 }
                 node.setChildren(children);
             }
@@ -91,8 +91,8 @@ public class ReverseAstBuilder<T> {
         return new ReverseAst(objectsToNodes, pointersToNodes);
     }
 
-    private List<Node> handleMap(Map<?, ?> map, JsonPointer pointer, Marker defaultMarker) {
-        final List<Node> nodes = new LinkedList<>();
+    private Deque<Node> handleMap(Map<?, ?> map, JsonPointer pointer, Marker defaultMarker) {
+        final Deque<Node> nodes = new LinkedList<>();
         final Marker marker = getMarker(map).orElse(defaultMarker);
 
         for (Map.Entry<?, ?> entry : map.entrySet()) {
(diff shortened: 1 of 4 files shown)


server/src/main/java/de/zalando/zally/util/ast/ReverseAstBuilder.java
```diff
@@ -60,9 +64,11 @@
     public ReverseAst build() throws ReverseAstException {
         while (!nodes.isEmpty()) {
             Node node = nodes.pop();
-
+            if (objectsToNodes.containsKey(node.object)) {
+                continue;
+            }
```
roxspring: This is the important bit short-circuiting the recursion when we've seen the object before.

As a side-effect, we now only find the _first_ pointer to a node, not all pointers, and not defaulting to the final one. This in itself breaks `ReverseAstTest`'s `.ast prefers pointers to shared locations` test so the rest of the changes address that by ensuring we explore shared definitions first and the `/paths` last.

tkrop: Sounds good, but if this is the short-circuit, what do we need to change the rest? I would prefer to stick to the minimal `Deque` instead of using the bloated `List` interface.

roxspring: Attempted to explain that immediately above but clearly I need another attempt.

The overall `ReverseAstBuilder.nodes` makes perfect sense as a `Deque` since it's being used as a stack to manage a depth-first traversal of the object tree. The various `handle*()` methods identify a collection of child nodes to be added to that queue. To make the result deterministic and prefer we need to control the order of those children. Specifically, now that we're short-circuiting, we need to ensure that the shared definitions (e.g. `/components`, `/definitions`, `/responses`) are processed before `/paths` so that the canonical pointer for a shared object is the shared definition rather than the site of the first usage.

To me, this ordered collection of children is most honestly represented as a `List`, whose elements get added to the front of the `nodes` queue in reverse order in order to maintain the intended traversal order. Non-destructive reverse traversal of a `Deque` doesn't seem to be an option, so the `Deque` alternative would be to make the contract for the `handle*()` methods to return a queue of children to traverse already in reverse order. So sticking to a `Deque` is certainly possible but I'm not sure that it's clear.

roxspring: To be clear though... if you still really prefer the `Deque` only implementation then let me know and I'll make the change.

tkrop: Your argumentation is convincing, I can live with both solutions. Still I would suggest to be consistent in application and adapt lines in 41 and 81-82.

roxspring: Meh. Turns out the switch from `Deque` to `List` wasn't a necessary part of the change.

Have switched back to the original code using `Deque` only and tests pass. In a way, my assessment above was correct since it passes because the `handle*()` methods _already_ returned the queue in reverse order, and therefore they end up in the correct order in the depth-first traversal queue. Feels a little like it works by accident, but I'm not sure how to improve on the clarity.

(Switching everything to operate in terms of `List` was pretty ugly so didn't pursue that approach far)

tkrop:
:+1:

tkrop:
:+1:

ddufourboivin-zalando:
👍

PR #785, opened 2018-08-10
Flexible Regex whitelist for PluralizeResourceNamesRule

### LINKED ISSUE #780: More flexible whitelist for PluralizeResourceNamesRule
Generally, I like the idea of `PluralizeResourceNamesRule` but internally we have it disabled due to fear of too many "false positives":

1. We have a standard `GET /status` API on each service used for basic health checks. Clearly, the resource here is not plural but we want to accept it anyway. 

2. We also have a whole bunch of `POST /things/{id}/action` APIs which may not be ideal from a RESTful point of view but are a regular solution all the same and we're willing to live with them. 

3. I'm in constant fear that we're going to find terms which the `PlingStemmer` doesn't correctly identify as plurals and I don't really want to open an IDE for every exception we need to make.

Problem (1) can be addressed by configuring `whitelist_prefixes` and that suggests that adding a `whitelist_suffixes` configuration might be a reasonable solution to (2). Given that elsewhere we've become comfortable using regex in configurations I'm tempted to suggest we just use a single `whitelist` of regexes `^/status/`, `/action$` instead. Thoughts?

Problem (3) screams out for a configurable whitelist of plural terms. What I'm not certain about is whether this should be specific to `PluralizeResourceNamesRule` or a whitelist shared with `PluralizeNamesForArraysRule` and future uses of `isPlural(String)`?

roxspring:
- Updated to use a whitelist of Regex patterns rather than just prefixes. Each matching regex is replaced with a `/` before the path is split into components and checked. So you can use:
  - `/api/` to whitelist that term anywhere in the path
  - `^/status/$` to whitelist a fixed standard singular path
  - `^/legacy/.*` to whitelist any path starting with a prefix
  - lots of other possibilities using regexes
- Also refactored to use Context #714 

Closes #780


roxspring:
Looks like I have some integration test expectations to fix... commit coming soon.

server/src/main/java/de/zalando/zally/rule/zalando/PluralizeResourceNamesRule.kt
```diff
@@ -17,23 +18,33 @@ import org.springframework.beans.factory.annotation.Autowired
...
         }
-        return if (res != null && res.isNotEmpty()) {
-            val desc = res.map { "'${it.first}'" }.toSet().joinToString(", ")
-            val paths = res.map { it.second }
-            Violation(String.format(description, desc), paths)
-        } else null
     }
+
+    fun components(path: String): List<String> {
+        var filtered = "/$path/"
```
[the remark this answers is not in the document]
roxspring: Hadn't come across `.fold` before - nice! Not sure it adds much in this situation but avoiding a mutable variable is good, and it allows things to be chained into a single line more easily... which can be good (??)

Good point re `whitelist`. I'll break out a separate `sanitize(path, whitelist)` method and make the intent clearer.

server/src/main/java/de/zalando/zally/rule/zalando/PluralizeResourceNamesRule.kt
```diff
@@ -17,23 +18,42 @@ import org.springframework.beans.factory.annotation.Autowired
...
+    fun validate(context: Context): List<Violation> {
+        return context.validatePaths { (path, definition) ->
+            val sanitized = sanitize(path, whitelist)
+            components(sanitized)
+                    .filter { violatingComponent(it) }
+                    .map { violation(context, it, definition) }
+        }
+    }
+
+    private fun components(path: String): List<String> {
```
tkrop: How about `pathSegments`? Component is anything :).

server/src/main/java/de/zalando/zally/rule/zalando/PluralizeResourceNamesRule.kt
```diff
@@ -17,23 +18,42 @@ import org.springframework.beans.factory.annotation.Autowired
...
+            updated.replace(regex, slash)
         }
-        return if (res != null && res.isNotEmpty()) {
-            val desc = res.map { "'${it.first}'" }.toSet().joinToString(", ")
-            val paths = res.map { it.second }
-            Violation(String.format(description, desc), paths)
-        } else null
     }
+
+    private fun violatingComponent(it: String) =
```
tkrop: I would liked `isNoneViolating(PathSegment)` ...

server/src/main/java/de/zalando/zally/rule/zalando/PluralizeResourceNamesRule.kt
```diff
@@ -17,23 +18,42 @@ import org.springframework.beans.factory.annotation.Autowired
...
-        val res = swagger.paths?.keys?.flatMap { path ->
-            val allParts = path.split("/".toRegex())
-            val partsToCheck = if (allParts.size > 1 && allowedPrefixes.contains(allParts.first())) allParts.drop(1)
-            else allParts
-
-            partsToCheck.filter { s -> !s.isEmpty() && !PatternUtil.isPathVariable(s) && !isPlural(s) }
-                .map { it to path }
+    fun validate(context: Context): List<Violation> {
+        return context.validatePaths { (path, definition) ->
+            val sanitized = sanitize(path, whitelist)
```
tkrop: IMHO would even be okay to join this with the next line now :)

server/src/main/java/de/zalando/zally/rule/zalando/PluralizeResourceNamesRule.kt
```diff
@@ -17,23 +18,42 @@ import org.springframework.beans.factory.annotation.Autowired
...
+                    .filter { violatingComponent(it) }
+                    .map { violation(context, it, definition) }
+        }
+    }
+
+    private fun components(path: String): List<String> {
+        return path.split(slashes).filter { !it.isEmpty() }
+    }
+
+    private fun sanitize(path: String, regexList: List<Regex>): String {
```
tkrop: How about `sanitize(d)Path`?

roxspring: All done.

tkrop:
@roxspring can you resolve the conflict, so and merge the master? Than we are ready to merge :).

tkrop:
:+1:

maxim-tschumak:
:+1:

PR #793, opened 2018-08-10
Error Detection and Structural Changes around parsing

ddufourboivin-zalando:
Those changes were first made to introduce some new features that in the end are no longer needed. However, some of us judge the structural changes interesting so I am going ahead with a PR on this.

Highlights:
- clearer result after content parsing 
  - before: null or not-null
  - now: clear separation between _Not Applicable_, _Error_ and _Success_
- pre-conversion checks in case of Swagger-to-OpenAPI scenarios
  - some cases cause exceptions at runtime. Patching the _Swagger_ object before conversion and checking for potential violations.
  - integrated changes recently made to patch the absence of OAUTH2 scopes (see recent pull requests)
- tests improvements and formatting


ddufourboivin-zalando:
Replaced by [this PR](https://github.com/zalando/zally/pull/794)

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
```diff
@@ -122,58 +128,138 @@ class DefaultContext(override val source: String, openApi: OpenAPI, swagger: Swa
...
+            }
+
+            //
+            // Resolve
+            //
+            try {
+                ResolverFully(true).resolveFully(convertResult.openAPI)
+            } catch (e: NullPointerException) {
+                log.warn("Failed to fully resolve Swagger schema.", e)
+            }
```
roxspring: Given the three distinct phases might it be clearer to split this into 3 distinct methods along the following lines:

```kotlin
        fun createSwaggerContext(content: String): ContentParseResult<Context> {
            val parseResult = parseSwaggerContext(content)
            if (parseResult !is ContentParseResult.Success) return parseResult

            val convertResult = convertSwaggerContext(parseResult)
            if (convertResult !is ContentParseResult.Success) return convertResult

            val resolveResult = resolveSwaggerContext(convertResult)
            if (resolveResult !is ContentParseResult.Success) return resolveResult

            val context = DefaultContext(content, convertResult.openAPI, parseResult.swagger)
            return ContentParseResult.Success(context)
        }
```

Suggest if doing so then take similar approach with `createOpenApiContext()` above.

server/src/main/java/de/zalando/zally/rule/RulesValidator.kt
```diff
@@ -1,26 +1,48 @@
...
+abstract class RulesValidator<RootT : Any>(val rules: RulesManager) : ApiValidator {
 
     private val log = LoggerFactory.getLogger(RulesValidator::class.java)
     private val reader = ObjectTreeReader()
 
-    override fun validate(content: String, policy: RulesPolicy): List<Result> {
-        val root = parse(content) ?: return emptyList()
+    private val useOpenApiRule: RuleDetails by lazy {
+        rules.rules.firstOrNull { it.rule.id == UseOpenApiRule.id }
+            ?: throw IllegalStateException("Rule 'UseOpenApi' with ID ${UseOpenApiRule.id} must be registered in 'RulesManager'.")
```
roxspring: Shouldn't this be `'UseOpenApiRule'`? Seems odd to drop the suffix here.

server/src/main/java/de/zalando/zally/rule/RulesValidator.kt
```diff
@@ -1,26 +1,48 @@
...
+                        description = violation.description,
+                        violationType = useOpenApiRule.rule.severity,
+                        pointer = violation.pointer)
+                }
+            is Success ->
+                rules
+                    .checks(policy)
+                    .filter { details -> isCheckMethod(details, parseResult.root) }
+                    .flatMap { details -> invoke(details, parseResult.root) }
+                    .sortedBy(Result::violationType)
```
roxspring: This all becomes super clear 👍

server/src/main/java/de/zalando/zally/rule/RulesValidator.kt
ddufourboivin-zalando: Applying to the [open PR](https://github.com/zalando/zally/pull/794) (this one was canceled)

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
ddufourboivin-zalando: The only way that will work type-wise is to introduce `Disjunction` (biased `Either`). The library _FunKTionale_ is great for this. I will propose an update.

ddufourboivin-zalando: Nevermind that. Too overkill just for a simple usecase. Adding a `flatMap` to `ContentParseResult`. Short, clean and nice.

ddufourboivin-zalando: Applying to the [open PR](https://github.com/zalando/zally/pull/794) (this one was canceled)

PR #794, opened 2018-08-10
content parsing structure change + test improvements (#773)

ddufourboivin-zalando:
Those changes were first made to introduce some new features that in the end are no longer needed. However, some of us judge the structural changes interesting so I am going ahead with a PR on this.

Highlights:
- clearer result after content parsing 
  - before: null or not-null
  - now: clear separation between _Not Applicable_, _Error_ and _Success_
- pre-conversion checks in case of Swagger-to-OpenAPI scenarios
  - some cases cause exceptions at runtime. Patching the _Swagger_ object before conversion and checking for potential violations.
  - integrated changes recently made to patch the absence of OAUTH2 scopes (see recent pull requests)
- tests improvements and formatting


ddufourboivin-zalando:
@roxspring : Please use THIS pull request. The other one you commented was already closed (canceled).

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
```diff
@@ -122,58 +132,145 @@ class DefaultContext(override val source: String, openApi: OpenAPI, swagger: Swa
...
+                .flatMap { resolveOpenApi(it) }
+                .map { DefaultContext(content, it.openAPI) }
+
+        fun createSwaggerContext(content: String): ContentParseResult<Context> =
+            parseSwagger(content).flatMap { parseResult ->
+                convertSwaggerToOpenAPI(parseResult)
+                    .flatMap { resolveSwagger(it) }
+                    .map { convertResult ->
+                        DefaultContext(content, convertResult.openAPI, parseResult.swagger)
+                    }
```
roxspring: Looks good but I'm struggling to get my head around the use of the terms `map` vs `flatMap` here. I see that `map` is being used to unconditionally transform to the same sort of `ContentParseResult` with the updated parameter type, which I don't have a problem with. But `flatMap` is being used to encode "if input is Success do XXX , else `map`", which is clearly what we want to do, but doesn't match what I'm used to a collection's `flatMap()` doing. 

I'd have been tempted to rename the `flatMap()` method... maybe to something like `mapSuccess()` or `mapIfSuccess()` but I can see that neither of those entirely captures the meaning either.

I'm guessing though, that I've just missed something about `map` vs `flatMap` and that this usage _is_ idiomatic?? (please feel free to educate me!)

ddufourboivin-zalando: In this case, it's a _biased_ `flatMap`/`map` approach (see _Disjunctions_ in _FunKTionale_, vs their _Either_ big brother that is unbiased).

But in _Scala_ and most _Kotlin_ functional libraries I saw so far, it's usually biased.

If a `Try` in Scala is already a failure, `map` returns the same `Try` as it is (only with the new type signature of the `map`.

Same thing for `Option` or `Future`.

I thought I'd the same standard (if it is one).

roxspring: Thanks for the pointers - it's starting to sink in but might take a while to get used to 😃

ddufourboivin-zalando: It took me some time as well when I started _Scala_ (before switching to _Kotlin_) to sink those in ;-) 

It's even more confusing in _Scala_ when you start playing with the for-comprehension construct.

This:
```scala
for {
  a <- x()
  b <- y(a)
  c <- z(a, b)
} yield Foo(a, b, c)
```

is syntaxic sugar for (let's say in _Kotlin_):

```scala
x().flatMap { a ->
  y(a).flatMap { b ->
    z(a, b).map { c ->
      Foo(a, b, c)
    }
  }
}
```

tkrop: Damned, it may be functional, but may brain rejects to grasp the mathematical function expressed by it. In the end I'm not even sure, that it is easier maintainable than doing the operations in loops crunching through lists of the changing results.

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
```diff
@@ -122,58 +132,145 @@ class DefaultContext(override val source: String, openApi: OpenAPI, swagger: Swa
...
+                Success(parseResult)
             }
-            return parseResult?.openAPI?.let {
-                ResolverFully(true).resolveFully(it) // workaround for NPE bug in swagger-parser
-                DefaultContext(content, it)
+        }
+
+        private fun resolveOpenApi(parseResult: SwaggerParseResult): ContentParseResult<SwaggerParseResult> {
+            try {
+                ResolverFully(true).resolveFully(parseResult.openAPI) // workaround for NPE bug in swagger-parser
```
tkrop: Shouldn't the NPE workaround be removed after implementing the guard?

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
```diff
@@ -122,58 +132,145 @@ class DefaultContext(override val source: String, openApi: OpenAPI, swagger: Swa
...
                         it.scopes = LinkedHashMap()
                     }
                 }
 
-                val convertResult = SwaggerConverter().convert(parseResult)
-                if (failOnParseErrors && convertResult.messages.orEmpty().isNotEmpty()) {
-                    val sep = "\n  - "
-                    val messageBulletList = parseResult.messages.joinToString(sep)
-                    throw RuntimeException("Swagger conversion to OpenAPI 3 failed with those errors:$sep$messageBulletList")
+            return violations
```
tkrop: Can't see adding of any violation that is added, so why is this provided mutable at the start?

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
```diff
@@ -122,58 +132,145 @@ class DefaultContext(override val source: String, openApi: OpenAPI, swagger: Swa
...
+            val preConvertViolations = preConvertChecks(parseResult)
+            if (preConvertViolations.isNotEmpty()) {
+                return ParsedWithErrors(preConvertViolations)
+            }
+
+            val convertResult = try {
+                SwaggerConverter().convert(parseResult)
+            } catch (t: Throwable) {
+                log.warn("Unable to convert specification from 'Swagger 2' to 'OpenAPI 3'. Error not covered by pre-checks.", t)
+                val violation = Violation("Unable to parse specification", JsonPointers.root)
```
tkrop: Nice. Lets inform consumers in addition, that the result is incomplete with respect to analyzing violations.

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
```diff
@@ -122,58 +132,145 @@ class DefaultContext(override val source: String, openApi: OpenAPI, swagger: Swa
...
+                return ParsedWithErrors(listOf(violation))
+            }
+            return if (convertResult.openAPI === null) {
+                if (convertResult.messages.orEmpty().isNotEmpty()) {
+                    ParsedWithErrors(convertResult.messages.mapNotNull(::errorToViolation))
+                } else {
+                    log.warn("Unable to convert specification from 'Swagger 2' to 'OpenAPI 3'. No error specified, but 'openAPI' is null.")
+                    val violation = Violation("Unable to parse specification", JsonPointers.root)
```
tkrop: Nice. Lets inform consumers in addition, that the result is incomplete with respect to analyzing violations.

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
```diff
@@ -122,58 +132,145 @@ class DefaultContext(override val source: String, openApi: OpenAPI, swagger: Swa
...
+            }
+
+            val convertResult = try {
+                SwaggerConverter().convert(parseResult)
+            } catch (t: Throwable) {
+                log.warn("Unable to convert specification from 'Swagger 2' to 'OpenAPI 3'. Error not covered by pre-checks.", t)
+                val violation = Violation("Unable to parse specification", JsonPointers.root)
+                return ParsedWithErrors(listOf(violation))
+            }
+            return if (convertResult.openAPI === null) {
```
tkrop: Can't we assure that the result is never null to simplify the error handling code and handle this situation above?

server/src/main/java/de/zalando/zally/rule/zalando/UseOpenApiRule.kt
```diff
@@ -19,7 +19,7 @@ import java.net.URL
 
 @Rule(
     ruleSet = ZalandoRuleSet::class,
-    id = "101",
+    id = UseOpenApiRule.id,
```
tkrop: What does this change help to achieve?

server/src/main/java/de/zalando/zally/rule/ContextRulesValidator.kt
```diff
@@ -12,8 +12,17 @@ import org.springframework.stereotype.Component
 @Component
 class ContextRulesValidator(@Autowired rules: RulesManager) : RulesValidator<Context>(rules) {
 
-    override fun parse(content: String): Context? =
-            DefaultContext.createOpenApiContext(content) ?: DefaultContext.createSwaggerContext(content)
+    override fun parse(content: String): ContentParseResult<Context> {
+        // first try to parse an OpenAPI (version 3+)
+        val parsedAsOpenApi = DefaultContext.createOpenApiContext(content)
+        return when (parsedAsOpenApi) {
```
tkrop: Lets express this simpler with `() ? x : y`

server/src/main/java/de/zalando/zally/rule/RulesValidator.kt
```diff
@@ -1,26 +1,48 @@
...
+            is ParsedWithErrors ->
+                parseResult.violations.map { violation ->
+                    Result(
+                        ruleSet = useOpenApiRule.ruleSet,
+                        rule = useOpenApiRule.rule,
+                        description = violation.description,
+                        violationType = useOpenApiRule.rule.severity,
+                        pointer = violation.pointer)
+                }
+            is Success ->
```
maxim-tschumak: Let's rename it to `ParsedSuccessfully`

server/src/test/java/de/zalando/zally/rule/DefaultContextTest.kt
```diff
@@ -1,86 +1,172 @@
...
-    fun `should recognize the used OpenAPI 3`() {
-        @Language("yaml")
-        val openapi3Context = DefaultContext.createOpenApiContext("""
-        openapi: 3.0.1
-        info:
-          title: Pets API
-          version: 1.0.0
-        paths: {}
-        """.trimIndent())!!
+    fun `SWAGGER -- error when info and path objects are missing`() {
```
maxim-tschumak: The test title doesn't represent the test behavior - please double check.

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
```diff
@@ -122,58 +132,145 @@ class DefaultContext(override val source: String, openApi: OpenAPI, swagger: Swa
...
+            } catch (e: NullPointerException) {
+                log.warn("Failed to fully resolve Swagger schema.", e)
+            }
+            return Success(convertResult)
+        }
+
+        private fun errorToViolation(error: String): Violation =
+            Violation(error, errorToJsonPointer(error))
+
+        private val attributeIsMissingRegEx = Regex("attribute [^ ]* is missing")
```
maxim-tschumak: I haven't seen any tests for that and the regex doesn't look right - please double check.
(I would assume a regex like `"attribute (.*) is missing"`)

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
ddufourboivin-zalando: @tkrop : There is nothing mathematical here. It is data transformation.

ddufourboivin-zalando: I can translate that to more imperative -- and easy to read -- code, if that is what is best for the project.

Since it was proposed that I re-use the structure that was created (`ContentParseResult`) for splitting the logic into sub-blocks, it felt natural to go for a `map`/`flatMap` approach, since it is in the end what we want: transform a first result into a second result into a final result.

server/src/main/java/de/zalando/zally/util/ast/JsonPointers.java
```diff
@@ -17,6 +17,9 @@
  * Utility to convert OpenAPI 3 JSON pointers to Swagger 2 pointers.
  */
 public final class JsonPointers {
+
+    public static final JsonPointer root = JsonPointer.compile("/");
```
roxspring: This pointer isn't pointing to the root node. It refers to an element in the root map with an empty name: 

```json
{ "": "This node is referenced" }
```

You should be using `JsonPointers.empty()` (which could be extracted as a constant admittedly)

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
ddufourboivin-zalando: To make it possible to simply add one when we see that case. Removed it and returning the static `emptyList` for the moment. I do not want to change the signature because it is already piped in a way that if there **is** a violation, it is easy to return it and the rest of the flow will react accordingly.

server/src/main/java/de/zalando/zally/rule/ContextRulesValidator.kt
ddufourboivin-zalando: That does not exist in Kotlin. Your options are:

1. the `if` construct
```kotlin
return if (parsedAsOpenApi is ContentParseResult.NotApplicable) {
  // if content was no OpenAPI, try to parse a Swagger (version 2)
  DefaultContext.createSwaggerContext(content)
} else {
  parsedAsOpenApi
}
```

2. the `when` construct
```kotlin
return when (parsedAsOpenApi) {
    is ContentParseResult.NotApplicable ->
        // if content was no OpenAPI, try to parse a Swagger (version 2)
        DefaultContext.createSwaggerContext(content)
    else ->
        parsedAsOpenApi
}
```

I personally find the `when` to be more readable.

It is a bit more readable when there are only 2 cases; it is a lot more readable when there are 3 or more and it allows you to avoid repeating your self (`foo ==` or `foo is` N times).

server/src/main/java/de/zalando/zally/rule/RulesValidator.kt
ddufourboivin-zalando: Fixed in commit 681833e1.

server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
ddufourboivin-zalando: I could not find a case that cause this code path to execute. I will simply remove this and use the root JsonPointer instead.

As of the regex, it would be matching "attribute " + all characters except spaces + " is missing".

server/src/main/java/de/zalando/zally/util/ast/JsonPointers.java
ddufourboivin-zalando: Fixed in commit 7a2c746f

server/src/main/java/de/zalando/zally/rule/zalando/UseOpenApiRule.kt
ddufourboivin-zalando: This ID is used also in `DefaultContext`. That is why I constant-icized it.

server/src/main/java/de/zalando/zally/rule/ContextRulesValidator.kt
tkrop: Why the f...? Don't mind.

tkrop:
:+1:

maxim-tschumak:
:+1:

server/src/main/java/de/zalando/zally/rule/ContextRulesValidator.kt
mkulak: ¯\_(ツ)_/¯ Kotlin doesn't have ternary (Scala doesn't have it as well BTW)

ddufourboivin-zalando: Well, it does sine the `if` construct does return a value. In a language like that, the ternary operator kind of becomes useless. It is still a lot easier to read (in my opinion), but purely syntaxic sugar. As in Java, C++, etc... it is needed, since the `if` does not resolve into a value.

Issue #712, opened 2018-05-17
Convert remaining server code from Java to Kotlin


roxspring:
In the OpenAPI 3 PR I [commented surprise](https://github.com/zalando/zally/pull/706/files/59e4895efa750befefe125a8ef55694aa5396227#r188894438) at adding more Java code to the codebase but figured I should elevate discussion. We've fully embraced Kotlin as a language choice for the rules, a few supporting classes, and many of the tests. There is a little (actually surprisingly little) cognitive dissonance mentally switching languages when switching between parts of the code, so I wonder if now is the time to fully embrace Kotlin and switch the rest of the server code over.

Are there good reasons not to make this switch?

PR #796, opened 2018-08-14
Convert `util` code to Kotlin

roxspring:
- Converted Java to Kotlin for `util` package, again with Intellij doing most of the work.
- Adapted some Java stream & optional logic to Kotlin idioms
- Adapted `Node` and `Marker` to `data` classes
- Will conflict with #794 initially, but is resolvable:
  - Either #794 is updated to not touch `JsonPointers.java`
  - Or I'll update this PR once that's in.

Leaves just `apireview` package to go.

Relates to #712


server/src/main/java/de/zalando/zally/util/JsonRawValueDeserializer.kt
```diff
@@ -0,0 +1,16 @@
...
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.JsonDeserializer
+
+import java.io.IOException
+
+class JsonRawValueDeserializer : JsonDeserializer<String>() {
+
+    @Throws(IOException::class)
+    override fun deserialize(jp: JsonParser, context: DeserializationContext): String {
+        return jp.readValueAsTree<TreeNode>().toString()
```
mkulak: Can be in short form: 
`override fun deserialize(jp: JsonParser, context: DeserializationContext): String = jp.readValueAsTree<TreeNode>().toString()`

server/src/main/java/de/zalando/zally/util/PatternUtil.kt
```diff
@@ -0,0 +1,76 @@
...
+    private const val SNAKE_CASE_PATTERN = "^[a-z0-9]+(?:_[a-z0-9]+)*$"
+    private const val KEBAB_CASE_PATTERN = "^[a-z]+(?:-[a-z]+)*$"
+    private const val VERSION_IN_URL_PATTERN = "(.*)/v[0-9]+(.*)"
+    private const val PATH_VARIABLE_PATTERN = "\\{.+\\}$"
+    private const val GENERIC_VERSION_PATTERN = "^\\d+\\.\\d+\\.\\d+$"
+    private const val PATTERN_APPLICATION_PROBLEM_JSON = "^application/(problem\\+)?json$"
+    private const val PATTERN_CUSTOM_WITH_VERSIONING = "^\\w+/[-+.\\w]+;v(ersion)?=\\d+$"
+
+    fun hasTrailingSlash(input: String): Boolean {
+        return input.trim { it <= ' ' }.endsWith("/")
```
mkulak: here I would suggest to use short form as well. It reduces moise for one-line methods.

server/src/main/java/de/zalando/zally/util/WordUtil.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import java.util.Arrays.asList
+
+object WordUtil {
+    private val PLURAL_WHITELIST = HashSet(asList("vat", "apis"))
+
+    fun isPlural(word: String): Boolean {
+        return if (PLURAL_WHITELIST.contains(word)) {
+            true
+        } else PlingStemmer.isPlural(word)
```
mkulak:
```kotlin
fun isPlural(word: String): Boolean = word in PLURAL_WHITELIST || PlingStemmer.isPlural(word)
```

server/src/main/java/de/zalando/zally/util/ast/JsonPointers.kt
```diff
@@ -0,0 +1,71 @@
...
+        ?.let { ptr ->
+            val ptrString = ptr.toString()
+            regexToReplacement
+                .filter { (regex, _) ->
+                    regex.matches(ptrString)
+                }
+                .map { (regex, replacement) ->
+                    regex.replace(ptrString, replacement)
+                }
+                .firstOrNull()
```
mkulak: `List.filter().map().firstOrNull()` is suboptimal in both readability and performance.
I suggest to write it like this:
```kotlin
val ptrString = ptr.toString()
val res = regexToReplacement.find { it.first.matches(ptrString) }
res?.let {  (regex, replacement) -> JsonPointer.compile(regex.replace(ptrString, replacement)) }
```

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.kt
```diff
@@ -0,0 +1,173 @@
...
+
+    private val skipMethods = HashSet<String>()
+    private val objectPointerCache = IdentityHashMap<Any, JsonPointer>()
+    private val methodPointerCache = IdentityHashMap<Any, IdentityHashMap<Method, JsonPointer>>()
+
+    internal class MethodCallRecorderException : Throwable {
+        constructor(message: String) : super(message)
+
+        constructor(message: String, cause: Throwable) : super(message, cause)
+    }
```
mkulak:
```kotlin
internal class MethodCallRecorderException(message: String, cause: Throwable? = null) : Throwable(message, cause)
```

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.kt
```diff
@@ -0,0 +1,173 @@
...
+
+    @Throws(MethodCallRecorderException::class)
+    private fun getGenericReturnValueType(m: Method): Class<*> {
+        val type = m.genericReturnType
+        if (type is ParameterizedType) {
+            val typeArgs = type.actualTypeArguments
+            return typeArgs[typeArgs.size - 1] as Class<*>
+        }
+        throw MethodCallRecorderException(m.returnType.toString())
```
mkulak: Can be this:
```kotlin
    private fun getGenericReturnValueType(m: Method): Class<*> =
        (m.genericReturnType as? ParameterizedType)?.let { it.actualTypeArguments.last() as Class<*> } 
                ?: throw MethodCallRecorderException(m.returnType.toString()) 
```

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.kt
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    fun skipMethods(vararg pointer: String): MethodCallRecorder<T> {
+        this.skipMethods.addAll(Arrays.asList(*pointer))
+        return this
+    }
```
mkulak:
```
        fun skipMethods(vararg pointer: String): MethodCallRecorder<T> { 
            skipMethods += pointer
            return this
        }
```

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.kt
```diff
@@ -0,0 +1,173 @@
...
+
+    @Suppress("UNCHECKED_CAST")
+    private fun <U> createProxy(`object`: U, parent: Method?): U {
+        val interceptor = createMethodInterceptor(`object` as Any, parent)
```
mkulak: If you put bound on your generic type ` private fun <U : Any> createProxy` it would save you from casting it to `Any`

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.kt
```diff
@@ -0,0 +1,173 @@
...
+                pointer = objectPointer!!.append(JsonPointers.escape(method, *arguments))
+                methodMap[method] = pointer
+            }
+        } else {
+            val methodMap = IdentityHashMap<Method, JsonPointer>()
+            pointer = objectPointer!!.append(JsonPointers.escape(method, *arguments))
+            methodMap[method] = pointer
+            methodPointerCache[`object`] = methodMap
+        }
+    }
```
mkulak:
```kotlin
        private fun updatePointer(obj: Any, method: Method, arguments: Array<Any>) {
            // Some methods should not be recorded in the JSON pointer string.
            if (method.name in skipMethods) {
                return
            }
            // In order to prevent multiple successive method calls the the proxied object to
            // endlessly append new fragments to the JSON pointer string, we must remember already
            // recorded pointers inside two caches:
            // objectPointerCache holds "base" pointers with the original objects on which methods are
            // called as keys.
            // methodPointerCache is a nested map that holds "method" pointers with the objects and the
            // called methods as keys.
            val objectPointer = objectPointerCache.getOrPut(obj) { pointer }
            val methodMap = methodPointerCache.getOrPut(obj) { IdentityHashMap<Method, JsonPointer>() }
            pointer = methodMap.getOrPut(method) { objectPointer.append(JsonPointers.escape(method, *arguments)) }
        }
```
(I haven't check this code extensively but should be something like this)

server/src/main/java/de/zalando/zally/util/ast/ReverseAst.kt
```diff
@@ -0,0 +1,65 @@
...
+    fun isIgnored(pointer: JsonPointer, ignoreValue: String): Boolean {
+        return isIgnored(this.pointersToNodes[pointer.toString()], ignoreValue)
+    }
+
+    private fun isIgnored(node: Node?, ignoreValue: String): Boolean {
+        if (node == null) {
+            return false
+        }
+        val ignored = isIgnored(node.marker, ignoreValue)
+        return ignored || node.hasChildren() && node.children.parallelStream().allMatch { c -> isIgnored(c.marker, ignoreValue) }
```
mkulak: Are we sure that usage of `parallelStream()` here leads to [performance improvement](https://stackoverflow.com/questions/23170832/java-8s-streams-why-parallel-stream-is-slower)?

server/src/main/java/de/zalando/zally/util/ast/ReverseAstBuilder.kt
```diff
@@ -0,0 +1,189 @@
...
+
+    private val nodes = LinkedList<Node>()
+    private val objectsToNodes = IdentityHashMap<Any, Node>()
+    private val pointersToNodes = HashMap<String, Node>()
+
+    class ReverseAstException internal constructor(message: String, cause: Throwable) : Exception(message, cause)
+
+    init {
+        nodes.push(Node(root as Any, JsonPointers.EMPTY, null))
+    }
```
mkulak: Can be combined with field declaration:
`private val nodes = mutableListOf(Node(root as Any, JsonPointers.EMPTY, null))`

server/src/main/java/de/zalando/zally/util/ast/ReverseAstBuilder.kt
```diff
@@ -0,0 +1,189 @@
...
+
+    fun withExtensionMethodNames(vararg names: String): ReverseAstBuilder<T> {
+        this.extensionMethodNames.addAll(listOf(*names))
```
mkulak: `extensionMethodNames += names`

server/src/main/java/de/zalando/zally/util/ast/ReverseAstBuilder.kt
```diff
@@ -0,0 +1,189 @@
...
+                val children = when {
+                    node.`object` is Map<*, *> -> handleMap(node.`object`, node.pointer, node.marker)
+                    node.`object` is List<*> -> handleList(node.`object`, node.pointer, node.marker)
+                    node.`object` is Set<*> -> handleSet(node.`object`, node.pointer, node.marker)
+                    node.`object` is Array<*> -> handleArray(node.`object`, node.pointer, node.marker)
+                    else -> handleObject(node.`object`, node.pointer, node.marker)
+                }
+                for (child in children) {
+                    nodes.push(child)
+                }
```
mkulak: `nodes += children`

server/src/main/java/de/zalando/zally/util/ast/ReverseAstBuilder.kt
```diff
@@ -0,0 +1,189 @@
...
+        if (map.containsKey(extensionName)) {
+            val value = map[extensionName]
+            if (value is String) {
+                return setOf(value)
+            }
+            if (value is Collection<*>) {
+                return value.map { it.toString() }.toSet()
+            }
+        }
+        return null
```
mkulak: Slightly better:
```kotlin
    private fun getVendorExtensions(map: Map<*, *>, extensionName: String): Collection<String>? =
        map[extensionName]?.let { value ->
            when (value) {
                is String -> setOf(value)
                is Collection<*> -> value.map { it.toString() }.toSet()
                else -> null
            }
        }
```

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.kt
```diff
@@ -67,69 +53,47 @@ class MethodCallRecorder<T>(`object`: T) {
...
-    }
+    private fun getGenericReturnValueType(m: Method): Class<*> =
+        (m.genericReturnType as? ParameterizedType)
+        ?.let { it.actualTypeArguments.last() as Class<*> }
+        ?: throw MethodCallRecorderException(m.returnType.toString())
 
-    fun skipMethods(vararg pointer: String): MethodCallRecorder<T> {
-        this.skipMethods.addAll(Arrays.asList(*pointer))
+    fun skipMethods(vararg methodNames: String): MethodCallRecorder<T> {
+        this.skipMethods += listOf(*methodNames)
```
mkulak: 1. There is no need to do `listOf(*methodNames)`, it can work directly as:
`skipMethods += methodNames`
2. I think `this.` is redundant here

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.kt
```diff
@@ -67,69 +53,47 @@ class MethodCallRecorder<T>(`object`: T) {
...
-                pointer = objectPointer!!.append(JsonPointers.escape(method, *arguments))
-                methodMap[method] = pointer
-            }
-        } else {
-            val methodMap = IdentityHashMap<Method, JsonPointer>()
-            pointer = objectPointer!!.append(JsonPointers.escape(method, *arguments))
-            methodMap[method] = pointer
-            methodPointerCache[`object`] = methodMap
-        }
+        val objectPointer: JsonPointer? = objectPointerCache.getOrPut(obj) { pointer }
```
mkulak: `pointer` is not nullable. Therefore objectPointer is also not-nullable

server/src/main/java/de/zalando/zally/util/ast/ReverseAst.kt
```diff
@@ -12,43 +12,37 @@ class ReverseAst internal constructor(private val objectsToNodes: Map<Any, Node>
...
-    fun getIgnoreValues(pointer: JsonPointer): Collection<String> {
-        return getIgnoreValues(this.pointersToNodes[pointer.toString()])
-    }
+    fun getIgnoreValues(pointer: JsonPointer): Collection<String> =
+        getIgnoreValues(this.pointersToNodes[pointer.toString()])
 
     private fun getIgnoreValues(node: Node?): Collection<String> {
-        if (node == null) {
-            return emptySet()
+        val markers = node?.let { getIgnoreValues(node.marker) } ?: emptySet()
```
mkulak: There is nice extension method `orEmpty()`, I think using it is a bit more idiomatic in this case:
`val markers = node?.let { getIgnoreValues(node.marker) }.orEmpty()`

server/src/main/java/de/zalando/zally/util/ast/ReverseAstBuilder.kt
```diff
@@ -44,85 +44,76 @@ class ReverseAstBuilder<T> internal constructor(root: T) {
...
-        return handleArray(set.toTypedArray(), pointer, marker)
-    }
+    private fun handleSet(set: Set<*>, pointer: JsonPointer, marker: Marker?): Deque<Node> =
+        handleArray(set.toTypedArray(), pointer, marker)
 
     private fun handleArray(objects: Array<*>, pointer: JsonPointer, marker: Marker?): Deque<Node> {
-        val nodes = LinkedList<Node>()
-
+        val nodes = ArrayDeque<Node>()
         for (i in objects.indices) {
```
mkulak: I propose to do this:
```kotlin
    private fun handleArray(objects: Array<*>, pointer: JsonPointer, marker: Marker?): Deque<Node> {
        val nodes = objects.filterNotNull().mapIndexed { i, value ->
            Node(value, pointer.append(JsonPointers.escape(i.toString())), marker)    
        }
        return ArrayDeque(nodes)
    }
```

server/src/main/java/de/zalando/zally/util/ast/ReverseAstBuilder.kt
```diff
@@ -158,18 +149,15 @@ class ReverseAstBuilder<T> internal constructor(root: T) {
         return null
     }
 
-    private fun getVendorExtensions(map: Map<*, *>, extensionName: String): Collection<String>? {
-        if (map.containsKey(extensionName)) {
+    private fun getVendorExtensions(map: Map<*, *>, extensionName: String): Collection<String>? =
+        map[extensionName]?.let {
             val value = map[extensionName]
```
mkulak: I think this is typo, should be 
```kotlin
map[extensionName]?.let { value ->
```

server/src/main/java/de/zalando/zally/util/WordUtil.kt
```diff
@@ -2,16 +2,8 @@ package de.zalando.zally.util
 
 import javatools.parsers.PlingStemmer
 
-import java.util.HashSet
-
-import java.util.Arrays.asList
-
 object WordUtil {
-    private val PLURAL_WHITELIST = HashSet(asList("vat", "apis"))
+    private val PLURAL_WHITELIST = hashSetOf("vat", "apis")
```
mkulak: since we don't need to mutate this set this should be `setOf`

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.kt
roxspring: Oh you're right - Intellij didn't like it but I'd left the * in place.

server/src/main/java/de/zalando/zally/util/PatternUtil.kt
```diff
@@ -0,0 +1,50 @@
+package de.zalando.zally.util
+
+/**
+ * Utility library for matching common patterns
+ */
+object PatternUtil {
```
ddufourboivin-zalando: Here the regex are recompiled every time they are used. 

Instead of having

```kotlin
private const val LOWER_CASE_HYPHENS_PATTERN = "^[a-z-]*$"
```

why not keep

```kotlin
private val LOWER_CASE_HYPHENS_PATTERN = "^[a-z-]*$".toRegex()
```

so they are compiled once and never again?

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.kt
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * MethodCallRecorder creates a Proxy around an obj, typically Swagger or OpenApi, and records
+ * successive method calls as a JSON pointer. Any returned property that is a complex obj or
+ * a container will be also wrapped in a Proxy. Null-values will be transformed into non-null
+ * Proxies as well, making it possible to generate JSON pointers for all possible properties.
+ */
+class MethodCallRecorder<T : Any>(obj: T) {
+
+    val proxy: T = createProxy(obj)
+    var pointer: JsonPointer = JsonPointers.EMPTY
```
ddufourboivin-zalando: Public var means that the outside world can get AND set it. Is that really what we want here?
A bit of a purism question, I agree; but worth asking.

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.kt
```diff
@@ -0,0 +1,137 @@
...
+    private fun isGetterMethod(m: Method): Boolean = m.name.startsWith("get") && m.returnType != null
+
+    private fun isPrimitive(o: Any): Boolean = isPrimitive(o.javaClass)
+
+    private fun isPrimitive(c: Class<*>): Boolean = Util.PRIMITIVES.contains(c)
+
+    private fun isGenericContainer(o: Any): Boolean = o is Collection<*> || o is Map<*, *>
+
+    @Suppress("UNCHECKED_CAST")
+    @Throws(MethodCallRecorderException::class)
```
ddufourboivin-zalando: Is this `@Throws` necessary?

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.kt
```diff
@@ -0,0 +1,137 @@
...
+    private fun <T> createInstance(c: Class<T>): T {
+        return when {
+            c.isAssignableFrom(Map::class.java) -> HashMap<Any, Any>() as T
+            c.isAssignableFrom(List::class.java) -> ArrayList<Any>() as T
+            c.isAssignableFrom(Set::class.java) -> HashSet<Any>() as T
+            c.isArray -> arrayOfNulls<Any>(0) as T
+            else -> try {
+                c.getConstructor().newInstance()
+            } catch (e: ReflectiveOperationException) {
+                throw MethodCallRecorderException("Cannot create " + c.toString(), e)
```
ddufourboivin-zalando: Please use

```kotlin
"Cannot create $c"
```

instead of

```kotlin
"Cannot create " + c.toString()
```

server/src/main/java/de/zalando/zally/util/JsonRawValueDeserializer.kt
```diff
@@ -0,0 +1,15 @@
...
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.core.TreeNode
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.JsonDeserializer
+
+import java.io.IOException
+
+class JsonRawValueDeserializer : JsonDeserializer<String>() {
+
+    @Throws(IOException::class)
```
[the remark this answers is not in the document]
roxspring: No idea. Just what Intellij's conversion did and I didn't have a good reason to correct. Happy to try and remove.

server/src/main/java/de/zalando/zally/util/PatternUtil.kt
roxspring: Wondered about that myself. Happy to change.

server/src/main/java/de/zalando/zally/util/ast/ReverseAst.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    fun isIgnored(pointer: JsonPointer, ignoreValue: String): Boolean =
+        isIgnored(this.pointersToNodes[pointer.toString()], ignoreValue)
+
+    private fun isIgnored(node: Node?, ignoreValue: String): Boolean {
+        return when {
+            node != null && isIgnored(node.marker, ignoreValue) -> true
+            node != null && node.hasChildren() && node.children.all { c -> isIgnored(c.marker, ignoreValue) } -> true
+            else -> false
+        }
```
ddufourboivin-zalando: We could maybe avoid the double checking of `node != null`.

```kotlin
return node
  ?.let { isIgnored(it.marker, ignoreValue) || it.children.all { c -> isIgnored(c.marker, ignoreValue) }
  ?: false
```
or something similar would avoid that useless second comparison.

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.kt
roxspring: Does seem unlikely! Will see if it can be tightened up.

server/src/main/java/de/zalando/zally/util/ast/ReverseAst.kt
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    private fun isIgnored(marker: Marker?, ignoreValue: String): Boolean =
+        marker != null &&
+        Marker.TYPE_X_ZALLY_IGNORE == marker.type &&
+        marker.values.contains(ignoreValue)
+
```
ddufourboivin-zalando: I really love this "operator alignment" on multi lines. I had that in C# with VS+ReSharper, but not able to recreate it with IntelliJ. Any tips for me?

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.kt
roxspring: (the previous answer is general)

server/src/main/java/de/zalando/zally/util/ast/ReverseAstBuilder.kt
```diff
@@ -0,0 +1,170 @@
...
+        })
+
+    @Throws(ReverseAstException::class)
+    private fun handleObject(obj: Any, pointer: JsonPointer, defaultMarker: Marker?): Deque<Node> {
+        val nodes = ArrayDeque<Node>()
+        val marker = getMarker(obj) ?: defaultMarker
+
+        for (m in traversalMethods(obj.javaClass)) {
+            val name = m.name
+            try {
```
ddufourboivin-zalando: Since there is no _per instance recovery_, I would suggest to move the `try` outside of the `for` loop.
I am not exactly sure about the potential cost of the entry/exit of a `try` block n-times.

server/src/main/java/de/zalando/zally/util/PatternUtil.kt
roxspring: Done in 1ef8764

server/src/main/java/de/zalando/zally/util/ast/MethodCallRecorder.kt
roxspring: 1ef8764 has a `private set` on following line so that only the getter is `public`.

ddufourboivin-zalando:
👍

maxim-tschumak:
:+1:

ddufourboivin-zalando:
👍

ddufourboivin-zalando:
👍

maxim-tschumak:
:+1:

PR #800, opened 2018-08-17
feat(server): migrate Zalando rule 171 to use the new context object …

maxim-tschumak:
Related to #714


server/src/main/java/de/zalando/zally/rule/zalando/FormatForNumbersRule.kt
```diff
@@ -2,52 +2,33 @@ package de.zalando.zally.rule.zalando
...
-        }
-        val result = fromObjects + fromParams + fromPathParams
-        return if (result.isNotEmpty()) {
-            val (props, paths) = result.unzip()
-            val properties = props.flatten().toSet().joinToString(", ")
-            Violation(description + properties, paths)
-        } else null
-    }
+    fun checkNumberFormat(context: Context): List<Violation> =
+        allSchemas(context.api)
```
ddufourboivin-zalando: That would be a good candidate for a method `validateSchemas` (if not already there) in the context.

If not, I would suggest making it an extension method (again, I think is already a file with a collection of extension methods for `OpenAPI` and/or `Swagger`).

#minorSuggestion ;)

ddufourboivin-zalando:
👍

maxim-tschumak:
:+1:

server/src/main/java/de/zalando/zally/rule/zalando/FormatForNumbersRule.kt
maxim-tschumak: Sorry, I've seen this comment after I've merged. I'll incorporate this change in the next PR

PR #802, opened 2018-08-20
feat(server): migrate PluralizeNamesForArraysRule rule to use the new context object

maxim-tschumak:
Related to #714


tkrop:
:+1:

maxim-tschumak:
:+1:

PR #804, opened 2018-08-20
Bugfix: Do not check format of null schemas

### LINKED ISSUE #803: FormatForNumbersRule explodes on missing schemas
The following definition results in a `NullPointerException` because the examples have a `null` schema which we then fail to check for number format:

```yaml
swagger: '2.0'
info:
  title: Minimal API
  version: 1.0.0
paths:
  /handlers:
    get:
      responses:
        200:
          description: OK
          examples:
            application/json:
              - name: Named thing
```

roxspring:
Closes #803

### DIFF
--- a/server/src/main/java/de/zalando/zally/util/OpenAPIObjectUtils.kt
+++ b/server/src/main/java/de/zalando/zally/util/OpenAPIObjectUtils.kt
@@ -8,17 +8,18 @@ import io.swagger.v3.oas.models.media.Schema
  * @param api OpenAPI 3 specification object
  * @return a collection of schemas
  */
-fun allSchemas(api: OpenAPI): Collection<Schema<Any>> = api.components.schemas.orEmpty().values +
-    api.components.responses.values.flatMap { it.content.values.map { it.schema } } +
-    api.components.requestBodies.values.flatMap { it.content.values.map { it.schema } } +
+fun allSchemas(api: OpenAPI): Collection<Schema<Any>> =
+    api.components.schemas.orEmpty().values +
+    api.components.responses.values.flatMap { it.content.values.mapNotNull { it.schema } } +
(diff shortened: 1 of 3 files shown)


server/src/main/java/de/zalando/zally/util/OpenAPIObjectUtils.kt
```diff
@@ -8,17 +8,18 @@ import io.swagger.v3.oas.models.media.Schema
  * @param api OpenAPI 3 specification object
  * @return a collection of schemas
  */
-fun allSchemas(api: OpenAPI): Collection<Schema<Any>> = api.components.schemas.orEmpty().values +
-    api.components.responses.values.flatMap { it.content.values.map { it.schema } } +
-    api.components.requestBodies.values.flatMap { it.content.values.map { it.schema } } +
+fun allSchemas(api: OpenAPI): Collection<Schema<Any>> =
+    api.components.schemas.orEmpty().values +
+    api.components.responses.values.flatMap { it.content.values.mapNotNull { it.schema } } +
+    api.components.requestBodies.values.flatMap { it.content.values.mapNotNull { it.schema } } +
```
ddufourboivin-zalando: Which editor and settings are you using to have the multiline operations aligned like this?

ddufourboivin-zalando:
👍

server/src/main/java/de/zalando/zally/util/OpenAPIObjectUtils.kt
roxspring: IntelliJ defaults, but with some hand editing since you guys keep correcting the 8 character "Continuation indent". The linebreak after `=` was by hand primarily because lining it up made it easier to keep my head around what was going on as I hunted down the `NullPointerException`.

maxim-tschumak:
:+1:

PR #805, opened 2018-08-21
714 query parameter collection format rule migration

maxim-tschumak:
Related to #714.
Related to the Guideline change proposed in https://github.com/zalando/restful-api-guidelines/pull/440


server/src/main/java/de/zalando/zally/rule/zalando/QueryParameterCollectionFormatRule.kt
```diff
@@ -1,46 +1,29 @@
...
-    fun createViolation(paths: List<String>): Violation {
-        return Violation(violationDescription, paths)
-    }
+    fun checkParametersCollectionFormat(context: Context): List<Violation> =
+        if (context.isOpenAPI3())
+            context.api.getAllParameters().entries
+                .filter { "query" == it.value.`in` && "array" == it.value.schema.type }
+                .filter { it.value.style == null || allowedStyle != it.value.style }
+                .map { context.violation(description, it.value) }
+        else emptyList()
```
ddufourboivin-zalando: I would here either use `.values` (since the key is not needed) or use deconstruction `{ (_, value) -> ...}`. Both ways would avoid the `.value` call.

server/src/main/java/de/zalando/zally/util/OpenApiUtil.kt
```diff
@@ -35,3 +36,33 @@ fun OpenAPI.getAllHeaders(): Set<HeaderElement> {
 
     return fromParams + fromPaths
 }
+
+/**
+ * Returns all defined schemas of an API specification
+ * @return a collection of schemas
+ */
+fun OpenAPI.getAllSchemas(): Collection<Schema<Any>> = this.components.schemas.orEmpty().values +
```
ddufourboivin-zalando: @maxim-tschumak :
What happened to the long exhaustive version of that that I did some weeks ago...? Did we discarded it?

server/src/main/java/de/zalando/zally/rule/zalando/QueryParameterCollectionFormatRule.kt
maxim-tschumak: Good hint!

server/src/main/java/de/zalando/zally/util/OpenApiUtil.kt
maxim-tschumak: Hm... I don't know. This is the function I've implemented last week and now I've just moved it to the right class (as you proposed).

ddufourboivin-zalando:
👍

maxim-tschumak:
:+1:

PR #806, opened 2018-08-22
Migrating Secure with OAuth2 Zalando Rule to use the new Context Object

maxim-tschumak:
Related to #714


cli/zally/integration_test.go
```diff
@@ -69,7 +69,7 @@ func TestIntegrationWithSomeOtherLocalYamlFile(t *testing.T) {
 
 		assert.Contains(t, out, "Provide API Specification using OpenAPI")
 		assert.Contains(t, out, "MUST violations: 40")
-		assert.Contains(t, out, "SHOULD violations: 24")
+		assert.Contains(t, out, "SHOULD violations: 23")
```
roxspring: Not specifically about this change but, what are these tests really achieving?

If people are carefully reviewing the 24 original violations and comparing them to the 23 new violations before adjusting the expectation, then that seems like an awful lot of effort being spent collectively.

If people are mechanically updating the number without investigation, which I've definitely been guilty of, then this seems like a mostly pointless chore unlikely to catch genuine mistakes.

If these are truly testing the integration, do we really need 6 of them? wouldn't 1 do??

If these are testing our rules against a known set of specs, why not do it in the java layer so that we can see the failures quicker?

ddufourboivin-zalando:
👍

ddufourboivin-zalando:
👍

maxim-tschumak:
:+1:

PR #810, opened 2018-08-22
feat(server): migrate extensible enum rule to context object to support openapi 3

maxim-tschumak:
Related to #714


tkrop:
:+1:

maxim-tschumak:
:+1:

PR #820, opened 2018-08-24
feat(server): migrate pascal case http headers rule to context object to support openapi 3

maxim-tschumak:
Related to #714

### DIFF
--- a/server/src/main/java/de/zalando/zally/util/OpenApiUtil.kt
+++ b/server/src/main/java/de/zalando/zally/util/OpenApiUtil.kt
@@ -22,7 +22,7 @@ fun OpenAPI.getAllHeaders(): Set<HeaderElement> {
         .map { HeaderElement(it.key, it.value) }
         .toSet()
 
-    val fromComponentsParams = components.parameters.orEmpty().values.extractHeaders()
+    val fromComponentsParams = components?.parameters.orEmpty().values.extractHeaders()
 
     val fromPaths = paths.orEmpty().flatMap { (_, path) ->
         val fromPathParameters = path.parameters.extractHeaders()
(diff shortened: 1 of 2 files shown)


server/src/test/java/de/zalando/zally/rule/zalando/PascalCaseHttpHeadersRuleTest.kt
```diff
@@ -1,68 +1,45 @@
...
 
-    private val rule = PascalCaseHttpHeadersRule(testConfig)
+    private val rule = PascalCaseHttpHeadersRule()
 
     @Test
-    fun simplePositiveCase() {
-        val swagger = swaggerWithHeaderParams("Right-Name")
-        assertThat(rule.validate(swagger)).isNull()
-    }
+    fun `checkHttpHeaders should return violation if not pascal case is used`() {
```
tkrop: This one negative check is definitely not enough as long as we do not have tests for `de.zalando.zally.util.OpenApiUtil.kt`. We need to add a check for each potential header location, to make sure that we do not miss one. Anyhow, I would prefer similar unit tests on `de.zalando.zally.util.OpenApiUtil.kt`.

maxim-tschumak: Agree, we should unit test `OpenApiUtil` class.

maxim-tschumak:
@tkrop please re-review, the improvements have been implemented

tkrop:
:+1:

maxim-tschumak:
:+1:

maxim-tschumak:
@tkrop please re-review, the improvements have been implemented

PR #821, opened 2018-08-24
feat(server): migrate limit number of subresources rule to context object to support openapi 3

maxim-tschumak:
Related to #714


tkrop:
:+1:

PR #824, opened 2018-09-03
Added AtMostOneBodyParameterRule validating Swagger proxy from Context

### LINKED ISSUE #822: Context should allow explicit checking of Swagger models
I have an in-house rule `AtMostOneBodyParameter` which ensures a POST operation has no more than 1 body parameter. It was originally implemented against `Swagger` and can't be converted to an `Context` based check since the OpenAPI model available for checking doesn't allow people to make such as stupid mistake in the first place.  I'd like to keep the rule in place while swagger documents are still in use, but upgrade it to use `Context` so that `x-zally-ignore` can be supported anywhere. Also because I suspect that we might want to drop support for old style rules in the near future as #714 is completed.

Here's what I'm thinking:
- `Context` exposes `val swagger: Swagger?` populated only when the upconvert occurred
  - Would this be better in a separate `SwaggerContext` interface so that it's not accidentally used elsewhere??
- `DefaultContext.recorder` renamed to `openApiRecorder`
- `DefaultContext.swaggerRecorder: MethodCallRecorder<Swagger>?` added
- `DefaultContext.violation(String,JsonPointer?)` would need modifying to decide which recorder to take the default pointer from.
  - Perhaps lookup both and choose the deepest??
  - Might also need to explicitly set both recorders point to EMPTY before running each rule

Any objections or improvements?

roxspring:
- `Context` exposes `val swagger: Swagger?` populated only when the upconvert occurred
- `DefaultContext.recorder` renamed to `openApiRecorder`
- `DefaultContext.swaggerRecorder: MethodCallRecorder<Swagger>?` added
- `DefaultContext.violation(String,JsonPointer?)` modified to decide which recorder to take the default pointer from, taking the first non-EMPTY pointer from `pointer`, `swaggerRecorder.pointer`, `openApiRecorder.pointer`.
- Added `AtMostOneBodyParameterRule` enforcing Swagger 2.0 requirement that there can only be one body parameter.
  - Still suspicious that the recorders might need resetting to EMPTY between checks, but couldn't demonstrate the need so haven't made the change.

Closes #822


server/src/main/java/de/zalando/zally/rule/zally/AtMostOneBodyParameterRule.kt
```diff
@@ -0,0 +1,32 @@
...
+    ruleSet = ZallyRuleSet::class,
+        id = "M009",
+        severity = Severity.MUST,
+        title = "At Most One Body Parameter"
+)
+class AtMostOneBodyParameterRule {
+    val description = "No more than one body parameter can be associated with an operation"
+
+    @Check(Severity.MUST)
+    fun validate(context: Context): List<Violation> = context.swagger?.paths
```
maxim-tschumak: Let's mark this check as `Swagger 2 only`

maxim-tschumak: Thanks for the clarification! I wouldn't go for different context objects - one should be enough, and also much easier to understand for the "newcomers". But let's not overengineer at this point - I think a simple function documentation (comment) is enough for now. We still can improve if there will be a need.

maxim-tschumak:
:+1:

tkrop:
:+1:

PR #827, opened 2018-09-06
Fix Use429HeaderForRateLimitRule NullPointerException on paths that have no response

### LINKED ISSUE #787: TargetInvocationException: null
# How to reproduce

Lint this API in Zally.

```yaml
swagger: "2.0"
info:
  x-audience: business-unit-internal
  x-api-id: d0184f38-b98d-11e7-9c56-68f728c1ba70
  title: foo
  description: bar
  version: 1.0.0
  contact:
    email: a@b.c
    name: James Bond
    url: https://www.mi6.co.gov
securityDefinitions:
  oauth2:
    type: oauth2
    flow: application
    tokenUrl: https://identity.example.com/oauth2/token
    scopes:
      lalala.read: Access right needed to read
      lalala.write: Access right needed to write
paths: 
  /foos/bars:
    get:
      description: asd
```

The front end will answer:

> ERROR
> Ooops something went wrong!

The server will log:
```
java.lang.reflect.InvocationTargetException: null
	at sun.reflect.GeneratedMethodAccessor184.invoke(Unknown Source) ~[na:na]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_181]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[na:1.8.0_181]
	at de.zalando.zally.rule.RulesValidator.invoke(RulesValidator.kt:60) ~[main/:na]
...
Caused by: java.lang.NullPointerException: null
	at de.zalando.zally.rule.zalando.Use429HeaderForRateLimitRule.checkHeadersForRateLimiting(Use429HeaderForRateLimitRule.kt:22) ~[classes!/:na]

```

roxspring:
- Added test demonstrating problem to prevent recurrence.
- Simple fix with `.orEmpty()` calls.
- Refactored to use a new `Context.validateResponses()` for future shared safety.

Closes #787

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/zalando/Use429HeaderForRateLimitRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/zalando/Use429HeaderForRateLimitRule.kt
@@ -5,6 +5,7 @@ import de.zalando.zally.rule.api.Context
 import de.zalando.zally.rule.api.Rule
 import de.zalando.zally.rule.api.Severity
 import de.zalando.zally.rule.api.Violation
+import io.swagger.v3.oas.models.responses.ApiResponse
 
 @Rule(
     ruleSet = ZalandoRuleSet::class,
@@ -19,11 +20,14 @@ class Use429HeaderForRateLimitRule {
 
     @Check(severity = Severity.MUST)
     fun checkHeadersForRateLimiting(context: Context): List<Violation> =
-        context.api.paths.orEmpty().values.flatMap { it.readOperations().flatMap { it.responses.orEmpty().entries } }
-            .filter { (code, _) -> "429" == code }.map { it.value }
-            .filterNot { containsRateLimitHeader(it.headers.orEmpty().keys) }
-            .map { context.violation(description, it) }
+        context.validateResponses(responseFilter = this::violatingResponse) { (_, response) ->
+            context.violations(description, response)
+        }
 
-    private fun containsRateLimitHeader(headers: Collection<String>): Boolean =
-        "Retry-After" in headers || headers.containsAll(xRateLimitHeaders)
+    private fun violatingResponse(entry: Map.Entry<String, ApiResponse>): Boolean {
+        val headers = entry.value.headers.orEmpty().keys
+        return "429" == entry.key &&
+            "Retry-After" !in headers &&
+            !headers.containsAll(xRateLimitHeaders)
+    }
 }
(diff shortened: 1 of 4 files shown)


server/src/test/java/de/zalando/zally/rule/zalando/Use429HeaderForRateLimitRuleTest.kt
```diff
@@ -45,4 +49,21 @@ class Use429HeaderForRateLimitRuleTest {
 
         assertThat(violations).isEmpty()
     }
+
+    @Test
+    fun `checkHeadersForRateLimiting avoid bug 787 NPE on missing response`() {
```
tkrop: It looks strange to have this check here, but not in other rules that may/will fail too. We should have general test patterns for NPEs in OpenAPI schemas which should be checked on every rule. @roxspring are you interested in providing a solution?

server/src/main/java/de/zalando/zally/rule/zalando/Use429HeaderForRateLimitRule.kt
```diff
@@ -19,11 +20,14 @@ class Use429HeaderForRateLimitRule {
 
     @Check(severity = Severity.MUST)
     fun checkHeadersForRateLimiting(context: Context): List<Violation> =
-        context.api.paths.values.flatMap { it.readOperations().flatMap { it.responses.entries } }
-            .filter { (code, _) -> "429" == code }.map { it.value }
-            .filterNot { containsRateLimitHeader(it.headers.orEmpty().keys) }
-            .map { context.violation(description, it) }
+        context.validateResponses(responseFilter = this::violatingResponse) { (_, response) ->
```
ddufourboivin-zalando: I do not think the ``this::`` is required here.

ddufourboivin-zalando:
👍

maxim-tschumak:
:+1:

maxim-tschumak:
:+1:

tkrop:
:+1:

maxim-tschumak:
:+1:

server/src/test/java/de/zalando/zally/rule/zalando/Use429HeaderForRateLimitRuleTest.kt
roxspring: Test added here purely because I was working TDD style. Happy to move it somewhere more specific - presumably a test for DefaultContext.

Do you have a more specific suggestion? A niaive approach to generally testing the NPEs would be to just manually add a whole bunch of NPE focused unit tests - which doesn't sound much fun and is likely to still have gaps unless we hit 100% path coverage.

Kotlin generally does a nice job of avoiding NPEs at compile time, but this is undermined here by using Java object models (Swagger, OpenAPI) that are assumed to be non-null responses. Is there any mechanism to tell Kotlin / Idea to assume these are nullable responses instead? (suspect @ddufourboivin-zalando might know).

PR #829, opened 2018-09-07
feat(server): migrate nested paths may be root paths rule to use context object to support OpenAPI 3

maxim-tschumak:
Related to #714

### DIFF
--- a/server/src/test/java/de/zalando/zally/rule/zalando/NestedPathsMayBeRootPathsRuleTest.kt
+++ b/server/src/test/java/de/zalando/zally/rule/zalando/NestedPathsMayBeRootPathsRuleTest.kt
@@ -15,7 +15,7 @@ class NestedPathsMayBeRootPathsRuleTest {
         val spec = """
             openapi: 3.0.1
             paths:
-              "/countries/{country-id}/cities/{city-id}": {}
+              "/countries/{country-id}/populated/cities/{city-id}": {}
         """.trimIndent()
         val context = getOpenApiContextFromContent(spec)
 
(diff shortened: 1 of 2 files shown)


server/src/test/java/de/zalando/zally/rule/zalando/NestedPathsMayBeRootPathsRuleTest.kt
```diff
@@ -1,23 +1,44 @@
...
     @Test
-    fun avoidLinkHeadersValidYaml() {
-        val swagger = getFixture("api_spa.yaml")
-        assertThat(rule.validate(swagger)).isNull()
+    fun `checkNestedPaths should return no violations if there are no paths containing nested sub resources`() {
+        @Language("YAML")
+        val spec = """
+            openapi: 3.0.1
+            paths:
+              pets/: {}
```
tkrop: Missing check for `/pets/dogs`. This is a legal pattern and should be interpreted as top level resource.

server/src/test/java/de/zalando/zally/rule/zalando/NestedPathsMayBeRootPathsRuleTest.kt
```diff
@@ -1,23 +1,44 @@
...
-    fun avoidLinkHeadersValidJson() {
-        val swagger = getFixture("api_spp.json")
-        val result = rule.validate(swagger)!!
-        assertThat(result.paths).hasSameElementsAs(listOf("/products/{product_id}/updates/{update_id}"))
+    fun `checkNestedPaths should return violations for paths containing nested sub resources`() {
+        @Language("YAML")
+        val spec = """
+            openapi: 3.0.1
+            paths:
+              "/countries/{country-id}/cities/{city-id}": {}
```
tkrop: We should also check for something like `/countries/{country-id}/populated/cities/{city-id}`.

maxim-tschumak:
@tkrop @ddufourboivin-zalando please have another look

ddufourboivin-zalando:
👍

maxim-tschumak:
:+1:

PR #830, opened 2018-09-07
feat(server): migrate no version in uri rul to use context object to support OpenAPI 3

maxim-tschumak:
Related to #714


server/src/main/java/de/zalando/zally/rule/zalando/NoVersionInUriRule.kt
```diff
@@ -1,24 +1,24 @@
...
+    private val description = "Server URL contains version number"
 
     @Check(severity = Severity.MUST)
-    fun validate(swagger: Swagger): Violation? {
-        val hasVersion = swagger.basePath != null && PatternUtil.hasVersionInUrl(swagger.basePath)
-        return if (hasVersion) Violation(description, emptyList()) else null
-    }
+    fun checkServerURLs(context: Context): List<Violation> =
+        context.api.servers.orEmpty()
+            .filter { PatternUtil.hasVersionInUrl(it.url) }
```
tkrop: Shouldn't this be moved to this rule? Besides, what is with about patterns: `https://inter.net/api/1.0`, `https://inter.net/api/xyz-1.0`. This are likely also versioned resources paths.

server/src/test/java/de/zalando/zally/rule/zalando/NoVersionInUriRuleTest.kt
```diff
@@ -1,51 +1,43 @@
...
+              - url: "https://inter.net/api/v1.0"
+        """.trimIndent()
+        val context = getOpenApiContextFromContent(spec)
+
+        val violations = rule.checkServerURLs(context)
+
+        assertThat(violations).isNotEmpty
+        assertThat(violations[0].description).contains("Server URL contains version number")
+        assertThat(violations[0].pointer.toString()).isEqualTo("/servers/0")
     }
```
ddufourboivin-zalando: I'd prefer stricter tests for those, as in ensuring that the list contains exactly (but in any order) "this" (or "those") violations.

In this case, if the check all of a sudden returns 3 violations instead of 1, the test still passes although it is not what we are expecting.

server/src/main/java/de/zalando/zally/rule/zalando/NoVersionInUriRule.kt
tkrop: I think, for the second path our pattern may not be sufficient. We actually need to check each resource description for numbers.

maxim-tschumak:
@ddufourboivin-zalando @tkrop please re-review - I've incorporated the proposed improvements

maxim-tschumak:
@tkrop @ddufourboivin-zalando I've incorporated your proposals, please re-review

tkrop:
:+1:

maxim-tschumak:
:+1:

Issue #832, opened 2018-09-10
Implement generic test pattern to catch NPEs, etc in all rules.


tkrop:
While moving to OpenAPI 3.0 and migrating to Kotlin we had repeated problems with NPEs and SOEs in rules. May be we need a generic test pattern to detect some generally happening programmer mistakes for incomplete OpenAPI specifications. Such a task will pay off the more rules Zally provides.

roxspring:
Anybody know if there's a way to persuade Kotlin / Idea that e.g. `OpenAPI.paths` should have a signature of `Map<String, PathItem>?` rather than the `Map<String, PathItem>` without modifying the upstream Java sources? Feels like that would go a long way to addressing the NullPointerException problem.

Don't have any great suggestion for addressing StackOverflowExceptions (presumably what's meant by SOE above).

tkrop:
We have fixed another generic issue. While doing so we discovered a minimal strategy to solve this issue (see [my comment](https://github.com/zalando/zally/pull/835/files#r217303561)). The basic concept is to create a minimal API specification that covers all important elements and step by replace leave values against `null` and remove leaves. A parameterized test will with all resulting APIs allows discovery all potential NPE reason.

roxspring:
Are there thoughts to do anything more with this issue or should we just close it?

maxim-tschumak:
The idea @tkrop has mentioned is not implemented yet (a test which executes all rules against API specs with various null values).

roxspring:
As of #867 I think I've implemented @tkrop's idea (which I kept overlooking) so suspect this is closable now.

PR #835, opened 2018-09-12
Fixing StackoverFlow, Adding tests for NPE and support for recursive specifications

### LINKED ISSUE #826: Checking recursive APIs schemas result in StackOverflowException
Example: https://github.bus.zalan.do/setanta/api-reviews/pull/1/files

maxim-tschumak:
Closes #826

This PR includes:
- Fix for the `StackOverflowError` in one of the rule checks
- A general test for all checks to prevent further `StackOverflowError`s
- A general test for all checks to check whether they can cope with empty specifications

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/zalando/MediaTypesRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/zalando/MediaTypesRule.kt
@@ -24,8 +24,8 @@ class MediaTypesRule {
     @Check(severity = Severity.SHOULD)
     fun validate(context: Context): List<Violation> =
         context.validateOperations { (_, operation) ->
-            val consumedMediaTypes = operation.requestBody?.content?.entries.orEmpty()
-            val producedMediaTypes = operation.responses?.values.orEmpty()
+            val consumedMediaTypes = operation.requestBody?.content?.entries.orEmpty().toList()
+            val producedMediaTypes = operation.responses?.values.orEmpty().toList()
                 .flatMap { it.content?.entries.orEmpty() }
             (consumedMediaTypes + producedMediaTypes)
                 .filter { isViolatingMediaType(it.key) }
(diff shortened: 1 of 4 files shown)


server/src/main/java/de/zalando/zally/rule/zalando/MediaTypesRule.kt
```diff
@@ -24,8 +24,8 @@ class MediaTypesRule {
     @Check(severity = Severity.SHOULD)
     fun validate(context: Context): List<Violation> =
         context.validateOperations { (_, operation) ->
-            val consumedMediaTypes = operation.requestBody?.content?.entries.orEmpty()
-            val producedMediaTypes = operation.responses?.values.orEmpty()
+            val consumedMediaTypes = operation.requestBody?.content?.entries.orEmpty().toList()
+            val producedMediaTypes = operation.responses?.values.orEmpty().toList()
```
ddufourboivin-zalando: Just curious about why you need to transform them to `List`?

maxim-tschumak: > Oh ... actually ... is THAT the fix? By using an explicit list, elements do not end up in a Set or anything Hash-based et voilà! Problem gelöst! ...?

Yes, this was the problem! We should always avoid using `.toString` and `.hashCode` on the elements of the `OpenAPI` object - it will fail for recursive specifications.

maxim-tschumak:
:+1:

server/src/test/java/de/zalando/zally/apireview/ApiViolationsControllerTest.kt
```diff
@@ -26,10 +29,71 @@ class ApiViolationsControllerTest {
...
+            .andExpect(status().isOk)
+    }
+
+    /**
+     * This test runs all rules against an empty OpenAPI object. It checks for possible NPE in the check implementation
+     * when traversing the OpenAPI object tree. By having this test in place we don't have to write such a test for each
+     * rule.
+     */
+    @Test
+    fun `all rules must be able to cope with empty (minimal) api specifications`() {
```
tkrop: Here is the place where we should be more generic and generate all kinds of incomplete specifications with implicit and explicit `null` values. The problem is how this generator should look like ... the minimal approach I would choose is providing a basic template and successively replacing values against `null` and removing leaves. until the template is empty.

tkrop: Should be fixed by #832.

PR #838, opened 2018-09-13
837 minor rule fixes after migration

### LINKED ISSUE #837: Some bugs after migrating to the context object (#714)
### Things to fix

- [x] Whitelist of standard response codes must include well understood and specified response codes
- [ ] ~~Check for specified OAuth2 scopes doesn't recognize the specified scopes correctly~~
- [x] Default response Problem JSON check should check the specification under `/paths/{path}/get/responses/default/content/application/problem+json/`

maxim-tschumak:
Fixes #837


tkrop:
:+1:

maxim-tschumak:
:+1:

PR #841, opened 2018-09-14
fix(server): hyphenated pascal case regex (#839)

### LINKED ISSUE #839: PascalCaseHttpHeadersRule is to strict
and e.g. doesn't allow `ETag` header which is a valid header name.

maxim-tschumak:
Fixes #839

### DIFF
--- a/server/src/test/java/de/zalando/zally/util/PatternUtilTest.kt
+++ b/server/src/test/java/de/zalando/zally/util/PatternUtilTest.kt
@@ -58,6 +58,7 @@ class PatternUtilTest {
     fun checkIsHyphenatedPascalCase() {
         assertTrue(isHyphenatedPascalCase("Test-Case"))
         assertTrue(isHyphenatedPascalCase("X-Flow-Id"))
+        assertTrue(isHyphenatedPascalCase("ETag"))
         assertFalse(isHyphenatedPascalCase("test-Case"))
         assertFalse(isHyphenatedPascalCase("TestCase"))
         assertFalse(isHyphenatedPascalCase("testCase"))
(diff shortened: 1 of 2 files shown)


server/src/test/java/de/zalando/zally/util/PatternUtilTest.kt
```diff
@@ -58,6 +58,7 @@ class PatternUtilTest {
     fun checkIsHyphenatedPascalCase() {
         assertTrue(isHyphenatedPascalCase("Test-Case"))
         assertTrue(isHyphenatedPascalCase("X-Flow-Id"))
+        assertTrue(isHyphenatedPascalCase("ETag"))
```
roxspring: _Is_ this valid pascal-case? Definitions I see around only talk about capitalizing the first letter of each word. I know that Java classes use this approach to cope with acronyms and initialisms in class names but I was never clear whether this was part of pascal-case.

Had assumed that #839 was about adding "ETag" to a whitelist.

ddufourboivin-zalando:
👍

server/src/test/java/de/zalando/zally/util/PatternUtilTest.kt
maxim-tschumak: As far as I know, the pascal-case is not a standard. That's why I've been following my "intuition" and proposed this change (accepting acronyms at the beginning of each hyphenated word). I think a less strict regex is better than having a whitelist mechanism.

Are you ok with this approach?

roxspring: No real objections, but have been thinking about the inconsistency of these patterns lately and was wondering what definitions we were working to. Questions to my mind:

1. Shouldn't NonHyphenatedPascalCase be allowed to have uppercase acronyms at the start of each word too?
2. What about the camelCase variations - should _they_ allow uppercase acronyms (yes, but only in 2nd+ word??)
3. Several of these case patterns block numbers entirely but several allow them; seems unlikely to me that this is right!

No reason to block this PR - I probably ought to raise this in a separate issue.

maxim-tschumak: Yes, let's discuss this in a separate issue

maxim-tschumak:
:+1:

server/src/test/java/de/zalando/zally/util/PatternUtilTest.kt
roxspring: Raised #842

PR #843, opened 2018-09-17
WIP POC feat(server): Rule configuration from static resource #560

roxspring:
Hi folks,

Have been getting a little frustrated at the spring based tests pausing for 40s to find and load all rule classes so experimented with moving the scan to build time:

- `./gradlew ktlint check` was taking ~ 3m locally, spending 40s in `RulesDefinitionConfiguration` scanning for `@Rule` annotated classes.
- Built a static list of those rules and constructed them in `RulesManagerConfiguration` directly
  - Similar to `ServiceLoader` but with optional `Config` injection.
  - Same build was taking ~ 2m 20s locally ~ 20% faster overall
  - API & rules no longer _needed_ Spring dependency injection to load
  - Spring annotations still present for now.
  - Spring AOP still needed via Context at runtime so perhaps no major win.
- Added an annotation processor to automate the static list at build time
  - Placed in api module - may not be ideal but needed to be built in separate module from rules and it didn't seem worth a separate module
  - Appears to work in isolation
  - Currently breaks the compileJava task in gradle for reasons I haven't yet understood.

Would be interested to hear from others whether this feels like a healthy route forward or not, along with any other feedback you might have!


ddufourboivin-zalando:
First thought: looks like you are re-creating what _Spring_ does already (class scanning for services).

Although I am not a fan of _Spring_ myself, this software is using it and we should probably stick to it (even though startup performances are terrrrrrrrible).

roxspring:
Yeah - this really isn't that different from what Spring's doing, except that the cost is incurred once at build time rather than every start (including tests). It's also scanning a much smaller subset of classes - at build time only the freshly built classes are scanned, at runtime _all_ loaded classes are scanned since custom rulesets can't be assumed to be under `de.zalando.zally.rule`.

For additional context there have been a couple of suggestions (#122, #748) that removing the spring dependency for the rule library itself would be good for reusability - e.g. easier to integrate them into other tools - although clearly the suggestion hasn't been acted on so far.

All that said - I like the speed up but not the custom loading and don't need rules as an independent library - so am not certain this is a great change. Hence the request for feedback :) (thanks btw)

tkrop:
@roxspring I must agree with @ddufourboivin-zalando. This does not sound like a way we should go. If scanning for rules takes 40s, with Spring we should find out why and fix this issue by configuring it appropriately. We could in general optimize Spring to be faster, e.g. https://alexecollins.com/spring-boot-performance/.

roxspring:
Skimming over those Spring optimizations it seems that it's mostly about avoiding the scan at all. I'll have a play with that approach instead.

PR #853, opened 2018-09-22
More reliable `x-zally-ignore`

### LINKED ISSUE #799: x-zally-ignore doesn't work for some rules
**Example / how to reproduce**
Lint the following specification:
```
swagger: '2.0'

info:
  title: Some API
  version: '1.0.0'
  contact:
    name: Team X
    email: team@x.com
    url: https://team.x.com

parameters:
  SNSMessageTypeHeader:
    name: X-Amz-Sns-Message-Type
    in: header
    type: string
    x-extensible-enum:
      - Notification
      - SubscriptionConfirmation
    description: header describing the type of SNS message
    required: true

paths:
  '/sns':
    post:
      x-zally-ignore:
        - '183' # this endpoint must return custom header (SNS requirement)
      summary: Deliver SNS notification
      description: Endpoint for SNS to notify the service about e.g. new docker image.
      consumes:
        - 'text/plain'
      operationId: 'postSNSMessage'
      tags:
        - SNS
      parameters:
        - $ref: '#/parameters/SNSMessageTypeHeader'
      responses:
        401:
          description: Could not read sns message
          headers:
            WWW-Authenticate:
              type: string
              default: Basic realm="sns"
        default:
          description: a specific error occurred - see status code for more information.
          schema:
```

roxspring:
Given the following spec:

```
swagger: '2.0'
x-zally-ignore: [215, 218, 219]
info:
  title: Some API
  version: '1.0.0'
  contact:
    name: Team X
    email: team@x.com
    url: https://team.x.com
paths: {}
```

- Rule 218 generated a violation against `/info/description` - a node which doesn't exist in the parsed tree
- Previously this pointer mapped to a null node and therefore no `x-zally-ignore` extensions
- Now, when `/info/description` doesn't resolve to a parsed node, `/info` and if necessary `` are checked for `x-zally-ignore` extensions

Manually tested with @maxim-tschumak's case and rule 183 gets properly ignored.

Closes #799

### DIFF
--- a/server/src/test/java/de/zalando/zally/util/ast/ReverseAstTest.kt
+++ b/server/src/test/java/de/zalando/zally/util/ast/ReverseAstTest.kt
@@ -196,4 +196,28 @@ class ReverseAstTest {
                 .isSameAs(sharedParamPointer)
                 .hasToString("/components/parameters/SharedParam")
     }
+
+    @Test
+    fun `isIgnored even ignores unforeseen descendants of ignored nodes`() {
+        @Language("yaml")
+        val content = """
+            swagger: '2.0'
+            x-zally-ignore: [215, 218, 219]
+            info:
+              title: Some API
+              version: '1.0.0'
+              contact:
+                name: Team X
+                email: team@x.com
+                url: https://team.x.com
+            paths: {}
+            """.trimIndent()
+
+        val swagger = SwaggerParser().parse(content)
+        val ast = ReverseAst.fromObject(swagger).withExtensionMethodNames("getVendorExtensions").build()
+
+        assertThat(ast.isIgnored(JsonPointer.compile(""), "218")).isTrue()
+        assertThat(ast.isIgnored(JsonPointer.compile("/info"), "218")).isTrue()
+        assertThat(ast.isIgnored(JsonPointer.compile("/info/description"), "218")).isTrue()
+    }
 }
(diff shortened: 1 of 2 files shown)


server/src/test/java/de/zalando/zally/util/ast/ReverseAstTest.kt
```diff
@@ -196,4 +196,28 @@ class ReverseAstTest {
                 .isSameAs(sharedParamPointer)
                 .hasToString("/components/parameters/SharedParam")
     }
+
+    @Test
+    fun `isIgnored even ignores unforeseen descendants of ignored nodes`() {
+        @Language("yaml")
+        val content = """
+            swagger: '2.0'
+            x-zally-ignore: [215, 218, 219]
```
tkrop: Meh! I think we have to create a configurable list, which rules may be disabled at all. From Zalando perspective, not providing this information is an absolute *no-go*.

@roxspring isn't the information you are excluding in this test, the absolute minimum each API should provide from your perspective too?

tkrop:
:+1:

server/src/test/java/de/zalando/zally/util/ast/ReverseAstTest.kt
roxspring: Don't get too hung up on the example - it came from trying to isolate a fairly minimal example simultaneously demonstrating working and non-working `x-zally-ignore`. 

Not sure about a list of rules that cannot be ignored. I think most of our `x-zally-ignore` usage these days is to temporarily work around broken rules rather than evade their true intent. Blocking rules from being ignored means that such breakages need emergency fix or reconfig of the Zally server, whereas temporary ignorance allows a less urgent resolution.

Regarding those specific rules though, in-house we're not using any of them right now. It would likely be useful to enable 218 but our APIs versions are currently invalid since our APIs are validated before CI injects the right version. Haven't really seen a need for 215, possibly because our team is still small, and although it's fair to say that various APIs have different audiences for 219, we don't really have anything to use that information.

Issue #834, opened 2018-09-11
Migrate checks to operate on the context object to support OpenAPI 3 (follow-up)


maxim-tschumak:
Follow-up issue for #714.

After implementing the initial support for OpenAPI3 we have to adjust the check implementations. Instead of operating on the Swagger object, the checks have to use the `Context` object. Doing this, the checks become universal - applicable against Swagger2 and OpenAPI3 specifications.

Most important rules have already been refactored in #714.

**Acceptance Criteria:**
1. The following checks are refactored to use the `Context` object
2. Minimal specifications are used in the tests

## List of `Rules` to be refactored
- [x] SecureAllEndpointsWithScopesRule (currently not used in production) ([PR](#825))
- [x] ExtractBasePathRule ([PR](#861))
- [x] HyphenateHttpHeadersRule (Removed, see [discussion] (https://github.com/zalando/zally/issues/834#issuecomment-422937952)) ([PR](#607))
- [x] NoProtocolInHostRule ([PR](#856))
- [x] NoUnusedDefinitionsRule ([PR](#872))

vadeg:
Rule {{HyphenateHttpHeadersRule}} linked to rule 131 in API Guidelines but there is no such rule. Should we remove this rule?

maxim-tschumak:
@vadeg You are right, we should remove the check. Would you take over or should I do this?

vadeg:
@maxim-tschumak I will take over in scope of PR for this ticket.

maxim-tschumak:
@vadeg great, thanks!

roxspring:
`AvoidXZallyIgnoreRule` operates on `JsonNode` not `Swagger` so I think can be considered done.

roxspring:
Thanks for all the reviews and merges everyone!

They should mean that we can tick these off now:
- `ExtractBasePathRule`
- `HyphenateHttpHeadersRule`
- `NoProtocolInHostRule`

And hopefully #872 will tick `NoUnusedDefinitionsRule` and allow us to close the issue.

maxim-tschumak:
Thanks for your contributions! And sorry that we cannot always respond in one-two days.
Maybe we should set up another communication channel. Something like Slack.

What do you think? @roxspring @tkrop @ddufourboivin-zalando

roxspring:
You're welcome! Instant responses are not necessary, I'm comfortable with the usual response time of a few working days.

[...]

PR #856, opened 2018-09-26
feat(server): NoProtocolInHostRule operates on Context

roxspring:
- Rule updated to operate on `Context` and produce violations with a pointer
- Continues to block `Swagger` hosts with protocols since ["This MUST be the host only and does not include the scheme nor sub-paths"](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#swagger-object)
- Allows equivalent  `OpenAPI` server urls with protocols since the spec [clearly expects a full URL](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#server-object)

Relates to #834

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/zally/NoProtocolInHostRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/zally/NoProtocolInHostRule.kt
@@ -1,11 +1,22 @@
 package de.zalando.zally.rule.zally
 
+import com.fasterxml.jackson.core.JsonPointer
 import de.zalando.zally.rule.api.Check
+import de.zalando.zally.rule.api.Context
 import de.zalando.zally.rule.api.Rule
 import de.zalando.zally.rule.api.Severity
 import de.zalando.zally.rule.api.Violation
 import io.swagger.models.Swagger
 
+/**
+ * Validates that [Swagger.host] does not contain a url scheme as required by
+ * [OpenAPI 2.0](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#swagger-object).
+ *
+ * > The host (name or ip) serving the API. This MUST be the host only and does not include the scheme nor sub-paths.
+ *
+ * Note that this rule operates on a [Context] but specifically targets [Swagger.host] since the OpenAPI 3.0.0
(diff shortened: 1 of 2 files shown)


server/src/test/java/de/zalando/zally/rule/zally/NoProtocolInHostRuleTest.kt
```diff
@@ -1,54 +1,99 @@
...
+            .descriptionsEqualTo("'https://google.com' contains protocol information which should be listed separately as schemes")
+            .pointersEqualTo("/host")
     }
 
     @Test
-    fun negativeCaseHttps() {
-        val swagger = Swagger().apply { host = "https://google.com" }
-        val res = rule.validate(swagger)
-        assertThat(res?.copy(description = "")).isEqualTo(expectedViolation)
+    fun `validate swagger with SPP json returns no violations`() {
```
maxim-tschumak: Let's remove `validate swagger with SPP json returns no violations` and `validate swagger with SPA yaml returns no violations` tests. I don't see much benefit in them and think we should avoid having full-size API specifications as test fixtures.

server/src/main/java/de/zalando/zally/rule/zally/NoProtocolInHostRule.kt
```diff
@@ -1,25 +1,38 @@
...
-    private val desc = "Information about protocol should be placed in schema. Current host value '%s' violates this rule"
 
     @Check(severity = Severity.MUST)
-    fun validate(swagger: Swagger): Violation? {
-        val host = swagger.host.orEmpty()
-        return if ("://" in host)
-            Violation(desc.format(host), emptyList())
-        else null
+    fun validate(context: Context): List<Violation> {
+        val host = context.swagger?.host.orEmpty()
```
maxim-tschumak: The fact that only `Swagger` specifications a checked is a bit hidden. Should we use `Context#isOpenAPI3()` to make it more explicit?

maxim-tschumak:
:+1:

ddufourboivin-zalando:
👍

ddufourboivin-zalando:
Thanks for that, @roxspring !

Issue #858, opened 2018-09-26
Remove deprecated paths field from Violation


roxspring:
The `Violation` class currently allows reporting in either unstructured paths or strongly typed json pointers. Once the transition to `Context` based rules is complete in #834 then there are a few more cases that need addressing to tidy up:

- [x] `AvoidXZallyIgnoreRule` pull request #859 
- [x] `UseProblemJsonRule` pull request #860 
- [x] `ExtractBasePathRule` pull request #861
- [x] `HyphenatedHttpHeadersRule` likely addressed via #834
- [x] `NoUnusedDefinitionsRule` likely addressed via #834
- [ ] `RulesValidator`
- [ ] Remove the fields and unnecessary constructors from `Violation`

PR #859, opened 2018-09-26
feat(server): AvoidXZallyIgnoreRule returns Violations with JsonPointer

roxspring:
- Converted to return multiple violations with pointers rather than paths
- Added support for spotting `x-zally-ignore` descendant from an array node 
- Removed "empty swagger" test in favour of recently added shared cross-rule version

Working towards #858


ddufourboivin-zalando:
👍

maxim-tschumak:
:+1:

PR #865, opened 2018-10-01
feat(server): Default configuration enables all rules #864

### LINKED ISSUE #864: Default application*.yml files should not ignore any rules
#835 added global tests for empty definitions and recursive structures so that we don't need to implement at unit test level. This seems great except I [just realised](https://github.com/zalando/zally/pull/861#discussion_r220891739) that they're not run for rules that we ignore by default!

Currently we're ignoring 3 rules, one of which no longer exists:

- H001 - ExtractBasePathRule
- S005 - NoUnusedDefinitionsRule
- S010 - No such rule!!

I propose that we remove these entries from application.yml and application-test.yml.

Objections?

roxspring:
- All rules enabled by default
- Added comment docs to the configuration file to help people configure it in future

Closes #864

### DIFF
--- a/server/src/main/resources/application-dev.yml
+++ b/server/src/main/resources/application-dev.yml
@@ -27,7 +27,6 @@ management:
     port: 8000 # listening on the same port (`server` and `management.server`)
 
 zally:
-  ignoreRules: S010, H001
   cli:
     releasesPage: https://github.com/zalando/zally/releases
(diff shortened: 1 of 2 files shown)


ddufourboivin-zalando:
👍

maxim-tschumak:
:+1:

PR #867, opened 2018-10-02
NullPointerExceptionTest

roxspring:
- Added `NullPointerExceptionTest` running purposefully null ridden specs against all rules
  - Based on some full-featured specs, progressively pruned
- Fixed various NullPointerExceptions

#832

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
+++ b/server/src/main/java/de/zalando/zally/rule/DefaultContext.kt
@@ -45,8 +45,8 @@ class DefaultContext(
      * @return a list of Violations and/or nulls where no violations are necessary
      */
     override fun validatePaths(
-        pathFilter: (Map.Entry<String, PathItem>) -> Boolean,
-        action: (Map.Entry<String, PathItem>) -> List<Violation?>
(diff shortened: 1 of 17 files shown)


ddufourboivin-zalando:
Might be a rhetorical question ... but was there no other way to fix that than making that much stuff nullable...?

roxspring:
> Might be a rhetorical question ... but was there no other way to fix that than making that much stuff nullable...?

Arguably it was already nullable, Kotlin just didn't realise it.

Clearly there is an alternate approach where we post-process the model and replace nulls with "empty values", either by handcoding a crawl of the model or by enhancing/copying the `ReverseAst` to do so automatically. For many cases an "empty value" object exists which could be used e.g. `PathItem`, `ApiResponses` but I'm not sure that always makes sense - e.g. is it acceptable to replace null Strings with ""? What about null Booleans?

ddufourboivin-zalando:
Checking for _empty value_ or for _null_ both require an extra effort. At that point, I am OK with using nullability (which is nicely supported in Kotlin). Thanks for the explanation.

ddufourboivin-zalando:
👍

maxim-tschumak:
:+1:

PR #868, opened 2018-10-04
Flexible CaseCheckerRule

### LINKED ISSUE #813: Flexible Case Checking Rules
There are a number of rules which check the case of various strings:

- `SnakeCaseInPropNameRule`
- `SnakeCaseForQueryParamsRule`
- `PascalCaseHttpHeadersRule`
- `KebabCaseInPathSegmentsRule`

For our installation, we prefer camelCase for properties and query parameters, and I'm tempted to create some additional rules to check the case of path (and form, body, cookie?) parameters and shared definitions. Rather than maintain separate in-house rules I propose that I/we:

- Rename the rules into a consistent `CheckCaseOf*Rule` pattern
- Generalize the above rules such that the required case is parameterized
  - Implies that rule titles might become less specific "Use snake_case (never camelCase) for Query Parameters" would need to become "Query Parameters Use Correct Case"
- Also generalise the shared list of cases & regexes so that adding new ones becomes a non-code change

Relevant configuration would become something like:

```
Cases {
  camelCase: ^[a-z]+(?:[A-Z][a-z]+)*$
  snake_case: ^[a-z0-9]+(?:_[a-z0-9]+)*$
  kebab-case: ^[a-z]+(?:-[a-z]+)*$
}

CheckCaseOfPropertyNamesRule {
  case: snake_case
}

CheckCaseOfQueryParametersRule {
  case: kebab-case
}

CheckCaseOfHeaderParametersRule {
  case: SCREAMING_SNAKE_CASE    <- config time error!!
}
```

The case checking and error message generation should be done in a shared utility class so that we produce helpful error messages along the lines of "Query Parameters should be kebab-case but 'nameOfThing' appears to be camelCase".

It's tempting to suggest that we implement each of these as separate checks within a single rule but the existing ones are each tied to different guidelines so that would require us to rewrite those guidelines into one or revise our rule vs check model.

Objections?

### LINKED ISSUE #842: Inconsistent Case Checking
As noted in #841 I've been thinking about the inconsistency of the case checking patterns lately. Questions to my mind:

1. Shouldn't NonHyphenatedPascalCase be allowed to have uppercase acronyms at the start of each word too?
2. What about the camelCase variations - should _they_ allow uppercase acronyms (yes, but only in 2nd+ word??)
3. Several of these case patterns block numbers entirely but several allow them; seems unlikely to me that this is right!
4. Is it right that Hyphenated-Pascal-Case will now accept anything in TRAIN-CASE (as [claimed in Wikipedia](https://en.wikipedia.org/wiki/Letter_case#Special_case_styles))

roxspring:
- Added flexible `CaseCheckerRule` with configurable pattern and whitelists #813
  - Property names
  - Path segments
  - Path parameters
  - Query parmeters
  - Header names
- Updated existing rules
  - `SnakeCaseInPropNameRule` uses shared `CaseChecker` utility
  - `SnakeCaseForQueryParamsRule` uses shared `CaseChecker` utility
  - `PascalCaseHttpHeadersRule` uses shared `CaseChecker` utility
  - `KebabCaseInPathSegmentsRule` uses shared `CaseChecker` utility
  - `HyphenateHttpHeadersRule` no longer needed #834
- Harmonised case checking patterns for improved consistency #842
  - Words can contain numbers, but not at the start of the patterns
  - PascalCase and camelCase variants allow runs of uppercase rather than only singular

Closes #813
Closes #842

(Trying my luck with one big PR - got in the flow and a little carried away - can rework as smaller PRs if needed!)


server/build.gradle
```diff
@@ -92,7 +92,8 @@ dependencies {
     compile("org.zalando:problem-spring-web:0.23.0")
     compile("org.zalando:twintip-spring-web:1.1.0")
     compile("org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version")
-    compile("com.typesafe:config:1.3.3")
+    compile("io.github.config4k:config4k:0.4.0")
```
ddufourboivin-zalando: What is `config4k` doing that `typesafe:config` was not?

roxspring: Extending typesafe with some [Kotlin friendliness](https://github.com/config4k/config4k#usage), and seemingly has [some endorsement](https://github.com/lightbend/config#kotlin-wrappers-for-the-java-library) from the typesafe project.

Specifically it allowed me to read this [relatively complex config](https://github.com/zalando/zally/blob/95470591bc34684da32530d6f21480cfa92b9a1f/server/src/main/resources/rules-config.conf#L1) into a data class model with a one liner:

```kotlin
fun load(config: Config): CaseChecker = config.extract("CaseChecker")
```

maxim-tschumak:
:+1:

server/rules.md
```diff
@@ -18,6 +18,17 @@ Information about protocol should be placed in schema and not as part of the hos
...
+
+Enforced that various terms match case requirements configured via
+CaseChecker section in rules-config.conf.
+
+Supports:
+
+- schema property names
+- query parameter names
+- path parameter names
+
```
ddufourboivin-zalando: Is that supposed to replace the `131` that you removed some lines below?

ddufourboivin-zalando: I see a configuration `headerNames` in the `.conf` file, so I guess it also covers headers. Right?

server/rules.md
```diff
@@ -29,7 +40,3 @@ If all paths start with the same prefix then it would be cleaner to extract that
 ## H002: Avoid `x-zally-ignore`
 
 The `x-zally-ignore` extension should be used sparingly for temporary exceptional circumstances. Use encourages deviation from agreed standards. For longer term solutions please discuss disabling or adjusting the rule with your team.
-
-## 131: Use Hyphenated HTTP Headers
```
ddufourboivin-zalando: 1. This rule is actually # `132`

2. Why did you remove this block completely?

tkrop: This should be not removed but changed to `## 132: Use Hyphenated HTTP Headers`. See https://github.com/zalando/zally/pull/868/files#diff-8e75cd8b486efb2eda865ffd3c7fa0e1R17

roxspring: This file is documentation for the `de.zalando.zally.rule.zally` package of rules and the section was removed as the rule itself was removed. Rules in the `de.zalando.zally.rule.zalando` package are already [well documented](https://opensource.zalando.com/restful-api-guidelines/) and shouldn't be mentioned here.

Perhaps it'd all be clearer if we renamed `de.zalando.zally.rule.zally` to `de.zalando.zally.rule.extras` or `de.zalando.zally.rule.extras`  or similar so that there's clearer distinction between "Zally Rules implementing Zalando Guidelines" from "Other Zally rules".

tkrop: Robe, as far as I know the rule numbers should match our guideline rule numbers. I think `131` should have originally been `132`, which matches the description correctly (see https://opensource.zalando.com/restful-api-guidelines/#132). I'm not absolutely sure what you are referring to with Zally rules, but as far as I remember the discussion with @maxim-tschumak these should be numbered with `M0xx`.

tkrop: @roxspring besides, as you have not removed the check for rule `132` we should still document it here.

roxspring: Take a look at the prior https://github.com/zalando/zally/blob/master/server/rules.md in context.

- Rule 132 
  - Is in ZalandoRuleSet and so it's documentation belongs elsewhere.
  - For some reason existed in addition to 131 while essentially duplicating it.
- Rule 131
  - Is in ZallyRuleSet and so it's documentation belongs here.
  - Should have been named in the `M0xx` pattern (or renamed if/when moved from ZalandoRuleSet).
  - But is superseded by `CaseCheckerRule` in this change hence removal.

server/rules.md
roxspring: Yes, covers the following currently and could be easily extended to cover more:
  - Property names
  - Path segments
  - Path parameters
  - Query parmeters
  - Header names

server/rules.md
tkrop: @roxspring Okay, thanks for clearing up the picture.

tkrop:
:+1:

maxim-tschumak:
:+1:

PR #872, opened 2018-10-18
feat(server): NoUnusedDefinitionsRule reimplemented checks JsonNode

roxspring:
- Repeatedly failed at upgrading this rule to use `Context` since both the `Swagger` and `OpenAPI` models are already partially resolved by the time the rules have access to it.
  - Possibly we should consider turning off the resolver, as e.g. resolution moves path level parameters to operation level and presumably we spit out violations against the wrong location.
  - But that would also block external refs being internalized and checked which doesn't seem ideal.
- Instead reimplemented checking against `JsonNode` but returning `JsonPointer` violations

Relates to #834


roxspring:
Thanks @ddufourboivin-zalando. Dealing with the discriminators has been on my todo list for ages - we use them quite heavily. Somehow hadn't occurred to me that a PR to a PR was an option - but it worked nicely!

maxim-tschumak:
:+1:

tkrop:
:+1:

ddufourboivin-zalando:
👍

ddufourboivin-zalando:
@roxspring : When you are experimenting with your suggestions locally, I find it a better solution than to cherry-pick all modification as comments. It's also a nice way of politely proposing alternatives and let the initial PR author to accept them or not.

:-) Long live GitHub! :)

PR #875, opened 2018-10-18
feat(server): exclude violations and full reviews from the statistics endpoint

### LINKED ISSUE #873: Optimize review-statistics calculations
We have pretty much linting requests these days. This is great :), but the `/review-statistics` endpoint does not cope with the load. Querying last week statistics can take up to one minute. The main reason is the **size** of the review statistics which are being transmitted from database to server and then back to the client.

I propose to exclude `review` and `violations` from the [`review-statistics`](https://github.com/zalando/zally/blob/master/server/src/main/resources/api/zally-api.yaml#L253) API resource and only send aggregated numbers - the actual statistics.

**CAUTION:** This is a breaking change of Zally API.

@roxspring Do you use the `/review-statistics` endpoint?

maxim-tschumak:
Closes #873

### DIFF
--- a/server/src/main/java/de/zalando/zally/apireview/ApiReviewRepository.kt
+++ b/server/src/main/java/de/zalando/zally/apireview/ApiReviewRepository.kt
@@ -1,11 +1,47 @@
 package de.zalando.zally.apireview
 
+import de.zalando.zally.statistic.ReviewStatistics
+import org.springframework.data.jpa.repository.Query
 import org.springframework.data.repository.CrudRepository
+import org.springframework.data.repository.query.Param
 import java.time.LocalDate
 
 interface ApiReviewRepository : CrudRepository<ApiReview, Long> {
 
-    fun findByDayBetween(from: LocalDate, to: LocalDate): Collection<ApiReview>
+    @Query("""
+        SELECT new de.zalando.zally.statistic.ReviewStatistics(
+	        COUNT(r) AS totalReviews,
+	        COUNT(DISTINCT r.name) AS totalReviewDeduplicated,
+	        COALESCE(SUM(CASE WHEN r.isSuccessfulProcessed = 'True' THEN 1 ELSE 0 END),0) AS successfulReviews,
+	        COALESCE(SUM(r.numberOfEndpoints),0) AS numberOfEndpoints,
+	        COALESCE(SUM(r.mustViolations),0) AS mustViolations,
+	        COALESCE(SUM(r.shouldViolations),0) AS shouldViolations,
+	        COALESCE(SUM(r.mayViolations),0) AS mayViolations,
+	        COALESCE(SUM(r.hintViolations),0) AS hintViolations)
+        FROM de.zalando.zally.apireview.ApiReview r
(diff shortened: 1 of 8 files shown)


server/src/main/java/de/zalando/zally/apireview/ApiReviewRepository.kt
```diff
@@ -1,11 +1,47 @@
...
+        WHERE r.day >= :from AND r.day <= :to
+    """)
+    fun getReviewStatistics(
+        @Param("from") from: LocalDate,
+        @Param("to") to: LocalDate
+    ): ReviewStatistics
 
-    fun findByUserAgentAndDayBetween(userAgent: String, from: LocalDate, to: LocalDate): Collection<ApiReview>
+    @Query("""
+        SELECT new de.zalando.zally.statistic.ReviewStatistics(
```
tkrop: This and the before query look very similar. I suggest to express this by a single query used for both request. I would use `IN` for that and allow to provide a list of `userAgents` in addition.

maxim-tschumak: Good idea! I've managed to merge the HQL queries. But, it will not work with a collection of `userAgent`s - it's not possible to get all entries.

maxim-tschumak:
@tkrop please re-review and re-aprove

server/src/main/java/de/zalando/zally/apireview/ApiReviewRepository.kt
```diff
@@ -1,11 +1,29 @@
...
+	        COUNT(r) AS totalReviews,
+	        COUNT(DISTINCT r.name) AS totalReviewDeduplicated,
+	        COALESCE(SUM(CASE WHEN r.isSuccessfulProcessed = 'True' THEN 1 ELSE 0 END),0) AS successfulReviews,
+	        COALESCE(SUM(r.numberOfEndpoints),0) AS numberOfEndpoints,
+	        COALESCE(SUM(r.mustViolations),0) AS mustViolations,
+	        COALESCE(SUM(r.shouldViolations),0) AS shouldViolations,
+	        COALESCE(SUM(r.mayViolations),0) AS mayViolations,
+	        COALESCE(SUM(r.hintViolations),0) AS hintViolations)
+        FROM de.zalando.zally.apireview.ApiReview r
+        WHERE day >= :from AND day <= :to AND user_agent LIKE :userAgent
```
tkrop: I'm not sure whether it is wise to expose the PostgreSQL pattern matching language and whether this would be a use case. That's why I suggested the slightly more complex way to use `IN` in conjunction of checking with an empty array checking [`array_length(:userAgents, 0)`](https://www.postgresql.org/docs/current/static/functions-array.html#ARRAY-FUNCTIONS-TABLE).

tkrop: Two arguments I can accept :smile_cat:

tkrop:
:+1:

maxim-tschumak:
:+1:

PR #878, opened 2018-10-26
ZallyAssertion should be an object

ddufourboivin-zalando:
fix #878 

ZallyAssertion should be an `object`, not a `class` with only a `companion object`.


maxim-tschumak:
:+1:

ddufourboivin-zalando:
👍

PR #886, opened 2018-11-01
'recursiveness' to 'recursion'

### LINKED ISSUE #883: Replace `recursiveness` by `recursion`
```java
    /**
     * This test runs all rules against an API specification with recursive references. It prevents check developers
     * from StackOverflow exceptions during the check execution.
     *
     * General advice: Don't use `hashCode` or `toString` methods on the OpenAPI elements since they don't support
     * recursiveness. Keep in mind that these methods can be called implicitly (e.g. when putting objects in a Set).
     */
    @Test
    fun `all rules must be able to cope with recursive api specifications`() {
```

`recursiveness` is not a word.
The word you are looking for it ~~`recursivity`~~ `recursion`.

Perform a project wide search to make sure that mistake is not anywhere else.

roxspring:
docs(server): Changed doc comments to use the word `recursion` rather than `recursiveness`

Closes #883


maxim-tschumak:
:+1:

ddufourboivin-zalando:
👍

PR #885, opened 2018-10-31
feat(server): Removed unnecessary Violation.paths property

roxspring:
* Removed `Violation.paths` since it's no longer being used.
* `Violation.pointer` is no longer optional.

Closes #858

PR #892, opened 2018-11-07
feat(server): Add line numbers to Result when source was YAML

roxspring:
- Server re-parses YAML files using `SnakeYAML` and looks up line numbers from `Violation.pointer` locations
  - If line numbers for `/path/to/thing` can't be found then line numbers for `/path/to`, `/path` and `​` will be used instead
  - Columns are also available if needed.
- CLI ~~and UI~~ append "(line 4)" or "(lines 4-10)" to pointer display strings
- UI unchanged for now since:
  - I can't get the tests passing locally at the moment (even without changes)
  - Fighting the tests via CI wore me down
  - Appending "(lines 4-10)" was very unimaginative use of the line numbers and we should take time to do much much better.
- No JSON support
  - Would need to find JSON parser that preserves line numbers and navigate via JsonPointer
- Addressed the Codacy commenting complaints in https://github.com/zalando/zally/pull/892/commits/0d2120b6fa14306af22864dd465dacc06b18d151

Partially addresses #78 which is currently closed.

### DIFF
--- a/cli/zally/domain/violation.go
+++ b/cli/zally/domain/violation.go
@@ -1,6 +1,7 @@
 package domain
 
 import (
+	"fmt"
 	"strings"
 )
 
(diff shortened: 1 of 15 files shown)


roxspring:
btw I'm aware that appending "(lines 5-10)" is a very boring use of the information in the web ui but I'm hoping that somebody will build on this to improve later now that the line numbers are available...

- Perhaps allowing the user to click on a violation and have the editor scroll to the right location with the lines highlighted?
- Perhaps highlighting the lines in the editor based on the severity of violations they've triggered?
- Perhaps displaying the violations inline with editor somehow?

... in the meantime it appears I screwed up a rebase and need to fix the CI build.

roxspring:
- Addressed the Codacy commenting complaints in https://github.com/zalando/zally/pull/892/commits/0d2120b6fa14306af22864dd465dacc06b18d151
- Ignoring the complexity complaint for now because:
  - Hopefully #885 will bring it back under the limit shortly
  - Suspect that `RulesValidator` is due a big tidy up, most of which is far beyond the scope of this PR.

Is that reasonable @ddufourboivin-zalando?

ddufourboivin-zalando:
That's wonderful, @roxspring .

What does it not like about `RulesValidator.invoke`? Too long body?

ddufourboivin-zalando:
👍

maxim-tschumak:
:+1:

roxspring:
By my reckoning my change to `RulesValidator.invoke` added 1 parameter, 1 statement, 1 branch. I'm not sure whether it's one of those or a combination that tripped the limit for this check. #885 should drop 1 statement and 1 branch though, so I'm hopeful that the net result isn't too bad.

tkrop:
@roxspring sorry for letting this pending so long. Can you fix the conflicts? I hope we are than able to merge this.

PR #899, opened 2018-11-13
feat(server): Add line numbers to Result when source was YAML (replacement pr)

roxspring:
- Server re-parses YAML files using `SnakeYAML` and looks up line numbers from `Violation.pointer` locations
  - If line numbers for `/path/to/thing` can't be found then line numbers for `/path/to`, `/path` and `​` will be used instead
  - Columns are also available if needed.
- CLI ~~and UI~~ append "(line 4)" or "(lines 4-10)" to pointer display strings
- UI unchanged for now since:
  - I can't get the tests passing locally at the moment (even without changes)
  - Fighting the tests via CI wore me down
  - Appending "(lines 4-10)" was very unimaginative use of the line numbers and we should take time to do much much better.
- No JSON support
  - Would need to find JSON parser that preserves line numbers and navigate via JsonPointer
- Addressed the Codacy commenting complaints in https://github.com/zalando/zally/pull/892/commits/0d2120b6fa14306af22864dd465dacc06b18d151

Replaces #892 

Partially addresses #78 which is currently closed.


maxim-tschumak:
:+1:

tkrop:
@roxspring Sorry, we missed to merge this directly after the environment fix last week.

Issue #593, opened 2017-11-16
Permalinks to reviews


codekultur:
It would be great, if Zally could provide a permalink to a review. So I could send findings via mail, chat etc. . Currently I always have to explain to go to Zally and either paste the API definition in the editor or provide the Git Hub link. If I had a permalink, the receiver of the message could just go to Zally and see what I mean.

netme:
Really great idea, we already talked a lot about it during GH Integration project and had some thoughts on how to automate it.

ruiaraujo:
@netme We could do it stateless for URL based checks.

We just encode the url in the link.

ruiaraujo:
The ui could handle that easily.

vadeg:
I can suggest to use an `id` field from `ApiReview` class in permalink and add an endpoint like `GET /api-violations/<review-id>` which will return review results. In general, following steps should be done:
* Implement `/api-violations/<review-id>` endpoint which should return review results.
* Update `ApiDefinitionResponse` by adding `review-id` which can be used to build permalink in UI.

What do you think?

vadeg:
> suggest we use a randomly assigned UUID for this purpose

Would be nice, but not necessary. Numeric value as a string and then can be safely replaced with uuid if needed. In this case no migration needed to already existing reviews.

roxspring:
How do you expect existing reviews to be accessed? Users won't have the ids for them, numeric or uuid, and I've seen no requirement for listing reviews.

vadeg:
Yeah, you're right. I forgot that there is no way to access it right now :).

roxspring:
(whoops - misclick!)

maxim-tschumak:
implemented in #901

roxspring:
API was implemented but no UI.

Hope to open a UI pull request tomorrow.

PR #901, opened 2018-11-25
API to support permalinks

roxspring:
- Modified `POST /api-validations`
  - `api_review` table includes a UUID external identifier.
  - `rule_violation` table include full violation representation.
  - `ApiDefinitionResponse` now built from database content.
  - Response includes a `Location` header like most `201 Created` responses.
  - Response status still `200 OK` since at least the CLI requires that.
- New `GET /api-validations/{uuid}`
  - Responds `200 OK` with the same `ApiDefinitionResponse` built from the database content.
  - Alternatively responds `404 Not Found` when the specified review does not exist.
  - No authorization applied - if you know the UUID you can retrieve API.

Provides API for #593

### DIFF
--- a/server/src/main/java/de/zalando/zally/apireview/ApiReview.kt
+++ b/server/src/main/java/de/zalando/zally/apireview/ApiReview.kt
@@ -29,7 +29,6 @@ class ApiReview : Serializable {
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     var id: Long? = null
 
-    @Column(nullable = false)
     var externalId: UUID? = null
 
     var name: String? = null
--- a/server/src/main/java/de/zalando/zally/apireview/ApiReview.kt
+++ b/server/src/main/java/de/zalando/zally/apireview/ApiReview.kt
@@ -11,6 +11,7 @@ import java.time.Instant
 import java.time.LocalDate
 import java.time.OffsetDateTime
 import java.time.ZoneOffset
+import java.util.UUID
 import java.util.Objects
 import javax.persistence.CascadeType
 import javax.persistence.Column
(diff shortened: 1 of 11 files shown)


roxspring:
Codacy is happy. Travis was happy on a second attempt. But now @zalando-robot is upset and obviously, I can't see the details. Hopefully it's something environmental and the change itself can be reviewed anyway!

maxim-tschumak:
It was a temporary problem - I've retriggered the build. I'll try to review this PR today.

server/src/main/resources/api/zally-api.yaml
```diff
@@ -59,6 +59,12 @@ paths:
             application/json:
```
maxim-tschumak: Please bump the API version (->`2.1.0`).

server/src/main/java/de/zalando/zally/apireview/ApiReview.kt
```diff
@@ -28,6 +29,9 @@ class ApiReview : Serializable {
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     var id: Long? = null
 
+    @Column(nullable = false)
```
maxim-tschumak: What happens if we retrieve an entry from the database with a `null` value in it - would the mapping fail?

roxspring: Not sure. I was under the impression that these annotations were only actually used for schema generation which we're clearly not using here since we're building it via FlyWay instead. The `var externalId: UUID? = null` is clearly optional, which I assume means that the mapping should cope anyway.

I'll remove the annotation and dodge any risk.

maxim-tschumak:
:+1:

tkrop:
:+1:

PR #909, opened 2019-01-02
fix: Correct pluralization/tense of description

### LINKED ISSUE #908: Incorrect pluralization of pluralization complaint
> Array property names has to be pluralized

Should likely be "... property name has ..." or "... property names have ..."

roxspring:
Original wording has bad pluralization and/or tense:

> Array property names has to be pluralized

Replaced with the following:

> Array property name appears to be singular

Adopting less confident phrasing of `PluralizeResourceNamesRule` since programmatic checking of plurals in English is not entirely reliable.

Closes #908


maxim-tschumak:
@tkrop please have a look

tkrop:
:+1:

Issue #918, opened 2019-01-18
Test Failures building under Java 11


roxspring:
Attempting to build under Java 11 results in the test failures below which cause confusion to those attempting to build and use Zally.

We should do at least one of the following:
* Fail fast when attempting to build under unsupported Java versions
* Make the code build/run/test under Java 11
* Consider dropping support for Java 8

Note that Zally currently uses a Zalando docker image `registry.opensource.zalan.do/stups/openjdk:latest` with no obvious Java 10/11 equivalent but the expectation is that we could switch to a non-Zalando docker image if necessary.

```
de.zalando.zally.configuration.PathMatchersTest > regex matcher matches allows query parameters FAILED
org.mockito.exceptions.base.MockitoException at PathMatchersTest.kt:63
Caused by: java.lang.UnsupportedOperationException at PathMatchersTest.kt:63
Caused by: java.lang.IllegalStateException at PathMatchersTest.kt:63
Caused by: java.lang.NoSuchMethodException at PathMatchersTest.kt:63

de.zalando.zally.configuration.PathMatchersTest > regex matcher matches requests with and without trailing slash FAILED
org.mockito.exceptions.base.MockitoException at PathMatchersTest.kt:63
Caused by: java.lang.UnsupportedOperationException at PathMatchersTest.kt:63
Caused by: java.lang.IllegalStateException at PathMatchersTest.kt:63
Caused by: java.lang.NoSuchMethodException at PathMatchersTest.kt:63

de.zalando.zally.configuration.PathMatchersTest > antMatcher without trailing slash matches just requests without trailing slash FAILED
org.mockito.exceptions.base.MockitoException at PathMatchersTest.kt:63
Caused by: java.lang.UnsupportedOperationException at PathMatchersTest.kt:63
Caused by: java.lang.IllegalStateException at PathMatchersTest.kt:63
Caused by: java.lang.NoSuchMethodException at PathMatchersTest.kt:63

de.zalando.zally.configuration.PathMatchersTest > antMatcher with trailing slash matches just requests with trailing slash FAILED
org.mockito.exceptions.base.MockitoException at PathMatchersTest.kt:63
Caused by: java.lang.UnsupportedOperationException at PathMatchersTest.kt:63
Caused by: java.lang.IllegalStateException at PathMatchersTest.kt:63
Caused by: java.lang.NoSuchMethodException at PathMatchersTest.kt:63
```
Thanks to @nunogilrocha for the stack trace.

PR #920, opened 2019-01-23
Java 11 Support

roxspring:
Update libraries to support building and testing under Java 11

#918


roxspring:
Looks to me like the @codecov-io drop is a false-negative report - suspect the upgrade in jacoco has made a slight change to how lines of code are counted resulting in the minor drop.

maxim-tschumak:
:+1:

Issue #919, opened 2019-01-18
Local zally Windows installation doesn't work 


vivanov19:
Hello, 

I followed the README documents and related issues and almost installed zally locally on the Windows 10 machine. I am able to execute all gradlew tasks, but when I am executing the bootRun it stops after initiating the back-end server 0% as script indicated). Back-end (localhost:8000) for sure works since the command `zally --linter-service "http://localhost:8000" --format pretty lint no_must_violations.yaml` brings the meaningful results. I can also get the list of the rules from the browser executing http://localhost:8000/supported-rules as well as all 200+ tests executed successfully.  Attached is the complete output from my DOS shell. 

On the other hand, it never brings front-end running on 8080 port. I am guessing it might be some proxy setting between zally back-end and front-end need to be done, but not sure.

I also successfully mounted / build the web-ui component separately with zally server running in its current state. Again all build tasks are complete successfully with npm and / or yarn, but I can run neither dev nor production builds. Both brings similar exceptions ''NODE_ENV' is not recognized as an internal or external command,'.  DEV mode also brings some JS exception. Again, the whole output from another DOS shell is attached.

Please let me know if any other information needed.

[FullZallyDOSShellOutput.txt](https://github.com/zalando/zally/files/2773991/FullZallyDOSShellOutput.txt)

[WebUIZallyDOSShellOutput.txt](https://github.com/zalando/zally/files/2774017/WebUIZallyDOSShellOutput.txt)

roxspring:
From your first paragraph it's pretty clear that you have the server running successfully on port 8000, and the command line working too. The front-end on port 8080 come from the web-ui which you're having trouble with.

The command `NODE_ENV=development nodemon server.js` being run isn't going to work directly under Windows as 1) it's attempting to set an environment variable in the command line and 2) nodemon is unlikely to be on the path. You should be able to do the equivalent by running `set NODE_ENV=development` and then `node_modules\.bin\nodemon.cmd server.js`. That alone isn't enough to get the web-ui running for me, so I suspect more changes are needed to get up and running directly under Windows.

For what it's worth, I routinely build and run Zally under Windows 10 but using Windows Subsystem for Linux (Ubuntu 18.04 specifically) and can strongly recommend that approach. Alternatively, if you have Docker for Windows installed then you might want to try out https://github.com/zalando/zally/pull/917, although I haven't got round to testing it yet.

vivanov19:
Hi Rob,

[...]

PR #933, opened 2019-02-28
feat: build-and-run.sh performs incremental builds of server

roxspring:
- `build-and-run.sh` performs incremental builds of server rather than clean build every run
  - Gradle has great caching and can avoid lots of wasted time when nothing has changed in the server between runs.
  - It's not like we were forcing any similar "clean" functionality for the web-ui build anyway so this brings parity between the component builds.

Relates to comments in #919


build-and-run.sh
```diff
@@ -14,15 +14,15 @@ WEB_UI_DIR=${SCRIPT_DIR}/web-ui
 GITHUB_INTEGRATION_DIR=${SCRIPT_DIR}/github_integration
 
 # Build server
-cd ${SERVER_DIR} && ./gradlew clean build
+cd ${SERVER_DIR} && ./gradlew build
```
maxim-tschumak: we can even: `./gradlew build -x test`

roxspring: Could do. 

If nothing has changed in the server then a second `./gradlew build` skips the test anyway and runs in 15s for me - I was good with that speed up alone.

If we're happy for `build-and-run.sh` to never run the tests though then I'm happy to make that tweak too - I guess it would bring even more parity with the web-ui build since I don't think this script runs those tests either.

maxim-tschumak: Let's skip the tests. Fast response is more important here.

maxim-tschumak:
:+1:

tkrop:
:+1:

maxim-tschumak:
:+1:

PR #958, opened 2019-03-19
fix(server): run only Zalando checks per default (#957)

### LINKED ISSUE #957: Zally server should only run Zalando checks per default

maxim-tschumak:
Fixes #957 

@roxspring We always tried to be consistent with the Zalando Guidelines per default. At some point, you commented out the `ignoredRules` and we didn't see that. Now, we want to switch this back. It may impact your setup, so please check that you have your own configuration.

### DIFF
--- a/cli/zally/integration_test.go
+++ b/cli/zally/integration_test.go
@@ -82,7 +82,7 @@ func TestIntegrationWithNoMustViolations(t *testing.T) {
 	must, should, may, hint := countViolations(out)
 
 	assert.Zero(t, must)
-	assert.True(t, should > 0)
+	assert.True(t, should == 0)
 	assert.True(t, may > 0)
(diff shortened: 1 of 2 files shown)


zeitlinger:
:+1:

maxim-tschumak:
> We should have a mechanism (perhaps extended syntax for ignoreRules) to ignore an entire ruleset at a time rather than have to remember to add to this list every time a non-guideline rule is added.

Agree, we will do this after the Rule Set epic is completed.

Thanks for the fast feedback!

zeitlinger:
:+1:

maxim-tschumak:
:+1:

maxim-tschumak:
:+1:

zeitlinger:
:+1:

PR #965, opened 2019-03-29
Mapping simplification based on headers

vincent-fuchs:
While browsing the code, I noticed code could be slightly improved - using Spring's mapping annotations, we can define the mapping directly, avoiding a check later on.

### DIFF
--- a/github-integration/src/main/java/de/zalando/zally/integration/ValidationController.kt
+++ b/github-integration/src/main/java/de/zalando/zally/integration/ValidationController.kt
@@ -15,22 +15,15 @@ import javax.servlet.http.HttpServletRequest
 
 @RestController
 class ValidationController(private val validationService: ValidationService) {
-    private val PULL_REQUEST_EVENT_NAME = "pull_request"
 
     private val log = KotlinLogging.logger {}
 
-    @PostMapping("/github-webhook")
+    @PostMapping("/github-webhook", headers = arrayOf("X-GitHub-Event=pull_request"))
     @ResponseStatus(HttpStatus.ACCEPTED)
     fun validatePullRequest(@RequestBody payload: String,
-                            @RequestHeader(value = "X-GitHub-Event") eventType: String,
                             @RequestHeader(value = "X-Hub-Signature") signature: String) {
 
-        if (eventType != PULL_REQUEST_EVENT_NAME) {
-            log.info("Received unsupported event: {}", eventType)
-            return
-        }
-
-        log.info("Received webhook: {}", eventType)
+        log.info("Received pull request event on webhook")
 
         validationService.validatePullRequest(payload, signature)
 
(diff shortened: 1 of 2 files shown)


vincent-fuchs:
OK, I'll create an issue first...

github-integration/src/main/java/de/zalando/zally/integration/ValidationController.kt
```diff
@@ -15,22 +15,15 @@ import javax.servlet.http.HttpServletRequest
 
 @RestController
 class ValidationController(private val validationService: ValidationService) {
-    private val PULL_REQUEST_EVENT_NAME = "pull_request"
 
     private val log = KotlinLogging.logger {}
 
-    @PostMapping("/github-webhook")
+    @PostMapping("/github-webhook", headers = arrayOf("X-GitHub-Event=pull_request"))
```
roxspring: Wasn't aware you could require a header value like this - useful to know!

github-integration/src/test/java/de/zalando/zally/integration/ValidationControllerIntegrationTest.kt
```diff
@@ -62,12 +62,7 @@ class ValidationControllerIntegrationTest {
 
         val response = restTemplate.postForEntity("/github-webhook", HttpEntity(body, headers), String::class.java)
 
-        assertThat(response.statusCode, `is`(HttpStatus.ACCEPTED))
-
-        githubServer.mock.verifyThatRequest()
-                .havingMethodEqualTo("POST")
-                .havingPath(Matchers.containsString("/statuses/"))
-                .receivedNever()
+        assertThat(response.statusCode, `is`(HttpStatus.NOT_FOUND))
```
roxspring: Don't know enough about GitHub webhooks myself but this change in response behaviour rings alarm bells for me... suggest somebody with GitHub webhook experience confirms that a 404 isn't going to confuse matters.

vincent-fuchs: in my experience, this is fine. In webhook config at organization level, you can mention which types of messages (https://developer.github.com/v3/activity/events/types/) need to be sent to the webhook. You can also see in Github UI the history of all messages delivery, and their status.

Previously, I feel it was misleading : whatever message type is sent, we would get a 2xx response code, as if everything was fine (on client side) - even though nothing was processed. That's why I feel it's more accurate to return a 4xx in case we send a message of unexpected type : we notify the client his/her request was invalid.  

It's arguable, but we could say that in that case, we should return 400 (bad request) rather than 404. In that case, I am not sure we'll be able to use the light syntax I was proposing. We may have to revert to previous code, and simply change the response, from "return" to a specific exception eventually translated in 400 with custom message. 

Let me know your thoughts, I can rework the PR if required

maxim-tschumak:
:+1:

zeitlinger:
:+1:

github-integration/src/test/java/de/zalando/zally/integration/ValidationControllerIntegrationTest.kt
roxspring: 👍

PR #968, opened 2019-03-29
Use actively maintained ktlint plugin

### LINKED ISSUE #967: ktlint doesn't run on zally-rule-api subproject
ktlint isn't being run on the `zally-rule-api` subproject, or any future subprojects as it stands.

At least part of the reason is that we're using a limited homegrown ktlint task which I propose replacing with an actively maintained plugin: https://plugins.gradle.org/plugin/org.jlleitschuh.gradle.ktlint

Using the latest version of the plugin entails a minor bump to Gradle version, but I propose bumping all the way to 5.3.1 to become current again.

roxspring:
- Upgrade server and gitlab-integration to Gradle 5.3.1
- Switch to a supported ktlint plugin
- Run ktlintFormat

Closes #967


roxspring:
(rebased on master)

maxim-tschumak:
:+1:

zeitlinger:
:+1:

PR #970, opened 2019-03-29
Server & GitHub Integration use Gradle Kotlin DSL

### LINKED ISSUE #969: Switch Gradle scripts to use Kotlin DSL
- #968 upgrades us to Gradle >= 5.0 where Kotlin DSL is officially stable and supported. 
- Kotlin DSL enables better autocompletion etc within Intellij.
- We're already using Kotlin everywhere else.
- As a bonus the ktlint plugin in #968 checks (and can format) the build scripts too.

(I got carried away with the #968 branch and have the additional change ready to go for the server, haven't touched the github integration yet)

roxspring:
- Builds on top of #968 
- Server & GitHub Integration Gradle scripts use Kotlin DSL
- Added & updated dependencies so that GitHub Integration builds under Java 11

Closes #969


server/build.gradle.kts
```diff
@@ -0,0 +1,134 @@
...
+        maven("https://repo.spring.io/libs-release")
+    }
+}
+
+plugins {
+    kotlin("jvm").version("1.3.21")
+    kotlin("plugin.jpa").version("1.3.21")
+    kotlin("plugin.noarg").version("1.3.21")
+    kotlin("plugin.spring").version("1.3.21")
+    kotlin("plugin.allopen").version("1.3.21")
```
roxspring: I'll extract a `var kotlinVersion = "1.3.21"`

server/build.gradle.kts
```diff
@@ -0,0 +1,134 @@
...
+    src("http://json-schema.org/draft-04/schema")
+    dest("$rootDir/src/main/resources/schemas/json-schema.json")
+    onlyIfModified(true)
+}
+
+tasks.register("downloadSwaggerSchema", Download::class) {
+    src("http://swagger.io/v2/schema.json")
+    dest("$rootDir/src/main/resources/schemas/swagger-schema.json")
+    onlyIfModified(true)
+}
```
roxspring: Making these changes I noticed that `src/main/resources/schemas/json-schema.json` is committed to the repo but `src/main/resources/schemas/swagger-schema.json` isn't, while both really come from an outside source and ought to be treated consistently.

- Should we keep a copy in the repo?
  - If so, is it still worth having the build script attempt to update from elsewhere??
- Or should we keep neither in the repo and only download - in which case we probably ought to `.gitignore` them too

server/build.gradle.kts
```diff
@@ -0,0 +1,134 @@
...
+        kotlinOptions.jvmTarget = "1.8"
+    }
+}
+
+val versions = mapOf(
+        "jacoco" to "0.8.2"
+)
+
+dependencies {
+    val spring_boot_version = "2.0.4.RELEASE"
```
roxspring: I'll rename the variable `springBootVersion` to be more Kotlin style.

server/build.gradle.kts
maxim-tschumak: I vote for keeping neither in the repo and downloading on the fly.

github-integration/build.gradle.kts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+tasks.jacocoTestReport {
+    reports {
+        xml.isEnabled = true
+    }
+}
+
+tasks.jar {
+    archiveBaseName.set("bark")
```
maxim-tschumak: Let's move away from "bark" to more functional i.e. `zally-github-integration`.

server/build.gradle.kts
roxspring: Done: 043cb41450049707294e2b0ad7615c56c9ec98e3

roxspring:
- rebased on master
- addressed review comments

maxim-tschumak:
:+1:

tkrop:
:+1:

PR #972, opened 2019-04-01
UseOpenApiRule is not special

### LINKED ISSUE #851: Stop claiming parse/convert failures are UseOpenApiRule violations
When schemas break the Swagger to OpenAPI conversion (e.g. #846) the following validation failure is produced, incorrectly implying that the `UseOpenApiRule` is at fault:

```
MUST Provide API Specification using OpenAPI

must use YAML format
/paths
```

A natural reaction to this is to try and temporarily `x-zally-ignore` the "apparently broken rule" which doesn't work since (1) the violation didn't come from that rule and (2) rule ignoring isn't processed for conversion time anyway. Additionally users can mistakenly get the impression that other MUST rules passed successfully, unaware that they weren't even able to run!

We should:
- Avoid treating `UseOpenApiRule` specially and reusing it for violations that it doesn't produce
- Find a way to more honestly report the failure than pretending it's a normal violation
  - Perhaps it's good enough to throw and exception and respond with an application/problem+json?
  - Or at least having a separate rule such that parse/convert violations can be correctly attributed.

Thoughts?

roxspring:
- Refactored `Result` to reference strings of interest rather than instances of `RuleSet` or `Rule`
- Avoid creating `RuleSet` and `Rule` instances in tests just to build `Result` instances
- `RuleValidator` no longer references a `RuleSet` or `Rule` when parsing fails
- Documented a new `InternalRuleSet`

Closes #851

### DIFF
--- a/server/rules.md
+++ b/server/rules.md
@@ -2,6 +2,12 @@
 
 Zally comes with a number of Rule Sets which can be selected and used. The built in rule sets are documented below.
 
+# InternalRuleSet
+
+Internal rules which exist simply for the purpose of reporting fatal
+errors found while attempting to parse the API specification.
+The rules cannot be disabled or configured.
+
 # ZalandoRuleSet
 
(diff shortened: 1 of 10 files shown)


zeitlinger:
:+1:

maxim-tschumak:
:+1:

PR #975, opened 2019-04-03
feat(server): LimitNumberOfResourcesRule supports path whitelist

### LINKED ISSUE #974: LimitNumberOfResourcesRule should whitelist some paths
In-house we have some standard endpoints, and often a whole path prefix for private resources. A colleague complained that these were counting towards the limit, which he considered to be a limit on the number of public resources.

roxspring:
- adds a configurable whitelist of paths to ignore when counting and limiting the number of resources (empty by default).

Closes #974


maxim-tschumak:
:+1:

zeitlinger:
:+1:

PR #980, opened 2019-04-07
fix(server): AvoidTrailingSlashesRule accepts /

### LINKED ISSUE #951: "Avoid trailing slashes" contradicts "Base path can be extracted"
Having all paths start with the segment `/examples` and having `POST /examples` means I can't satisfy both rules at the same time because a path `/` is considered a trailing slash.

roxspring:
- `AvoidTrailingSlashesRule` accepts `/` as special case (presumably as intended).
- Moved logic from `PatternUtil` as it wasn't `Pattern` based and is rule-specific.
- Hopefully made the logic more understandable.

Closes #951


tkrop:
:+1:

PR #982, opened 2019-04-08
Overhauled MediaTypesRule

### LINKED ISSUE #950: JSON Patch is not considered a standard media type
Screenshot from 2019-03-14 15-43-31

https://tools.ietf.org/html/rfc6902

I guess the same applies to JSON Merge Patch.

roxspring:
- Rule accepts `application/json-patch+json` and `application/merge-patch+json` as discussed in #950.
- Rule now checks shared components, not just media types used via paths.
- Logic is self contained and not dependant on `PatternUtil` (it wasn't used by anything else anyway).
- Heavily refactored for increased clarity.

Closes #950

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/zalando/MediaTypesRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/zalando/MediaTypesRule.kt
@@ -5,6 +5,8 @@ import de.zalando.zally.rule.api.Context
 import de.zalando.zally.rule.api.Rule
 import de.zalando.zally.rule.api.Severity
 import de.zalando.zally.rule.api.Violation
+import io.swagger.v3.oas.models.Operation
+import io.swagger.v3.oas.models.media.MediaType
 
 /**
  * @see "https://opensource.zalando.com/restful-api-guidelines/#172"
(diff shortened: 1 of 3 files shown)


server/src/main/java/de/zalando/zally/rule/zalando/MediaTypesRule.kt
```diff
@@ -21,17 +24,49 @@ class MediaTypesRule {
...
+            .filterNot { (type, _) ->
+                isVersionedMediaType(type)
+            }
+            .map { (_, value) ->
+                context.violation(description, value)
+            }
+
+    fun isStandardJsonMediaType(mediaType: String): Boolean = mediaType in standardMediaTypes
+
+    fun isVersionedMediaType(mediaType: String): Boolean = "^\\w+/[-+.\\w]+;(v|version)=\\d+$".toRegex().matches(mediaType)
```
maxim-tschumak: Let's make a static field out of the regex (not creating on every function invocation).

server/src/main/java/de/zalando/zally/rule/zalando/MediaTypesRule.kt
```diff
@@ -21,17 +24,49 @@ class MediaTypesRule {
...
-                .flatMap { it.content?.entries.orEmpty() }
-            (consumedMediaTypes + producedMediaTypes)
-                .filter { isViolatingMediaType(it.key) }
-                .map { context.violation(description, it.value) }
-        }
-
-    private fun isViolatingMediaType(mediaType: String) =
-        !isApplicationJsonOrProblemJson(mediaType) && !isCustomMediaTypeWithVersioning(mediaType)
+        context.api.mediaTypes()
+            .filterNot { (type, _) ->
```
maxim-tschumak: Can we do this in one line (per higher-order function)

server/src/main/java/de/zalando/zally/rule/zalando/MediaTypesRule.kt
```diff
@@ -21,17 +24,49 @@ class MediaTypesRule {
...
+            }
+            .map { (_, value) ->
+                context.violation(description, value)
+            }
+
+    fun isStandardJsonMediaType(mediaType: String): Boolean = mediaType in standardMediaTypes
+
+    fun isVersionedMediaType(mediaType: String): Boolean = "^\\w+/[-+.\\w]+;(v|version)=\\d+$".toRegex().matches(mediaType)
+
+    private fun OpenAPI.mediaTypes(): List<Pair<String, MediaType>> =
```
tkrop: Why do you append this private function to `OpenAPI`. I would find it cleaner, if it sticks to this class where it is needed solemnly?

tkrop:
:+1:

server/src/main/java/de/zalando/zally/rule/zalando/MediaTypesRule.kt
roxspring: You mean to reformat it onto one line `.filterNot { (type, _) -> isStandardJsonMediaType(type) }` ?

Or are you asking to merge the two `filterNot()` calls into a single filter/filterNot?

server/src/main/java/de/zalando/zally/rule/zalando/MediaTypesRule.kt
roxspring: Particularly for functions which would only take one parameter, I tend to find that an extension function is more pleasant to read at invocation, and no worse at definition. I *feel* this is also the more idiomatic Kotlin approach but I have absolutely no evidence to cite. Also, unless I'm much mistaken, the fact that it's private means that it's only available within `MediaTypesRule.kt` class and so there's no downside of leaking this functionality to other users to OpenAPI.

server/src/main/java/de/zalando/zally/rule/zalando/MediaTypesRule.kt
maxim-tschumak: Yes, `.filterNot { (type, _) -> isStandardJsonMediaType(type) }`

server/src/main/java/de/zalando/zally/rule/zalando/MediaTypesRule.kt
roxspring: Always good to get new perspectives 😃 

I note though that [Kotlin extensions are resolved statically](https://kotlinlang.org/docs/reference/extensions.html#extensions-are-resolved-statically) and so the byte code manipulation I assume you're fearing isn't happening:

> Extensions do not actually modify classes they extend. By defining an extension, you do not insert new members into a class, but merely make new functions callable with the dot-notation on variables of this type.

From a reflection point of view, the function is listed as a member of `MediaTypesRule` and not of `OpenAPI` which seems reasonable to me.

Does any of that help with your fears?

tkrop: Oh, interesting. Kotlin is not handling this straight forward. Than it is okay for me.

tkrop:
:+1:

maxim-tschumak:
:+1:

PR #1013, opened 2019-06-26
Rules to check tags

### LINKED ISSUE #1010: TagAllOperationsRule
In-house we're about to have a push on making sure APIs are properly tagged so that Swagger UI docs look good, and are neatly divided up using tags as headings. Consequently, I'm considering contributing a `TagAllOperationsRule` with checks for:

* Operations have at least one tag
* Tag names are Title Case (configurable like `CaseChecker`)
* All tags appear in top level `tags:` section (defining heading order)
* All tags have a description

I'm leaning towards a `MUST` severity in each case but am happy to be argued down. Also happy if folks want it disabled in the default configuration.

I'm hoping that this might be accepted into the `ZallyRuleSet`.

roxspring:
- Add `CaseCheckerRule.checkTagNames()`
  - implements "Tag names are Title Case"
- Add `TagAllOperationsRule`
  - All operations have at least one tag
  - All tags are defined in the top level `tags:` section (defining order of groups)
  - All defined tags are used
  - All defined tags have a description
- Fix typo "avioded" in rules.md

Closes #1010

### DIFF
--- a/server/src/test/java/de/zalando/zally/rule/zally/TagAllOperationsRuleTest.kt
+++ b/server/src/test/java/de/zalando/zally/rule/zally/TagAllOperationsRuleTest.kt
@@ -14,6 +14,7 @@ class TagAllOperationsRuleTest {
     private fun TagAllOperationsRule.checkAll(context: Context): List<Violation> =
         checkOperationsAreTagged(context) +
             checkOperationTagsAreDefined(context) +
+            checkDefinedTagsAreUsed(context) +
             checkDefinedTagsAreDescribed(context)
 
     @Test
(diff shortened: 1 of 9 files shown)


server/src/test/java/de/zalando/zally/rule/zally/TagAllOperationsRuleTest.kt
```diff
@@ -0,0 +1,148 @@
...
+
+        val violations = cut.checkAll(context)
+
+        ZallyAssertions
+            .assertThat(violations)
+            .isEmpty()
+    }
+
+    @Test
+    fun `checkOperationsAreTagged with untagged operations returns violations`() {
```
zeitlinger: what's the rationale for this check?

It means that every spec needs at least one tag defined.

Wouldn't it make more sense to define the required tags somewhere else, such as in a config file?

zeitlinger:
@roxspring I like the idea - but I questions on "All operations have at least one tag" - see my comment.

maxim-tschumak:
@zeitlinger This is going to be a part of the `ZallyRuleSet` and deactivated per default. I think it's a valid (optional) "rule" to expect all operations to be tagged - it helps to describe the operations and group them. I also don't think that predefined `tags` make sense since the can be very generic.

maxim-tschumak:
:+1:

zeitlinger:
@maxim-tschumak OK, after some explanation, it makes sense now.

It would be nice to be able to configure checks instead of rules, so that the "All operations have at least one tag" could be configured separately, but we can address this when needed.

zeitlinger:
:+1:

roxspring:
Thanks both - forgot to leave it disabled by default though - will open another PR!

PR #1034, opened 2019-08-19
feat: make cli errors distinguish (#1033)

### LINKED ISSUE #1033: Zally CLI should finish successful in case of server errors (5xx, 4xx)
User Story
As a developer
I want the Zally CLI to succeed on (internal) server errors.
So that the build pipeline can proceed successful in case of Zally problems.

Description
We had a Zally outage because we could not reach the token info endpoint for authorization. As result we returned 502 - Bad Gateway.

tkrop:
fixes #1033.

### DIFF
--- a/cli/zally/commands/lint.go
+++ b/cli/zally/commands/lint.go
@@ -1,16 +1,12 @@
 package commands
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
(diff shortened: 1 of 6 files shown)


zeitlinger:
:+1:

tkrop:
:+1:

PR #1063, opened 2019-10-09
Add support for Rule 235 from API Guidelines

### LINKED ISSUE #1062: Add rule 235 support
Add [rule 235](https://opensource.zalando.com/restful-api-guidelines/#235) support.

vadeg:
Add [rule 235](https://opensource.zalando.com/restful-api-guidelines/#235) support from API Guidelines.

Fixes #1062

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/zalando/DateTimePropertiesSuffixRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/zalando/DateTimePropertiesSuffixRule.kt
@@ -1,0 +1,46 @@
+package de.zalando.zally.rule.zalando
+
+import com.typesafe.config.Config
+import de.zalando.zally.rule.api.Check
+import de.zalando.zally.rule.api.Context
+import de.zalando.zally.rule.api.Rule
+import de.zalando.zally.rule.api.Severity
+import de.zalando.zally.rule.api.Violation
+import de.zalando.zally.util.getAllProperties
+import io.swagger.v3.oas.models.media.Schema
+
+@Rule(
+    ruleSet = ZalandoRuleSet::class,
+    id = "235",
+    severity = Severity.SHOULD,
+    title = "Name date/time properties using the \"_at\" suffix"
+)
+class DateTimePropertiesSuffixRule(rulesConfig: Config) {
+
+    private val patterns: Set<Regex> = rulesConfig
+        .getStringList("${DateTimePropertiesSuffixRule::class.java.simpleName}.patterns")
+        .map { Regex(it) }
+        .toSet()
+
+    private val propertyFormats = setOf("date", "date-time")
+
+    @Check(severity = Severity.SHOULD)
+    fun validate(context: Context): List<Violation> {
+        return context.api.getAllProperties().map {
+            val schema = it.value
+            val result = checkProperty(it.key, schema)
+            result?.let { context.violation(result, schema) }
+        }.filterNotNull()
+    }
+
+    private fun checkProperty(name: String, schema: Schema<Any>): String? {
+        if (schema.type == "string" && schema.format in propertyFormats) {
+            patterns.find { it.matches(name) } ?: return generateMessage(name, schema.type, schema.format)
+            return null
+        }
+        return null
+    }
+
+    internal fun generateMessage(name: String, type: String, format: String) =
+        """Property "$name" of type "$type" and format "$format" should match one of the patterns ${patterns.map { it.toString() }}""""
+}
--- a/cli/zally/integration_test.go
+++ b/cli/zally/integration_test.go
@@ -81,9 +81,9 @@ func TestIntegrationWithNoMustViolations(t *testing.T) {
 
 	must, should, may, hint := countViolations(out)
 
-	assert.Zero(t, must)
-	assert.True(t, should == 0)
-	assert.True(t, may > 0)
+	assert.Zero(t, must, "No MUST violations expected")
+	assert.True(t, should == 0, "No SHOULD violation expected")
+	assert.True(t, may > 0, "At least one MAY violation expected")
 	assert.Equal(t, 0, hint)
 	assert.Nil(t, e)
 }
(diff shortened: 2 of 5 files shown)


server/src/main/java/de/zalando/zally/rule/zalando/DateTimePropertiesSuffixRule.kt
```diff
@@ -0,0 +1,44 @@
...
+    fun validate(context: Context): List<Violation> {
+        return context.api.getAllProperties().map {
+            val schema = it.value
+            val result = checkProperty(it.key, schema)
+            result?.let { context.violation(result, schema) }
+        }.filterNotNull()
+    }
+
+    private fun checkProperty(name: String, schema: Schema<Any>): String? {
+        if (schema.type == "string" && name in propertyNames && schema.format in propertyFormats) {
```
roxspring: Is the rule really intended to just check a prescripted set of `propertyNames`? I read that list as examples and thought the rule was implying _any_ properties of `date` or `date-time` format should match the pattern. I was hoping it'd be implemented using a configurable regex so that e.g. I could reconfigure it to check for `.*-at` regex instead.

server/src/main/java/de/zalando/zally/rule/zalando/DateTimePropertiesSuffixRule.kt
```diff
@@ -0,0 +1,44 @@
...
+import de.zalando.zally.rule.api.Severity
+import de.zalando.zally.rule.api.Violation
+import de.zalando.zally.util.getAllProperties
+import io.swagger.v3.oas.models.media.Schema
+
+@Rule(
+    ruleSet = ZalandoRuleSet::class,
+    id = "235",
+    severity = Severity.SHOULD,
+    title = "Name date/time properties using the \"_at\" suffix"
```
roxspring: Obviously a self-serving comment but it seems a shame to hard code the expected extension into the rule title as it means that the rule can't easily be used by those of us that don't subscribe to `snake_case` property naming.

server/src/main/java/de/zalando/zally/rule/zalando/DateTimePropertiesSuffixRule.kt
tkrop: @vadeg valid request can we go into this direction to allow for customization.

server/src/main/java/de/zalando/zally/rule/zalando/DateTimePropertiesSuffixRule.kt
tkrop: @roxspring hm, do you now how should fix this?

vadeg: @roxspring I would not say it is a shame because this rule serves exactly one purpose [mentioned in the rule name](https://opensource.zalando.com/restful-api-guidelines/#235). I understand and agree with your point below about camel case. I will think how to implement it.

server/src/main/java/de/zalando/zally/rule/zalando/DateTimePropertiesSuffixRule.kt
vadeg: Well, I do not think that any `date` or `date-time` properties should have suffix. For example API can have properties like `start_date` or `end_date` which do not need prefix.

roxspring: Seems to me like you've just established a 2nd valid pattern of `.*_date` there 😄.

The guideline heading and description seem very general to me:

> ### Should: Name date/time properties using the `_at` suffix
> Dates and date-time properties should end with _at to distinguish them from boolean properties which otherwise would have very similar or even identical names:

To implement the rule as checking the 4 stated examples + 2 stated exceptions, and noting that there are untold other patterns used but we don't care about them either seems overly specific to me.

As a more general implementation, I'm imagining the following default configuration in `rules-config.conf`, specifying the accepted patterns (and stated exceptions):

```
DateTimePropertiesSuffixRule {
  whitelist: [
    .*_at,
    .*_date,
    created,
    modified
  ]
}
```

One can imagine another organisation utilising the same rule to enforce kebab-case date/time properties that always end in `-tstamp`.

server/src/main/java/de/zalando/zally/rule/zalando/DateTimePropertiesSuffixRule.kt
roxspring: We've hit this before. We "resolved" it by duplicating the specific rule (e.g. `SnakeCaseForQueryParamsRule`) with a more general one (`CaseCheckerRule.checkQueryParameterNames`) and worked to share the implementation. As a resolution I find it sub-optimal as when running with a different configuration I have to just disable the Zalando specific rule.

I understand the desire to have Zalando rules in lock-step with the guidelines word for word, but it comes at the cost of the rules being very rigid and unconfigurable. As an alternative, I'd have thought that here a title of "Date/Time properties should match expected patterns" and a message of "Date/Time property ${propertyName} does not match an expected pattern (${acceptedPatterns.joinToString()})" would functional just as well, but be responsive to configuration.

Just my 2p

tkrop: Still we have the problem of needing rules with description (exactly) matching the (specific) API guideline.

@vadeg I suggest to not change the title, but refactor the rule, so that @roxspring can easily reuse the code for an own rule.

roxspring: I'd be happy with that compromise 👍

vadeg: Yep, agree. Will update the code with regex configuration.

tkrop: @vadeg Thanks. Let's go the last mile and close this PR.

tkrop:
:+1:

roxspring:
Won’t get chance to review until next week. Happy to raise separate issue/mr if you’re in a hurry and I find something later. 

Thanks,

Rob

vadeg:
👍

PR #1065, opened 2019-10-09
Fix build warnings

### LINKED ISSUE #1064: Fix build warnings
Fix the following warnings during the build:
```
WebMvcConfiguration.kt: (11, 58): 'WebMvcConfigurerAdapter' is deprecated. Deprecated in Java
```
```
de/zalando/zally/rule/CaseChecker.kt: (136, 29): Name shadowed: schema
```
```
de/zalando/zally/util/SwaggerUtil.kt: (28, 42): 'getter for schema: Property!' is deprecated. Deprecated in Java
```

vadeg:
* Replace `WebMvcConfigurerAdapter` with `WebMvcConfigurer`.
* Rename `schema` in `CaseChecker` to `propertySchema` to remove shadowing.
* ~~Replace `schema` with `responseSchema` in `SwaggerUtil` and update test which checks which JSON objects are traversed.~~ Remove `SwaggerUtil`.

Fixes #1064


server/src/test/java/de/zalando/zally/util/SwaggerUtilTest.kt
```diff
@@ -31,7 +31,9 @@ class SwaggerUtilTest {
                 "/definitions/NameForParam",
                 "/definitions/PetName last_name",
                 "/definitions/PetName middle_name items",
-                "/definitions/Error"
+                "/definitions/Error",
+                "/definitions/Pets",
+                "/definitions/Names"
```
roxspring: Curious why these two entries need to be added to the test, given everything else seems like pure refactor.

tkrop:
:+1:

server/src/test/java/de/zalando/zally/util/SwaggerUtilTest.kt
tkrop: Good question.

vadeg: @roxspring @tkrop sorry for big delay in reply. I will update you with "why".

vadeg: I think I figured out why it happens. First of all
* It is related only to Swagger 2.0
* Method `SwaggerUtilKt#getAllJsonObjects` used only in tests. 
* `Response.schema` property is never initialized and has been deprecated swagger-api/swagger-parser#567 because of incorrect final schema resolution.

## Before change

* [`response.schema` call](https://github.com/zalando/zally/blob/master/server/src/main/java/de/zalando/zally/util/SwaggerUtil.kt#L28) returns `RefProperty` instance
```
        is RefProperty -> {
            val model = definitions.orEmpty()[property.simpleRef]
            findJsonObjects(model, "/definitions/${property.simpleRef}", visitedPaths)
        }
```
* instance resolved by [another `findJsonObjects` function](https://github.com/zalando/zally/blob/master/server/src/main/java/de/zalando/zally/util/SwaggerUtil.kt#L58) which traverse directly on _model properties_ in `#/definitions/<model ref> `

Having schema
```
swagger: "2.0"
info:
  version: 1.0.0
  title: Swagger Petstore
host: petstore.swagger.io
paths:
  /pets:
    get:
      summary: List all pets
      operationId: listPets
      responses:
        "200":
          description: An paged array of pets
          schema:
            $ref: '#/definitions/Pets'

definitions:
  Pet:
    required:
      - id
      - name
    properties:
      id:
        type: integer
        format: int64
  Pets:
    type: array
    items:
      $ref: '#/definitions/Pet'
```
method will return
```
[/definitions/Pet]
```
which is incorrect because `/definitions/Pets` is ignored.

## After change

* `response.responseSchema` returns `RefModel` (instead of `RefProperty`) before.
* `RefModel` processing happens [here](https://github.com/zalando/zally/blob/master/server/src/main/java/de/zalando/zally/util/SwaggerUtil.kt#L74)
```
        is RefModel ->
            findJsonObjects(model.properties, "/definitions/${model.simpleRef}", visitedPaths)
```

Since `model.properties == null` (in this example) the only path written will be `/definitions/Pets` which is correct according to schema. The final output will be
```
[/definitions/Pet, /definitions/Pets]
```

## Summary

I have two options here:

1. Remove this method because
1.1. It is overcomplicated and unclear what exactly should be returned.
1.2. It is related only to Swagger 2.0 which is some sort of outdated because Swagger 3.0 released. The last time when Swagger 2.0 was updated was 5 years ago.
1.3. This is method used only in tests.

2. Fix traversal path by specifying what exactly the method should return.

I vote for option 1. Would like to hear your opinions.

roxspring: Ahhh 🤦‍♂ - so I was wrong to look at this as a _pure_ refactor because I was missing that the following change was meaningful:

```diff
-                findJsonObjects(response.schema, "$path $verb $code", visitedPaths)
+                findJsonObjects(response.responseSchema, "$path $verb $code", visitedPaths)
```

Thanks for the detailed explanation. I hadn't noticed that `Swagger.getAllJsonObjects()` was only being used in tests at this point, and all real schema traversal had moved to use `OpenAPI.getAllSchemas()` or  `OpenAPI.getAllTransitiveSchemas()` instead. Based on this then I agree lets cut out the confusion and remove the unused method.

As a sidenote, this all reminds me how confusing I find the method names `OpenAPI.getAllSchemas()` and `OpenAPI.getAllTransitiveSchemas()`. The former pointedly does not get all schemas as the transitive ones are missed out, while the latter's name suggests to me that it gets _only_ the transitive schemas but actually returns far more. It makes me wonder whether we have bugs where the former is being used but the latter intended and whether we really need both at all!

vadeg: As for me I also find `OpenAPI.getAllSchemas()` and `OpenAPI.getAllTransitiveSchemas()` confusing.

tkrop:
:+1:

tkrop:
@vadeg lets merge.

vadeg:
👍

PR #1068, opened 2019-10-18
feat(server): FunctionalNamingForHostnamesRule supports Swagger specs

### LINKED ISSUE #1067: host turns into basePath
the following patch fails with

```
org.junit.ComparisonFailure: 
Expected :"/host"
Actual   :"/basePath"
<Click to see difference>
```

@roxspring the reason for basePath is `de.zalando.zally.util.ast.JsonPointers#regexToReplacement` but I don't understand why `servers` should be translated to `basePath` rather than `host`

patch
```
Index: server/src/test/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRuleTest.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- server/src/test/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRuleTest.kt  (date 1569243428000)
+++ server/src/test/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRuleTest.kt  (date 1571404509132)
@@ -1,6 +1,7 @@
 package de.zalando.zally.rule.zalando
 
 import de.zalando.zally.getOpenApiContextFromContent
+import de.zalando.zally.getSwaggerContextFromContent
 import de.zalando.zally.rule.api.Context
 import org.assertj.core.api.Assertions.assertThat
 import org.intellij.lang.annotations.Language
@@ -86,6 +87,27 @@
         assertThat(rule.mayFollowFunctionalNaming(context)).isEmpty()
     }
 
+    @Test
+    fun `host should return no violations `() {
+        @Language("YAML")
+        val context = getSwaggerContextFromContent(
+            """
+            swagger: '2.0'
+            info:
+              x-audience: company-internal
+            schemes:
+              - https
+            host: google.com
+        """
+        )
+
+        val violations = rule.shouldFollowFunctionalNaming(context)
+

```

roxspring:
Closes #1067

### DIFF
--- a/server/src/main/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRule.kt
@@ -53,7 +53,7 @@ class FunctionalNamingForHostnamesRule {
         .map { context.violation(description, it.url) }
         .toList()
 
-    private fun checkHostnamesInSwaggerHost(context: Context): List<Violation> = context.swagger?.host.let { host ->
+    private fun checkHostnamesInSwaggerHost(context: Context): List<Violation> = context.swagger!!.host.let { host ->
         when {
             host == null || isUrlValid(host) -> emptyList()
             else -> context.violations(description, JsonPointer.compile("/host"))
--- a/server/src/main/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRule.kt
+++ b/server/src/main/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRule.kt
@@ -1,5 +1,6 @@
 package de.zalando.zally.rule.zalando
 
+import com.fasterxml.jackson.core.JsonPointer
 import de.zalando.zally.rule.api.Check
 import de.zalando.zally.rule.api.Context
 import de.zalando.zally.rule.api.Rule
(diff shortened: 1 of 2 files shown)


server/src/main/java/de/zalando/zally/rule/zalando/FunctionalNamingForHostnamesRule.kt
```diff
@@ -38,16 +39,24 @@ class FunctionalNamingForHostnamesRule {
     internal fun isUrlValid(url: String): Boolean = functionHostnameURLRegEx.matches(url)
 
     private fun checkHostnames(audiencesToCheck: List<String>): (context: Context) -> List<Violation> = { context ->
```
roxspring: I'm curious about this existing method signature... is there any advantage over the following?

    private fun checkHostnames(audiencesToCheck: List<String>, context: Context): List<Violation>

I'd naturally go with this signature as it seems simpler to understand and use (although I'd probably switch the param order) so wondering if there's some functional programming advantage I've missed.

vadeg: I would say having two sets of arguments instead of two just because of signature complicates the code and raises more questions than answers. 
What can be simpler than a function with two arguments? Function with one argument but not with two argument sets :)

But from another side you can go further and change
```
fun mustFollowFunctionalNaming(context: Context): List<Violation> = checkHostnames(mustFollow)(context)
```
to
```
val mustFollowFunctionalNaming: (Context) -> List<Violation> = checkHostnames(mustFollow)
```

tkrop: @roxspring aren't that two totally different functions? One is returning a `List<Violation>` the other a function `(Context) -> List<Violation>`. If the function is required, I see no alternative. Else I would also always go for the simple version.

roxspring: The function was private so the existing signature wasn't required.

I've updated the PR with the simplified signature `checkHostnames(context: Context, audiencesToCheck: List<String>): List<Violation>` (Context first for consistency with other functions)

tkrop:
:+1:

tkrop:
@vadeg @zeitlinger @gewald can one approve and merge?

zeitlinger:
:+1:

PR #1073, opened 2019-10-22
feat(server): apply inherited x-zally-ignores in addition to local ones

### LINKED ISSUE #1072: Low level x-zally-ignores should aggregate with higher level ones rather than replace them
A colleague was surprised today to find that rules ignored at root level are not ignored if there's a shallower `x-zally-ignore` that doesn't ignore the rule. For example a violation of `IGNORED_AT_ROOT` rule in `/info/contact/email` in the following spec does not get ignored:

```yaml
swagger: '2.0'
info:
  title: Some API
  version: '1.0.0'
  contact:
    name: Team X
    email: team@x.com
    url: https://team.x.com
  x-zally-ignore: [IGNORED_AT_INFO]
paths: {}
x-zally-ignore: [IGNORED_AT_ROOT]
```

Instead we suggest that any violation under `/info` in this example should ignore violations of both `IGNORED_AT_INFO` and `IGNORED_AT_ROOT`.

roxspring:
fixes #1072


zeitlinger:
:+1:

tkrop:
:+1:

Issue #1075, opened 2019-10-23
docker-compose Zally web UI not rendering


mckibbenc:
When running docker-compose on Mac OS, the web-ui will not render.  The Zally API server works just fine, and I'm able to get the active rules and lint my APIs.  There are no build errors when running docker-compose, and I don't see any errors in the web-ui logs

docker container ls>
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS                     PORTS                              NAMES
f23c0a377db0        zally_web-ui        "node server.js"         3 minutes ago       Up 3 minutes               0.0.0.0:8080->3000/tcp             zally_web-ui_1
4462b742b339        zally_server        "/bin/sh -c 'java $(…"   3 minutes ago       Up 3 minutes (unhealthy)   0.0.0.0:8000->8000/tcp, 8080/tcp   zally_server_1
55d8f4af4219        postgres:9.6        "docker-entrypoint.s…"   3 minutes ago       Up 3 minutes               0.0.0.0:54321->5432/tcp            zally_postgres_1

docker logs zally_web-ui_1>
zally-web-ui mounted on "/"
zally-web-ui running at http://localhost:3000

mckibbenc:
I may have isolated this issue.  I modified the web UI Dockerfile to use the latest version of Node.js (version 13) and rebuilt the image.  Now the web UI renders properly.  My guess is the current version of node being referenced in the web ui Dockerfile (registry.opensource.zalan.do/stups/node:10.15.3-45) can't properly support the latest changes to the web ui, maybe due to outdated packages.  Here is the Dockerfile I used to build the working web ui image:

FROM node

MAINTAINER "http://zalando.github.io/"

COPY package.json ./
COPY yarn.lock ./

RUN npm install -g yarn
RUN yarn --production

COPY src ./src
COPY server.js ./

EXPOSE 3000

CMD ["node", "server.js"]


I would create a PR for this, but my guess is you guys will want to vendor your own updated node image instead of pulling from Docker hub.

mckibbenc:
So it may not have to do with the node version.  I did a fresh clone on my Mac with the new node version, ran ./build-and-run.sh and then the docker-compose up, and it was still the same issue.  I could only get it to work if I manually built the web-ui by running npm install and yarn build within the web-ui directory.  ./build-and-run.sh fails on my Mac because of a :dokka issue.  I'm going to close this comment as it's probably user error on my end.

dfrechette:
[...]

PR #1078, opened 2019-10-24
fix for issue #1075 - Update Dockerfile

mckibbenc:
This fixes issue #1075

The previous version of node referenced in the web-ui Dockerfile may have had outdated packages needed by the web-ui, as attempting to deploy the web-ui via Docker resulted in the web-ui not rendering the page.  Once I updated node to v12, the web-ui properly rendered the Zally editor.

### DIFF
--- a/web-ui/Dockerfile
+++ b/web-ui/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.opensource.zalan.do/stups/node:10.15.3-45
+FROM node:12
 
 MAINTAINER "http://zalando.github.io/"
 


web-ui/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.opensource.zalan.do/stups/node:10.15.3-45
+FROM node:12
```
tkrop: Can you please try with `registry.opensource.zalan.do/stups/node:12.10.0-48` and amend the commit comment to contain `(#1075)` as reference to the issue at the end.

mckibbenc:
Going to open this PR back up with the suggested fixes.

PR #1080, opened 2019-10-24
fix issue #1075 - update dockerfile

mckibbenc:
Updated web-ui Dockerfile Node dependency to v12.  

This fixes issue #1075.


vadeg:
Let's update [package.json](https://github.com/zalando/zally/blob/master/web-ui/package.json#L7)   as well then.

mckibbenc:
Should I bump the version to 0.4.1 or 0.5.0?  Also, are you asking me to bump the node version to >=12?

mckibbenc:
FYI, please see issue #1075 for updates.  I don't think this PR is necessary anymore as the issue may have been user error on my part.

PR #1093, opened 2019-11-24
Drop unnecessary MethodCallRecorder

### LINKED ISSUE #1092: Drop unnecessary MethodCallRecorder
`MethodCallRecorder` uses AOP proxies to track the usage of `OpenAPI` and `Swagger` models in order to power automatic guesses at the `JsonPointer` location where violations should be attached. As a side benefit it also delivers some `NullPointerException` protection in a couple of areas. But...

- Reliance on AOP means that breaking out the core and rules into subprojects (as discussed in #560) necessitates taking some pretty big dependencies along for the ride.
- Guessing the location amounts to remembering the last element of the model that was accessed, and frequently points to the wrong element's location.
- The `NullPointerException` protection only affects precious few cases and we have to be explicitly defensive in other cases, so why not be explicitly defensive in all cases?

I think it's time to drop the `MethodCallRecorder` and require that locations (or model elements linked to locations) are identified explicitly by the rules.

roxspring:
- Drop MethodCallRecorder
- Require explicit location parameters to Context calls
- Move getContext methods from TestUtil to DefaultContextFactory in preparation for splitting out sub-modules.

Fixes #1092


zeitlinger:
:+1:

tkrop:
Looks great. Thanks for the contribution. I only think whether https://github.com/zalando/zally/pull/1093/files#diff-7bfeede15f02605078282e929acd0454R24-R34 could be expressed more speaking or easier to reuse, however I have no clear idea now.

If you have also no suggestion than lets merge.

roxspring:
I'm not impressed with the readability of that either. The reason the change was required was to expose the `responses` in order to look up the base `context.getJsonPointer(responses)` JsonPointer. The only options that spring to mind are to change the `context.validateResponses`'s lamda's signature from:

```
    fun validateResponses(
        pathFilter: (Map.Entry<String, PathItem?>) -> Boolean = { true },
        operationFilter: (Map.Entry<PathItem.HttpMethod, Operation?>) -> Boolean = { true },
        responseFilter: (Map.Entry<String, ApiResponse?>) -> Boolean = { true },
        action: (Map.Entry<String, ApiResponse?>) -> List<Violation?>
    ): List<Violation>
```

Either adding the responses so that the lookup can be done easily:

```
    fun validateResponses(
        pathFilter: (Map.Entry<String, PathItem?>) -> Boolean = { true },
        operationFilter: (Map.Entry<PathItem.HttpMethod, Operation?>) -> Boolean = { true },
        responseFilter: (Map.Entry<String, ApiResponse?>) -> Boolean = { true },
        action: (ApiResponses, Map.Entry<String, ApiResponse?>) -> List<Violation?>
    ): List<Violation>

```

Or doing the lookup in advance and passing that location:

```
    fun validateResponses(
        pathFilter: (Map.Entry<String, PathItem?>) -> Boolean = { true },
        operationFilter: (Map.Entry<PathItem.HttpMethod, Operation?>) -> Boolean = { true },
        responseFilter: (Map.Entry<String, ApiResponse?>) -> Boolean = { true },
        action: (JsonPointer, Map.Entry<String, ApiResponse?>) -> List<Violation?>
    ): List<Violation>
```

I'm guessing the former might be more useful (e.g. allowing you to check something about a response in the context of its siblings) but either would be possible.

That said, it looks like this was the only rule using that convenience method anyway, and this change leaves it unused. Maybe we need to review how convenient the context.validate*() convenience methods really are!

tkrop:
The first options is looking to misleading from the semantic. The second option as a concept looks much better, however if we do not add this to other places consistently, I would not start doing this here. So lets stick with the current code.

tkrop:
:+1:

PR #1097, opened 2019-12-18
Extract Zally-Core subproject

roxspring:
- Created "zally-core" subproject
  - The intention is to move all of the core rule processing here, hopefully without spring dependencies if they can be avoided.
  - Paving the way for clearly separate modules as discussed in #560 
- Moved JsonPointers utilities into the core project
  - Refactored them to be more idiomatic Kotlin extension functions.


zeitlinger:
:+1:

vadeg:
👍

vadeg:
@tkrop, @maxim-tschumak do want also to check or should I merge it?

server/src/main/java/de/zalando/zally/util/ast/ReverseAstBuilder.kt
```diff
@@ -94,7 +96,9 @@ class ReverseAstBuilder<T : Any> internal constructor(root: T) {
                         // We must not use the method name but re-use the current pointer.
                         nodes.push(Node(value, pointer, marker, /* skip */true))
                     } else {
-                        nodes.push(Node(value, pointer.append(JsonPointers.escape(m)), marker))
+                        nodes.push(Node(value, pointer + m.name
+                            .let { if (it.startsWith("get")) it.drop(3) else it }
```
mkulak: .removePrefix("get")

vadeg:
👍

vadeg:
@tkrop @zeitlinger could you please review and bump this PR?

server/zally-core/src/main/kotlin/de/zalando/zally/core/JsonPointerExtensions.kt
```diff
@@ -0,0 +1,65 @@
...
+ * @return the combined JsonPointer
+ */
+operator fun JsonPointer.plus(other: JsonPointer): JsonPointer = append(other)
+
+/**
+ * Escapes a string and appends it to an existing JsonPointer.
+ * Escaping implemented according to https://tools.ietf.org/html/rfc6901
+ * @return the combined JsonPointer.
+ */
+operator fun JsonPointer.plus(unescaped: String): JsonPointer = append(unescaped
```
zeitlinger: This violates the principle of least surprise - I would expect the following to be equal:

1. `"/a".toJsonPointer() + "/b".toJsonPointer()`
2. `"/a".toJsonPointer() + "/b"`

It's not the same because (2) escapes the `/`.

server/src/main/java/de/zalando/zally/rule/zally/NoUnusedDefinitionsRule.kt
```diff
@@ -88,19 +87,19 @@ class NoUnusedDefinitionsRule {
...
         description: String,
         used: Set<JsonPointer>
     ): List<Violation> {
-        val ptr = JsonPointer.compile(pointer)
+        val ptr = pointer.toJsonPointer()
         return root.at(ptr)
             ?.fieldNames()
             ?.asSequence()
-            ?.map { ptr.append(JsonPointers.escape(it)) }
+            ?.map { ptr + it }
```
zeitlinger: this is the example that let me to the conclusion that the hidden escape is too magic.

zeitlinger:
@roxspring great work - sorry that I didn't catch this thing with escape before.

server/zally-core/src/main/kotlin/de/zalando/zally/core/JsonPointerExtensions.kt
roxspring: Yeah, that's fair enough. I think I got carried away with the DRY principle since we use this behaviour so much. I'll try a `String.toEscapedJsonPointer()` extension function instead.

zeitlinger: sounds good - the same happened to me a couple of times :smile: 
DRY is still right - don't lose faith!

zeitlinger:
:+1:

vadeg:
👍

PR #1107, opened 2020-01-19
Drop Spring dependencies from rule loading & processing

### LINKED ISSUE #122: Consider using JDK-Standard 'ServiceLoader' instead of Spring

roxspring:
- Removed some Spring dependencies in preparation for spring-free core subproject #560:
  - RulesValidators
  - Remove `SwaggerRulesValidator` which was no longer being used for production.
  - RuleSet implementations
- Rule instantiation is now driven ServiceLoader style without dependency on Spring
  - Added Kotlin annotation processor `RuleProcessor` to produce resources based on `@Rule` annotations
  - Added `RuleManager` companion object with factory methods to process resources and load the instances.
  - Closes #122


zeitlinger:
:+1:

zeitlinger:
@roxspring always a pleasure to read your Kotlin code - I'll recommend it to our Kotlin guild!

tkrop:
:+1:

tkrop:
@roxspring thanks for the contribution.

PR #1114, opened 2020-01-23
Move server to its own module and clean up dependencies

gsdatta:
I was working on the same refactor as you, so I decided to pause and just branch off your existing work. Thanks so much for taking this on!

This PR does a couple things:

- Move server code into its own submodule rather than top level `src`, which is the suggested spring way of doing things. This will make it easier to publish the subprojects
- Remove all spring dependencies from the top level and move to the server module

Not completely cleaned up, but this might be a useful starting point.


roxspring:
Great! I’d wondered whether this was the next step. Will take a proper look soon!

gsdatta:
Awesome, thanks! 

It's essentially based on the concepts outlined here https://spring.io/guides/gs/multi-module - I've used this structure with some success in the past. It becomes especially useful if you have DTOs you want to share across projects or (more relevant) split core functionality into domain vs web libs.

zeitlinger:
@gsdatta please rebase your branch

tkrop:
@gsdatta as we have merged the changes from this branch to our master you can also target it directly.

roxspring:
Sorry. Was deleting some of my own dead branches and accidentally deleted this one too. Trying to figure out what I can do to reopen or otherwise make up for it.

gsdatta:
No prob, let me rebase and open a new PR against master

server/build.gradle.kts
```diff
@@ -21,14 +21,8 @@ plugins {
 
     kotlin("jvm").version(kotlinVersion)
     kotlin("kapt").version(kotlinVersion)
```
roxspring: If there's no code left in the root project then I'm guessing we don't really need these plugins applied to the root project either? ... perhaps the `allprojects` block will need to become a `subproject` block to accommodate that change too?

server/build.gradle.kts
```diff
@@ -118,51 +119,13 @@ allprojects {
...
-    compile("org.postgresql:postgresql:42.2.4")
-    compile("org.hibernate:hibernate-core:5.3.5.Final")
-    compile("org.jadira.usertype:usertype.core:7.0.0.CR1") {
-        exclude("org.hibernate", "hibernate-entitymanager")
-    }
     compile("com.fasterxml.jackson.module:jackson-module-parameter-names:2.10.2")
     compile("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.10.2")
     compile("com.fasterxml.jackson.datatype:jackson-datatype-jdk8:2.10.2")
     compile("com.fasterxml.jackson.module:jackson-module-kotlin:2.10.2")
     compile("org.yaml:snakeyaml:1.24")
```
roxspring: Similarly to above... if the code is all gone from the root project then it seems unlikely we should be left with any dependencies.

server/build.gradle.kts
```diff
@@ -171,10 +134,6 @@ jacoco {
     toolVersion = "0.8.2"
 }
 
-tasks.bootRun {
-    jvmArgs = listOf("-Dspring.profiles.active=dev")
-}
```
roxspring: I wonder if it's worth keeping a `bootRun` task - even if it just depends on `:zally-server:bootRun` - just to keep an easy way of running a dev server.

server/zally-server/build.gradle.kts
```diff
@@ -0,0 +1,66 @@
+buildscript {
+    repositories {
+        gradlePluginPortal()
+        maven("https://repo.spring.io/libs-release")
+    }
+}
```
roxspring: Do we need this `buildscript` block? I don't think the other subprojects have one...

PR #1115, opened 2020-01-27
require only 1 zalando approval 

zeitlinger:
require only 1 zalando approval

### DIFF
--- a/.zappr.yaml
+++ b/.zappr.yaml
@@ -6,7 +6,8 @@
 approvals:
   groups:
     zalando:
-      minimum: 2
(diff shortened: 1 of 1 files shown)


zeitlinger:
:+1:

zeitlinger:
> Nice idea, but I think this would violate our open source policy as well as our 4-eyes deployment principles, as this is directly going to production.

the 1 approval is ignoring the last committer, so it is 4 eyes, but not necessarily 4 zalando eyes.

tkrop:
For Zalando only internal eyes count.

PR #1117, opened 2020-01-30
Simplified RuleSet implementations

roxspring:
- Simplified most `RuleSet` implementations by using default logic in `AbstractRuleSet` (as requested by @tkrop)
- Removed `TestRuleSet` which was accidentally left behind in refactoring
- Removed `RuleSet.title` property since it wasn't actually being used anywhere 👈 might be preferable to "fix" this by actually using the title somewhere but I figure we can add the title back if/when that comes up.

#560


server/src/test/java/de/zalando/zally/apireview/RestApiTestConfiguration.kt
```diff
@@ -29,18 +28,7 @@ class RestApiTestConfiguration {
             TestCheckAlwaysReport3MustViolations()
         ))
 
-    class TestRuleSet : AbstractRuleSet() {
-
-        override val id: String = javaClass.simpleName
-
-        override val title: String = "Test Rules"
```
tkrop: Funny. Have we really nowhere use the title?

tkrop:
:+1:

zeitlinger:
:+1:

server/src/test/java/de/zalando/zally/apireview/RestApiTestConfiguration.kt
roxspring: Really looks like we haven't. 🤦‍♂

PR #1119, opened 2020-02-03
fix: quick openapi 3 parsing link fix (#1119)

tkrop:
fixes #1119.

To stop the bleeding.


tkrop:
Does not work, but #1120

PR #1120, opened 2020-02-03
fix(server): Redirect openapi 3 schema references to local resource

roxspring:
Not really happy with this but I think it might be a good temporary fix if it gets through CI.

#1119


tkrop:
:+1:

vadeg:
👍

PR #1124, opened 2020-02-05
Check which rules @SpringBootTests are testing

roxspring:
fixes #1123

PR #1126, opened 2020-02-08
Simplify UseOpenApiRule

### LINKED ISSUE #1125: Simplify UseOpenApiRule
The recent #1119 explosion made me take a closer look at `UseOpenApiRule` and I'm concerned it's over complicated and surprising:

1. The underlying `JsonSchemaValidator` assumes that each validation message will have a location within the document, leading to the `NullPointerException` in #1119. The violation location should default to the empty JsonPointer in such cases.
2. We have built-in `openapi-2-schema.json` and `swagger-schema.json` which are essentially the same thing. I'd like to consolidate on just one copy. Further, we should align the constant names and resource names so that the latter can be derived.
3. By default we don't even use these built-in versions, instead grabbing fresh schemas from the internet on startup. I think our default configuration should use the built-in schemas, and only reach out to the internet if explicitly configured to do so - for repeatability sake if nothing else.
4. Misconfigured schema locations are glossed over with a warning. This could lead to surprises when people think they're validating against a custom schema but it's defaulted back to the built-in version. Explicitly misconfiguration should make the server fail fast and avoid such surprises. Falling back to the built-in version should occur _only_ when there is no custom version configured.
5. When _either_ schema (OpenAPI 3 vs swagger) fails we fallback to the built-in ones for _both_, which is potentially surprising. The logic should be refactored so that defaulting behaviour is isolated per schema.
6. Our built-in `openapi-3-schema.json` appears to predate the official version maintained http://github.com/OAI/OpenAPI-Specification. We ought to resync our versions with theirs.
7. When using configured schemas we remove the `#/id` node to avoid internal references being resolved to an external location and redownload, but we fail to apply the same logic for the built-in schemas leading those internal references to be resolved externally (and failing if the `#/id` refers to an out of date URL). We should remove the `#/id#` node from the built-in schemas to avoid resolving anything externally.

Faced with the same circumstance again:

- Fixing 1 would have meant that we saw a faulty validation message rather than a 500 response from the server. People would have been able to work around the problem by disabling or ignoring the rule.
- Fixing 4 would have meant that we got a clear failure at startup, rather than limping along in a broken state. Admins could have been alerted to the problem more quickly.
- Fixing either 6 or 7 would avoid the problem at all, without needing to add the redirects we hurried in as a quick fix.
- Fixing the rest could have made the code simpler and easier to follow, speeding up the fix.

Any objections?

roxspring:
- Based on #1124 since I was getting failed tests otherwise
- `JsonSchemaValidator` defaults to empty violation pointer
- Use official schema content and references from OAI/OpenAPI-Specification
- Fall back to built-in schemas independently per schema version
- Avoid resolving against id for built-in schemas
- Fail fast on invalid schema whether internal or external
- Use internal built-in schemas by default, not external URLs

Closes #1125


zeitlinger:
:+1:

tkrop:
:+1:

tkrop:
@roxspring great improvement!

Issue #964, opened 2019-03-28
deploy the jars on central repo (Maven or other)


vincent-fuchs:
Hi,

One thing that I feel would help people to experiment various things with Zally would be to publish the jars on a central repository (Maven central or another, doesn't make much difference - but I see there's already a zalando namespace existing on https://mvnrepository.com/artifact/org.zalando )

The advantage of such approach is that we can reuse very easily all Zally components in a different context. It's the approach I've followed on some of my projects, (see https://github.com/societe-generale/github-crawler#extending-the-crawler-and-contributing-to-it- for an example) : for instance, I could reuse part of your github-integration components into my existing custom webhooks.

Also, that would be one way of implementing https://github.com/zalando/zally/issues/560 to use easily custom rules : start the application "from the imported jar" , use Spring Boot conditional beans to load Zalando and/or Zally rules by default, if no other RuleSet is provided in the config. 
If I want to use my own RuleSet, then I instanciate it (potentially mixing some out of the box rules from the jar, with mine that are in a separate jar or declared in my custom project) - similar to what I do in this [custom project](https://github.com/vincent-fuchs/my-custom-github-crawler/blob/master/src/main/java/com/github/vincent_fuchs/PersonalGitHubCrawlerApplication.java) : using the jar of the "official" project mentionned previously, I instanciate a customOutput that takes precendence on the default one.
And the day you publish a new Zally version, all I need to do is to upgrade the Zally dependency version on my side.

I've done it in a couple of projects using Maven and Travis-CI, where the jars get deployed when the code is tagged. It's a good opportunity for me to learn to do the same with Gradle, so let me know if you would you be interested in a PR implementing that change, I can probably take care of it !

Thanks

# Notes

- [x] Need #1157 first

natrem:
I propose to follow Vincent's example on his projects:
- update the [build.gradle](https://github.com/zalando/zally/blob/master/server/build.gradle.kts) to declare the release repositories
- add a few secrets in Travis (maven central credentials and signature) --> your job ;)
- perform the build in 2 steps:
    1. equivalent of `mvn release:prepare` that builds the project and add a tag
    2. let Travis react on tag to call equivalent of `mvn release:perform` 

There are other examples available online but for a project with "not-too-frequent" releases this allows control and ease of use once it's set.

@roxspring @maxim-tschumak @zeitlinger would you be ok with that?

roxspring:
[...]

PR #1131, opened 2020-02-25
Sign `zally` artifacts

vadeg:
* Add [signing plugin](https://docs.gradle.org/current/userguide/signing_plugin.htm) which signs artifacts before publishing to Maven Central
* Update `readme` with information how to configure publish step

Related to #964


zeitlinger:
:+1:

tkrop:
:+1:

Issue #1134, opened 2020-03-18
Easier local development and debugging of zally server


GoWind:
Currently, the default build tasks that are executed by the `build-and-run.sh` script takes a lot of time to complete and bring up a zally instance.
This becomes a problem especially when one needs to modify only one of the components (server/web-ui) and receive fast feedback during development.
For the `server` component especially, the default build does a lot of tasks which takes in the order of minutes to complete.
This slows down my development style a lot. 
I would like to  add some helper scripts to the server to make the dev/debug cycle quicker by:

a) Be able to run the server without needing to resort to `docker-compose` and 
b) run only the `bootRun` tasks, which starts the application server (takes about 50 secs instead of mins on a laptop) and additionally
c) bring up the web frontend without needing to use `docker-compose`.

roxspring:
The `build-and-run.sh` script is mostly aimed at getting up and running quickly with minimal dependencies on the host system, running against a PostgreSQL database for maximum realism. When developing you generally need to have the relevant development toolchains installed (JDK, NodeJS, Go) and have little need to use docker at all, other than final manual checks.

In particular, if you need to quickly run a development version of the server then your best bet is to just to run `./gradlew bootRun` - it'll rebuild any modified server code and run the instance against an internal H2 database, so doesn't even need PostgreSQL or any environment variables set up. Does it need to be made simpler than that?

vadeg:
I will close it for now, because there is no need in additional script.
1. Postgres database can run using `docker-compose run postgres` command
2. Zally server can start using `./gradlew bootRun`

Feel free to reopen it if needed.

PR #1136, opened 2020-03-24
Make it easy to start the server standalone. Refs #1134

GoWind:
Currently, the server is started as a part of the `docker-compose`
command, which downloads docker images and runs a lot of other gradle
tasks, lengthening the development cycle.
Add a new script which will run the pre-requisite postgres db as a
docker image, expose the DB Port to the application and run the
`bootRun` task, which starts the server.

The `web-ui` can be run separately to use the `web-ui` for linting APIs

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,8 +40,11 @@ and database) by executing:
 ```bash
 ./build-and-run.sh
 ```
-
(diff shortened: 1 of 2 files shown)


vadeg:
`docker-compose run postgres` will run only `postgres` service from the `docker-compose` file.

PR #1140, opened 2020-04-20
fix(server): Revert openapi3-schema.json to earlier content

### LINKED ISSUE #1138: OpenAPI3 linting bug when validating `url` field under `contact`
In Pull Request https://github.com/zalando/zally/pull/1126/files, I noticed that the openapi3 spec was updated to the most recent published version, as the previous spec document that was being used was a draft version.

The `openapi3_schema.json` introduced in the PR, replaced an existing `openapi-3-schema.json` and the `format` of the `url` field changed from `url` to `uri-reference` when the schema was updated.

It looks like due to that change, when linting API specs, the linting fails and the web-UI displays the following violation.

I did a manual test, by updating the `format` of the `url` field to `url` instead of `uri-reference` in the new `openapi3-schema.json` file and Zally successfully validates the API locally.

```
VIOLATIONS
MUST – Provide API Specification using OpenAPI
Does not match openapi3 schema: Format attribute "uri-reference" not supported

Rule: https://zalando.github.io/restful-api-guidelines/#101

Paths:

```
You can reproduce the bug with the following API spec mentioned below

For validating the schema, it looks we use the 
`https://github.com/java-json-tools/json-schema-validator`

My initial assumption was that this was because there is no support for the format `uri-reference` in the validator library.

But on closer inspection, I found out that field `url` under `oauth2flows -> clientCredentials` also has the format type `uri-reference` and the validator doesn't throw an error when  we add a `clientCredentials` in the API definition.

So either a) there is a bug where the `url` field in the `contact` alone doesn't support the `uri-reference` format or 
b) we are not validating the `oauth2flows` part of the API (and possibly other sections),when we upload an API definition for linting.

```
openapi: 3.0.0
info:
  x-api-id: 4ab85265-7e3a-4371-a549-3e70d67f8afa
  title: Chronicles of Amber
  description: A fight amongs the brothers for the throne
  version: 1.0.0
  x-audience: component-internal
  contact:
    name: Oberon
    email: Oberon@amber.com
    url: http://king.amber.com
paths:
  /sites/{site_name}/fetch-me-the-thrones:
    post:
      description: >
        Demand giving up the throne for the person making the request
        There can be only one liege at a time
      security:
        - oauth2:
            - amber.write
      parameters:
        - name: site_name
          in: path
          required: true
          schema:
            type: string
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/format_input'
      responses:
        '200':
          description: Your request for the throne has been accepted
          content:
            application/json:
              schema:
                properties:
                  official_title:
                    type: string
                    description: Title as a king
        '400':
          description: Invalid request.
        '409':
          description: A war is already ongoing, another one cannot be started.

 

servers:
  - url: amber.zalan.do
    description: Production
  - url: amber.shadow.zalan.do
    description: Staging
components:
  schemas:
    format_results:
      type: array
      items:
        type: array
        items:
          type: string
          description: item ID

    format_input:
      required:
        - who
      properties:
        who:
          type: string
          

    
  securitySchemes:
    oauth2:
      type: oauth2
      flows:
        clientCredentials:
          tokenUrl: hakka
          scopes:
            amber.write : Access right needed to demand a throne.
``

roxspring:
- Revert openapi3-schema.json to prior content as discussed in #1138.

Closes #1138

### DIFF
--- a/server/zally-core/src/main/resources/schemas/openapi3-schema.json
+++ b/server/zally-core/src/main/resources/schemas/openapi3-schema.json
@@ -1,74 +1,72 @@
 {
-  "id": "https://spec.openapis.org/oas/3.0/schema/2019-04-02",
+  "title": "A JSON Schema for OpenAPI 3.0.",
+  "id": "http://openapis.org/v3/schema.json#",
   "$schema": "http://json-schema.org/draft-04/schema#",
-  "description": "Validation schema for OpenAPI Specification 3.0.X.",
   "type": "object",
+  "description": "This is the root document object of the OpenAPI document.",
   "required": [
     "openapi",
     "info",
     "paths"
   ],
+  "additionalProperties": false,
+  "patternProperties": {
+    "^x-": {
+      "$ref": "#/definitions/specificationExtension"
+    }
+  },
   "properties": {
     "openapi": {
-      "type": "string",
-      "pattern": "^3\\.0\\.\\d(-.+)?$"
+      "type": "string"
     },
     "info": {
-      "$ref": "#/definitions/Info"
-    },
-    "externalDocs": {
-      "$ref": "#/definitions/ExternalDocumentation"
+      "$ref": "#/definitions/info"
     },
     "servers": {
       "type": "array",
       "items": {
-        "$ref": "#/definitions/Server"
-      }
-    },
-    "security": {
-      "type": "array",
-      "items": {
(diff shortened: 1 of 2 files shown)


tkrop:
:+1:

zeitlinger:
:+1:

GoWind:
:+1:

PR #1141, opened 2020-04-20
fix(server): StringPropertyLengthBoundsRule ignores enum strings #1137

### LINKED ISSUE #1137: Don't report StringPropertyLengthBoundsRule if property has enum
When a property of type string has an enum setting it does not make sense to specify minimum and maximum length. 

Reproduce with an object in a body or response like this:
```yaml
type: object
properties:
  fooBar:
    type: string
    enum: [foo, bar]
```

Gets complaint from rule S007 SHOULD Define bounds for lengths of string properties 

Expected: 
- Don’t insist on adding `minLength` and `maxLength`
- In fact it might make sense to complain if they are present

roxspring:
- No violations created when string property is an enum

Closes #1137

### DIFF
--- a/server/zally-ruleset-zally/src/test/kotlin/de/zalando/zally/ruleset/zally/StringPropertyLengthBoundsRuleTest.kt
+++ b/server/zally-ruleset-zally/src/test/kotlin/de/zalando/zally/ruleset/zally/StringPropertyLengthBoundsRuleTest.kt
@@ -280,4 +280,31 @@ class StringPropertyLengthBoundsRuleTest {
             .assertThat(violations)
             .isEmpty()
     }
+
+    @Test
+    fun `checkStringLengthBounds with enum returns no violations`() {
+        @Language("YAML")
+        val context = DefaultContextFactory().getOpenApiContext(
+            """
+            openapi: 3.0.2
+            info:
+              title: Thing API
+              version: 1.0.0
+            components:
+              schemas:
+                Thing:
+                  type: object
+                  properties:
+                    theString:
+                      type: string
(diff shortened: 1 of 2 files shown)


vadeg:
Thank @roxspring!

vadeg:
👍

server/zally-ruleset-zally/src/test/kotlin/de/zalando/zally/ruleset/zally/StringPropertyLengthBoundsRuleTest.kt
```diff
@@ -280,4 +280,31 @@ class StringPropertyLengthBoundsRuleTest {
             .assertThat(violations)
             .isEmpty()
     }
+
+    @Test
+    fun `checkStringLengthBounds with enum returns no violations`() {
+        @Language("YAML")
+        val context = DefaultContextFactory().getOpenApiContext(
+            """
```
zeitlinger, line 288: there is no min/max length in the test

roxspring: That’s the intention - In the case of enums the min/max are implicitly defined by the shortest/longest accepted enum so requiring explicit min/max seems redundant.

zeitlinger: ok, that was a bit indirect - but now I get it.

zeitlinger:
:+1:

PR #1145, opened 2020-05-11
Scripts for migrating Database

GoWind:
The production environment for zally has been updated to run in a new
cluster. To support that, we need to migrate the existing data in the
api_review and rule_violation tables to a new Database.
Checking in the script application that will download the data to the
local machine for migrating it into a newer database

### DIFF
--- a/migration/README.md
+++ b/migration/README.md
@@ -1,0 +1,6 @@
+### Running the script 
+
+You will need Java 8+ and [Leiningen](leiningen.org) to run the script.
+
+Use `lein run table-name` to run the script. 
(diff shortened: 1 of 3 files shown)


vadeg:
I don't think it is a good idea to add Clojure and its tooling here. Zally uses `Flyway` for migrations. 
If it is some sort of one-shot task, then probably it should be in a separate repo. Zally code on GitHub is not a good place for that.

tkrop:
This was designed for a very special situation. I think it makes no sense to merge this.

PR #1151, opened 2020-06-10
feat: Working docker-compose healthchecks

### LINKED ISSUE #1133: Invalid Docker Compose Healthcheck
The `docker-compose.yaml` defines a fairly reasonable health check based on the SpringBoot management endpoint, running on port 7979. The `MANAGEMENT_PORT` environment variable is also set to 7979, which `application.properties` uses to control the management port used by the server.

The `docker-compose.yaml` also defines `SPRING_PROFILES_ACTIVE=dev` making the server use `application-dev.properties`, which [hardcodes the management port to 8000](https://github.com/zalando/zally/blob/master/server/src/main/resources/application-dev.yml#L27). It's not clear to me why the dev profile should force a particular management port, but there were clearly many attempts to make that the case in the early days.

As a result, when running via the default docker-compose configuration, the health check can never succeed and so the server is always marked "unhealthy".

I'm very tempted to "fix" this by dropping all reference to the management port from the dev profile... if anybody can explain why this is a bad idea then please do so!

(Obviously, another "fix" option is to remove reference to health checks from the docker-compose configuration. This seems like a backwards step to me, but am open to persuasion.)

roxspring:
Listing docker compose containers now shows healthy state:
```
      Name                    Command                  State                    Ports              
---------------------------------------------------------------------------------------------------
zally_postgres_1   docker-entrypoint.sh postgres    Up (healthy)   0.0.0.0:54321->5432/tcp         
zally_server_1     /bin/sh -c java $(java-dyn ...   Up (healthy)   0.0.0.0:8000->8000/tcp, 8080/tcp
zally_web-ui_1     node server.js                   Up (healthy)   0.0.0.0:8080->3000/tcp          
```

Closes #1133

### DIFF
--- a/server/zally-server/src/main/resources/application.yml
+++ b/server/zally-server/src/main/resources/application.yml
@@ -28,7 +28,7 @@ management:
       exposure:
         include: health, metrics
   server:
-    port: ${MANAGEMENT_PORT}
+    port: ${MANAGEMENT_PORT:7979}
 
 twintip:
   mapping: /api
@@ -45,4 +45,3 @@ zally:
 ---
 spring.profiles: local
 TOKEN_INFO_URI: https://auth.example.com/oauth2/tokeninfo
-MANAGEMENT_PORT: 7979
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -18,6 +18,12 @@ services:
     - server:server.local
     ports:
     - "8080:3000"
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:3000"]
+      interval: 5s
+      timeout: 2s
+      retries: 10
+
   server:
     build: ./server
     depends_on:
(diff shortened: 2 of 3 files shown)


tkrop:
:+1:

roxspring:
(I have noticed the CI failure btw - will try and get back to this soon if nobody beats me to it!)

vadeg:
👍

server/zally-server/src/main/resources/application.yml
```diff
@@ -18,7 +18,7 @@ spring:
             non_contextual_creation: true
   oauth2:
     resource:
-      tokenInfoUri: ${TOKEN_INFO_URI}
+      tokenInfoUri: ${TOKEN_INFO_URI:https://auth.example.com/oauth2/tokeninfo}
```
vadeg, line 21: I don't think it is safe to point to some auth service by default. Users can be just unaware that it must be configured and will send their data to unknown service.

roxspring: Fair point. Switched to just defaulting `MANAGEMENT_PORT`.

vadeg:
👍

zeitlinger:
:+1:

Issue #1161, opened 2020-07-01
Adding Prometheus metrics to Zally server


WilliamTwill:
Hi there, for our own dev stack we wanted to scrape metrics into our grafana dashboards showing historical information on Violations for each individual API.

I have already written the code needed to expose the prometheus endpoint as well as some code to expose violation metrics.

zeitlinger:
it looks like this is similar to `/review-statistics`. Could you just use that?

WilliamTwill:
Similar but different in that `/review-statistics` returns an _aggregated_ view of all api reviews.
I could modify the existing endpoint to be more detailed (so show stats for api reviews grouped by api name for instance). 
However exposing the prometheus endpoint using micrometer is also important since we want to be able to scrape metrics from it and make dashboards in visualization tools like grafana.

zeitlinger:
OK, I guess that makes sense.

the endpoint also needs to be added to OAuthConfiguration - otherwise it can't be accessed.

@vadeg can you also check this issue?

tkrop:
@WilliamTwill Thanks for your great contribution, and sorry for the bad support in the last month. I like your request and think we should merge it. However, before I would like to understand your `label`-approach. Can you please elaborate on how reviews are supposed to be labeled and what do you exactly want to achieve with that?

Carnifexje:
> @WilliamTwill Thanks for your great contribution, and sorry for the bad support in the last month. I like your request and think we should merge it. However, before I would like to understand your `label`-approach. Can you please elaborate on how reviews are supposed to be labeled and what do you exactly want to achieve with that?

So the change is two-fold.

1. Exposure of Prometheus metrics. In general this means Zally can be monitored by prometheus (and visualized with for example Grafana). By using these metrics, we can for example auto-scale Zally server in a kubernetes cluster based on the usage metrics coming from Prometheus if there are many teams validating their API changes at the same time.

2. Custom labels. As of right now, my change will allow the passing of a key-value pair when requesting a validation. These labels are also exposed within the prometheus metrics. In our specific case we pass the git branch (feature/x, master, etc) as a key-value pair ({"branch": "feature/x"}) so that we can distinguish validations in the database (and thus the exposed prometheus metrics).

[...]

PR #1162, opened 2020-07-01
Implementation for #1161

WilliamTwill:
Implementation for #1161

### DIFF
--- a/server/zally-server/build.gradle.kts
+++ b/server/zally-server/build.gradle.kts
@@ -43,6 +43,7 @@ dependencies {
     compile("org.zalando.stups:stups-spring-oauth2-server:1.0.22")
     compile("org.zalando:problem-spring-web:0.23.0")
     compile("org.zalando:twintip-spring-web:1.1.0")
+    compile("io.micrometer:micrometer-registry-prometheus")
 
     testCompile(project(":zally-test"))
     testCompile("net.jadler:jadler-core:$jadlerVersion")
--- a/server/zally-server/build.gradle.kts
+++ b/server/zally-server/build.gradle.kts
@@ -43,7 +43,7 @@ dependencies {
     compile("org.zalando.stups:stups-spring-oauth2-server:1.0.22")
     compile("org.zalando:problem-spring-web:0.23.0")
     compile("org.zalando:twintip-spring-web:1.1.0")
-    compile("io.micrometer:micrometer-registry-prometheus")
+    compile("io.micrometer:micrometer-registry-prometheus:1.0.6")
 
     testCompile(project(":zally-test"))
     testCompile("net.jadler:jadler-core:$jadlerVersion")
--- a/server/zally-server/src/main/java/org/zalando/zally/apireview/ApiReviewRepository.kt
+++ b/server/zally-server/src/main/java/org/zalando/zally/apireview/ApiReviewRepository.kt
@@ -34,7 +34,12 @@ interface ApiReviewRepository : CrudRepository<ApiReview, Long> {
         """
         SELECT r
         FROM org.zalando.zally.apireview.ApiReview r
-        WHERE (r.name, r.created) IN (SELECT name, MAX(created) FROM org.zalando.zally.apireview.ApiReview GROUP BY name)
+        WHERE (r.name, r.created) IN (
+            SELECT a.name, MAX(a.created) 
+            FROM org.zalando.zally.apireview.ApiReview a
+            LEFT OUTER JOIN a.customLabels l
+            GROUP BY a.name, INDEX(l), l
+        )
     """
     )
     fun findLatestApiReviews(): List<ApiReview>
--- a/server/zally-server/src/main/java/org/zalando/zally/apireview/ApiReviewRepository.kt
+++ b/server/zally-server/src/main/java/org/zalando/zally/apireview/ApiReviewRepository.kt
@@ -5,7 +5,7 @@ import org.springframework.data.repository.CrudRepository
 import org.springframework.data.repository.query.Param
 import org.zalando.zally.statistic.ReviewStatistics
 import java.time.LocalDate
-import java.util.*
+import java.util.UUID
 
 interface ApiReviewRepository : CrudRepository<ApiReview, Long> {
 
--- a/server/zally-server/src/main/java/org/zalando/zally/apireview/ApiReviewRepository.kt
+++ b/server/zally-server/src/main/java/org/zalando/zally/apireview/ApiReviewRepository.kt
@@ -1,11 +1,11 @@
 package org.zalando.zally.apireview
 
-import org.zalando.zally.statistic.ReviewStatistics
 import org.springframework.data.jpa.repository.Query
 import org.springframework.data.repository.CrudRepository
 import org.springframework.data.repository.query.Param
+import org.zalando.zally.statistic.ReviewStatistics
 import java.time.LocalDate
-import java.util.UUID
+import java.util.*
 
 interface ApiReviewRepository : CrudRepository<ApiReview, Long> {
 
@@ -29,6 +29,15 @@ interface ApiReviewRepository : CrudRepository<ApiReview, Long> {
         @Param("to") to: LocalDate,
         @Param("userAgent") userAgent: String = "%"
     ): ReviewStatistics
+
+    @Query(
+        """
+        SELECT r
+        FROM org.zalando.zally.apireview.ApiReview r
+        WHERE (r.name, r.created) IN (SELECT name, MAX(created) FROM org.zalando.zally.apireview.ApiReview GROUP BY name)
+    """
+    )
+    fun findLatestApiReviews(): List<ApiReview>
 
     /**
      * Find ApiReview instance by it's externalId UUID.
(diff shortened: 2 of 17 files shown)


zeitlinger:
:+1:

zeitlinger:
:+1:

server/zally-server/src/main/java/org/zalando/zally/statistic/ReviewMetrics.kt
```diff
@@ -0,0 +1,81 @@
...
+                Pair(MAY_VIOLATIONS, AtomicLong(statistic.mayViolations)),
+                Pair(HINT_VIOLATIONS, AtomicLong(statistic.hintViolations))
+            )
+        ))
+    }
+
+    private fun registerGaugeMetric(statisticReference: Map<String, Map<String, AtomicLong>>) {
+        statisticReference.entries.forEach { entry ->
+            entry.value.forEach { metric ->
+                val snakeCasedApiName = entry.key.replace(Regex("\\p{Zs}+"), "_").toLowerCase()
```
vadeg, line 62: Why do you need that?

server/zally-server/src/main/java/org/zalando/zally/statistic/ReviewMetrics.kt
```diff
@@ -0,0 +1,81 @@
...
+        LOG.debug("Updating metrics for review statistics")
+        apiReviewRepository.findLatestApiReviews()
+            .map(ReviewStatisticsByName.Companion::of)
+            .forEach { statistic ->
+                statisticsReferences
+                    .filter { it.containsKey(statistic.name) }
+                    .map { it[statistic.name] }
+                    .map {
+                        LOG.debug("Updating metrics reference values for review statistic ${statistic.name}")
+                        it?.get(NUMBER_OF_ENDPOINTS)?.set(statistic.numberOfEndpoints)
```
vadeg, line 31: I think if would be more readable to store `ReviewStatisticsByName` as a map value and expose it to multiple gauges. It would help to avoid crazy constructions like this `<Map<String, Map<String, AtomicLong>>>`.

vadeg (commented):
Thank you for the PR. I left some comments.

server/zally-server/src/main/java/org/zalando/zally/statistic/ReviewMetrics.kt
WilliamTwill: Yes I realize this can be done in a much better and readable way. I will go back to the drawing board and take a more robust approach.
This was an initial version which I should have improved a bit more to be honest.

server/zally-server/src/main/java/org/zalando/zally/statistic/ReviewMetrics.kt
WilliamTwill: I added a code comment, it's to make sure there is no whitespaces in metric names because those are illegal characters.

server/zally-server/src/main/java/org/zalando/zally/apireview/ApiReviewRepository.kt
```diff
@@ -30,6 +30,15 @@ interface ApiReviewRepository : CrudRepository<ApiReview, Long> {
         @Param("userAgent") userAgent: String = "%"
     ): ReviewStatistics
 
+    @Query(
+        """
+        SELECT r
```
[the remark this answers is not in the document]
WilliamTwill: Before yes, but now I need to join custom labels so this is needed.

server/zally-server/build.gradle.kts
```diff
@@ -43,6 +43,7 @@ dependencies {
     compile("org.zalando.stups:stups-spring-oauth2-server:1.0.22")
     compile("org.zalando:problem-spring-web:0.23.0")
     compile("org.zalando:twintip-spring-web:1.1.0")
+    compile("io.micrometer:micrometer-registry-prometheus")
```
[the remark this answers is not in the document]
WilliamTwill: Done

server/zally-server/src/main/java/org/zalando/zally/statistic/ReviewMetrics.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.scheduling.annotation.Scheduled
+import org.springframework.stereotype.Component
+import org.zalando.zally.apireview.ApiReviewRepository
+import java.util.concurrent.atomic.AtomicLong
+
+@Component
+class ReviewMetrics(private val apiReviewRepository: ApiReviewRepository, private val meterRegistry: MeterRegistry) {
+
+    final val statisticsReferences = mutableListOf<Map<String, Map<String, AtomicLong>>>()
```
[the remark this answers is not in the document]
WilliamTwill: Created abstractions for this as it was confusing indeed.

tkrop:
@WilliamTwill Thanks for your great contribution, and sorry for the bad support in the last month. I like your request and think we should merge it. However, before I would like to understand your `label`-approach. Can you please elaborate on how reviews are supposed to be labeled and what do you exactly want to achieve with that?

Carnifexje:
> @WilliamTwill Thanks for your great contribution, and sorry for the bad support in the last month. I like your request and think we should merge it. However, before I would like to understand your `label`-approach. Can you please elaborate on how reviews are supposed to be labeled and what do you exactly want to achieve with that?

@tkrop So the change is two-fold.

Exposure of Prometheus metrics. In general this means Zally can be monitored by prometheus (and visualized with for example Grafana). By using these metrics, we can for example auto-scale Zally server in a kubernetes cluster based on the usage metrics coming from Prometheus if there are many teams validating their API changes at the same time.

Custom labels. As of right now, my change will allow the passing of a key-value pair when requesting a validation. These labels are also exposed within the prometheus metrics. In our specific case we pass the git branch (feature/x, master, etc) as a key-value pair ({"branch": "feature/x"}) so that we can distinguish validations in the database (and thus the exposed prometheus metrics).

On top of these metrics we have built a Grafana dashboard for our dev teams which shows how many (must, should, may) violations their services have. Of course we also inject any violations coming back from Zally into merge requests directly as comments (as per #648 (comment)). So our devs can immediately spot violations during the dev cycle.
The reason we also publish the API violation results to a dashboard is to prove we work using high-quality standards to outside the teams.

Whyeasy:
Hi there,

Please wait with approving/merging this PR. We are having some performance issues with this PR and the database. I will come back with some data.

Whyeasy:
So after some searching, testing and trying different approaches this is what I came up with.

So the current query that is used to fetch the data to create the metrics (found [here](https://github.com/zalando/zally/pull/1162/files#diff-8551d27416b15902d3e260808c67e6253873482bb484d543aac111b6334e0811R33)), retrieves all the entries and labels inside of the custom_labels_mapping and the entries for it in the api_review tables.

This creates a heavy load on the database, but next to that it also creates a high cardinality of labels within Prometheus which in the end degrades performance of Prometheus.

So my idea to solve this is to add an extra WHERE IN clause in the subquery where you can specify which label values you want to monitor for metrics. This way it's scalable and the user can define by them self how much they want to monitor, but also determine how far they can stretch it with their database.

I will prepare this change tomorrow and add it to this PR.

Whyeasy:
Had a quick chat with @Carnifexje and in the end he implemented the change for me. 

I would like to confirm the change works and database is acting normal again. This also solves the cardinality for Prometheus.

tkrop:
Hi @WilliamTwill sorry for the long delay again.

@vadeg and I had a discussion about the implementation and agree that it definitely provides value for the community. We currently, in some way misuse the authorization to distinguish use cases by storing and utilizing the creator where a label would have been a cleaner approach. Still we are not sure about the complexity the additional table for the labels adds.

Wouldn't it be sufficient to just support a single label field in the review table? @vadeg has a suggestion he likes to explain.

Anyhow, one last think I'm currently missing in this pull request is to document the effective changes to the API. @WilliamTwill Can you please update the API specification in https://github.com/zalando/zally/blob/master/server/zally-server/src/main/resources/api/zally-api.yaml?

Thanks in advance

Tronje

vadeg:
@WilliamTwill thank you for your contribution and sorry for a very long delay. 

Zally uses JPA as an abstraction layer on top of database. For tests Zally uses [HyperSQL DB](http://hsqldb.org/), but in the docker containers it uses Postgres and migration scripts are written for Postgres. Despite the fact that in memory databases are good for tests, lack of features in these databases enforces to make some design decisions. In this case this decision is having a separate table for labels which I think can be done in a more elegant way.
Postgres supports [JSON data type](https://www.postgresql.org/docs/current/datatype-json.html) which can contain JSON object. In addition, it has many [JSON operators and functions](https://www.postgresql.org/docs/current/functions-json.html).  The combination of both fits perfectly this use case. Instead of a separate table for labels, labels can be stored together with reviews in a separate column.

If continue with this approach, then we need to remove HyperSQL DB from unit tests and replace this database with Postgres using [TestContainers](https://www.testcontainers.org/modules/databases/postgres/). 

The downside of this approach that it would not be possible to start Zally for development without a Postgres server running. In our current world with Docker this should not be a problem at all. 
The benefit of this solution is that we would remove a boundary which enforces us to use only generic SQL which fits for all databases and allow to use features of specific database.

vadeg:
Closed because of inactivity. Labels task is moved into a separate issue #1323

PR #1210, opened 2021-03-02
Fixes for #1209

### LINKED ISSUE #1209: Code smell: String literal "/paths" is being duplicated
I am running a fork of Zally attached to Sonarqube and there is a code smell because the string literal "/paths" is being duplicated in multiple places.

tharakadesilva:
Fixes #1209

### DIFF
--- a/server/zally-core/src/main/kotlin/org/zalando/zally/core/util/OpenApiSections.kt
+++ b/server/zally-core/src/main/kotlin/org/zalando/zally/core/util/OpenApiSections.kt
@@ -1,0 +1,7 @@
+package org.zalando.zally.core.util
+
+class OpenApiSections {
+    companion object {
+        const val PATHS = "/paths"
+    }
+}
--- a/server/zally-core/src/test/kotlin/org/zalando/zally/core/ast/ReverseAstTest.kt
+++ b/server/zally-core/src/test/kotlin/org/zalando/zally/core/ast/ReverseAstTest.kt
@@ -1,8 +1,6 @@
 package org.zalando.zally.core.ast
 
 import com.fasterxml.jackson.core.JsonPointer
-import org.zalando.zally.core.ObjectTreeReader
-import org.zalando.zally.core.toJsonPointer
 import io.swagger.parser.OpenAPIParser
 import io.swagger.parser.SwaggerParser
 import io.swagger.util.Json
@@ -12,6 +10,9 @@ import org.apache.commons.io.IOUtils
 import org.assertj.core.api.Assertions.assertThat
 import org.intellij.lang.annotations.Language
 import org.junit.Test
+import org.zalando.zally.core.ObjectTreeReader
+import org.zalando.zally.core.toJsonPointer
+import org.zalando.zally.core.util.OpenApiSections.Companion.PATHS
 
 @Suppress("UndocumentedPublicClass", "UndocumentedPublicFunction", "StringLiteralDuplication")
 class ReverseAstTest {
@@ -77,7 +78,7 @@ class ReverseAstTest {
         assertThat(ast.isIgnored(pointer, "*")).isTrue()
         assertThat(ast.getIgnoreValues(pointer)).hasSize(1).contains("*")
 
-        pointer = "/paths".toJsonPointer()
+        pointer = PATHS.toJsonPointer()
         assertThat(ast.isIgnored(pointer, "*")).isFalse()
         assertThat(ast.getIgnoreValues(pointer)).isEmpty()
 
@@ -141,7 +142,7 @@ class ReverseAstTest {
         assertThat(ast.isIgnored(pointer, "*")).isTrue()
         assertThat(ast.getIgnoreValues(pointer)).hasSize(1).contains("*")
 
-        pointer = "/paths".toJsonPointer()
+        pointer = PATHS.toJsonPointer()
         assertThat(ast.isIgnored(pointer, "*")).isFalse()
         assertThat(ast.getIgnoreValues(pointer)).isEmpty()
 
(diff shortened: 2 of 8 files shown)


tkrop:
:+1:

server/zally-core/src/main/kotlin/org/zalando/zally/core/ast/ReverseAst.kt
```diff
@@ -48,6 +49,9 @@ class ReverseAst internal constructor(
     }
 
     companion object {
+        const val pathsJsonPointerString = "/paths"
```
vadeg, line 52: Please follow [kotlin guidelines](https://kotlinlang.org/docs/coding-conventions.html#property-names) for naming.

server/zally-core/src/main/kotlin/org/zalando/zally/core/ast/ReverseAst.kt
```diff
@@ -48,6 +49,9 @@ class ReverseAst internal constructor(
...
+        val pathsJsonPointer = pathsJsonPointerString.toJsonPointer()
```
vadeg, line 53: Same here about Kotlin guidelines.

vadeg (changes requested):
Hi @tharakadesilva !

Thank you for your contribution, but I don't think that `ReverseAst` is the right place to define `paths` property. This is a part of OpenAPI / Swagger spec. I think it should be somewhere in the `Context` or as an extension for `OpenAPI` or `Swagger` classes.

server/zally-core/src/main/kotlin/org/zalando/zally/core/ast/ReverseAst.kt
tharakadesilva: Fixed.

server/zally-core/src/main/kotlin/org/zalando/zally/core/ast/ReverseAst.kt
tharakadesilva: Fixed.

tharakadesilva:
> Hi @tharakadesilva !
> 
> Thank you for your contribution, but I don't think that `ReverseAst` is the right place to define `paths` property. This is a part of OpenAPI / Swagger spec. I think it should be somewhere in the `Context` or as an extension for `OpenAPI` or `Swagger` classes.

Hi @vadeg, thank you for the feedback. I am a little torn on where to move these. I feel that the recommendation to move it as an extension to `OpenAPI` is the best approach. `OpenAPI` already has some extensions in the `OpenApiUtil.kt` file. Do you think that this might be a good place to place these values as well?

vadeg:
@tharakadesilva I would suggest something like this
```
class OpenApiPaths {
    companion object {
        const val PATHS = "/paths"
    }
}
```
Unfortunately, Kotlin [doesn't support](https://youtrack.jetbrains.com/issue/KT-11968) static extensions for Java classes and the original idea to make someting like 
```
io.swagger.v3.oas.models.Paths
val Paths.paths: String 
    get() = "sdafa"
```
and use like `Paths.paths` will not work.

tharakadesilva:
@vadeg I have fixed the code as you suggested. Would appreciate it if you could review it again now.

server/zally-core/src/main/kotlin/org/zalando/zally/core/util/OpenApiPaths.kt
```diff
@@ -0,0 +1,10 @@
+package org.zalando.zally.core.util
+
+import org.zalando.zally.core.toJsonPointer
+
+class OpenApiPaths {
+    companion object {
+        const val PATHS_JSON_POINTER_STRING = "/paths"
+        val PATHS_JSON_POINTER = PATHS_JSON_POINTER_STRING.toJsonPointer()
```
vadeg, line 8: I thinks this is not needed here, because it is up to the client how to interpret the string.

server/zally-core/src/main/kotlin/org/zalando/zally/core/util/OpenApiPaths.kt
```diff
@@ -0,0 +1,10 @@
+package org.zalando.zally.core.util
+
+import org.zalando.zally.core.toJsonPointer
+
+class OpenApiPaths {
+    companion object {
+        const val PATHS_JSON_POINTER_STRING = "/paths"
```
vadeg, line 7: This is not a JSON pointer but a json path. Could you rename it to `PATHS`?

vadeg (commented):
Hi @tharakadesilva 

Sorry for the delay. I have left a couple of comments.

server/zally-core/src/main/kotlin/org/zalando/zally/core/util/OpenApiPaths.kt
tharakadesilva: I have removed this.

server/zally-core/src/main/kotlin/org/zalando/zally/core/util/OpenApiPaths.kt
tharakadesilva: @vadeg I renamed this to `PATHS`. However, now I feel that the class name `OpenApiPaths` might not appropriate anymore. I was thinking about renaming `OpenApiSections`, what do you think?

vadeg: Good idea.

tharakadesilva: Fixed.

tharakadesilva:
> Hi @tharakadesilva
> 
> Sorry for the delay. I have left a couple of comments.

Hi @vadeg, thank you for the feedback. I have addressed the comments and I think it is ready for a review again. Thank you for the continuous feedback :)

vadeg (approved):
@tharakadesilva thank you very much for your contribution!

vadeg:
👍

tkrop:
:+1:

PR #1222, opened 2021-04-10
Add Fashion Platform GW headers

vadeg:
Fashion Platform GW headers and exclude `X-Zalando-Request-URI` from the [Rule 166](https://opensource.zalando.com/restful-api-guidelines/#166), 

Fixes #1216

Issue #168, opened 2017-03-14
NPE in SuccessResponseAsJsonObject rule


fmueller:
With our swagger file in examples/ folder, server throws a null pointer exception and crashes.

PR #1225, opened 2021-04-12
Add Rule 127 - Use JSON as payload data interchange format

### LINKED ISSUE #1217: Editorial updates around Standard Data Format rules
Please check if the PR https://github.com/zalando/restful-api-guidelines/pull/643 introduces changes which are relevant to the Zally project.

- [ ] Add support of 10 [IANA](https://www.iana.org/assignments/media-types/media-types.xhtml) formats.
- [ ] Generate configuration with allowed formats for the rule from the list
- [ ]

vadeg:
Implement [Rule 167](https://opensource.zalando.com/restful-api-guidelines/#167)

- [ ] Add support for the acceptance list of media types

Fixes #1217

### DIFF
--- a/server/zally-core/src/main/kotlin/org/zalando/zally/core/MediaType.kt
+++ b/server/zally-core/src/main/kotlin/org/zalando/zally/core/MediaType.kt
@@ -1,0 +1,5 @@
+package org.zalando.zally.core
+
+enum class MediaType(val type: String) {
+    APPLICATION_JSON("application/json")
+}
--- a/server/zally-core/src/main/kotlin/org/zalando/zally/core/util/OpenApiUtil.kt
+++ b/server/zally-core/src/main/kotlin/org/zalando/zally/core/util/OpenApiUtil.kt
@@ -1,6 +1,8 @@
 package org.zalando.zally.core.util
 
 import io.swagger.v3.oas.models.OpenAPI
+import io.swagger.v3.oas.models.Operation
+import io.swagger.v3.oas.models.PathItem
 import io.swagger.v3.oas.models.media.Schema
 import io.swagger.v3.oas.models.parameters.Parameter
 import io.swagger.v3.oas.models.responses.ApiResponse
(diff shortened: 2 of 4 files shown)


tkrop:
I'm not sure, but may be this implementation is going to far, since we explicitly relax this rule in [#168](https://opensource.zalando.com/restful-api-guidelines/#168).

vadeg:
@tkrop thanks for the comment. My impression was that in these cases API designers have to exclude Rule 167 explicitly. From API perspective it is hard to distinguish where should be JSON and where something else.

tkrop:
@vadeg Yes, I agree. Excluding rules is definitely an option. However, currently we do not encourage the usage of `x-zally-ignore` to prevent users from freely applying it to other rules too. I think we should carefully consider the impact of such a strategy change.

vadeg:
Let's discuss this on our meeting.

vadeg:
We decided not to implement this rule, because this would lead many exceptions and it would be difficult to validate each case.

tkrop:
May be you can separate https://github.com/zalando/zally/pull/1225/commits/38da19aefde6694eff5df00984c8ac819a8e15de, so that we can merge this quickly.

tkrop:
The build seems to be failing because of the same reason as in https://github.com/zalando/zally/pull/1227:

```
--- FAIL: TestIntegrationWithNoMustViolations (0.36s)
    integration_test.go:84: 
        	Error Trace:	integration_test.go:84
        	Error:      	Should be zero, but was 17
        	Test:       	TestIntegrationWithNoMustViolations
        	Messages:   	No MUST violations expected
    integration_test.go:88: 
        	Error Trace:	integration_test.go:88
        	Error:      	Expected nil, but got: &domain.AppError{e:(*errors.errorString)(0xc000182c00), c:3}
        	Test:       	TestIntegrationWithNoMustViolations
```

vadeg:
> May be you can separate [38da19a](https://github.com/zalando/zally/commit/38da19aefde6694eff5df00984c8ac819a8e15de), so that we can merge this quickly.

@tkrop this has been merged in #1222 . I will update this PR to remove this change.

tfrauenstein:
hint: we discussed / decided the topic in the API guild meeting --> see https://docs.google.com/document/d/1NahSuMC0LRwFTTrln2m0iDLq0nHzGmF59Gv7xBhKPfQ/edit#bookmark=id.p63ihhuscg86

ePaul:
Instead of this, we want #1316.

PR #1252, opened 2021-06-07
Add validation of required path parameters

### LINKED ISSUE #459: New rules proposition
I've found a description of extra rules implemented by some swagger validator:
https://github.com/apigee-127/swagger-tools/blob/master/docs/Swagger_Validation.md

Most of them are not supported by Zally and schema validation does not handle them too:

1. Each operation should have only one parameter of type body
1. An operation cannot have a form or formData parameter if it has a body parameter
1. The items property is required for all schemas/definitions of type array.
1. Each defined operation path parameters must correspond to a named element in the API's path pattern. (For example, you cannot have a path parameter named id for the following path /pets/{petId} but you must have a path parameter named petId.)
1. Each API path should be unique. (For Swagger 1.2, this applies to both the Resource Listing and the API Declarations. For all versions, being unique is both based on verbatim equality and equivalency. Example: /pets/{id} and /pets/{petId} are equivalently the same but not the same verbatim.)
1. A definition/model cannot declare a property that is already defined by one of its ancestors.
As far as I understand this is about allOf schema compositions (the cursor property is duplicated)
```
  Page:
    type: object
    required:
      - notlisted
    properties:
      cursor:
        type: string
      href:
        type: string
  ProductPage:
    allOf:
    - "$ref": "#/definitions/Page"
    - type: object
      properties:
        cursor:
          type: string

```
7. A definition/model's ancestor cannot be a descendant of said model. (Circular Reference)
1. Each reference must point to an existing definition.
1. Each definition/model property listed in the required array must be defined in the properties of the model itself or one of its ancestors.

Lets create separated tickets for those which can be useful.

vadeg:
Add [path parameter validation](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#fixed-fields-10)

> Determines whether this parameter is mandatory. If the parameter location is "path", this property is REQUIRED and its value **MUST** be true. Otherwise, the property MAY be included and its default value is false.

Fixes #459


vadeg:
👍

server/zally-ruleset-zally/src/test/kotlin/org/zalando/zally/ruleset/zally/PathParameterRuleTest.kt
```diff
@@ -0,0 +1,80 @@
+package org.zalando.zally.ruleset.zally
+
+import org.intellij.lang.annotations.Language
+import org.junit.Test
+import org.zalando.zally.core.DefaultContextFactory
+import org.zalando.zally.test.ZallyAssertions.assertThat
+
+class PathParameterRuleTest {
```
tkrop, line 8: Let us add test for path parameters declared in `/components/parameters` too.

tkrop (commented):
Looks good, but I would let us cover all use cases with tests.

server/zally-ruleset-zally/src/test/kotlin/org/zalando/zally/ruleset/zally/PathParameterRuleTest.kt
vadeg: Added.

vadeg:
Waits #1258 to be merged.

tkrop (approved):

tkrop:
:+1:

vadeg:
👍

PR #1254, opened 2021-06-08
Migrate to Java 11 & Kotlin 1.4.x

vadeg:
Update dependencies

* Dokka to 1.4.32
* Kotlin to 1.4.32
* `jackson` to 2.12.2
* Migrate gradle scripts
* Fix dokka Metaspace error
* Move build to Java 11

Fixes #920


github-integration/build.gradle.kts
```diff
@@ -46,57 +46,57 @@ dependencies {
...
+    implementation("com.github.java-json-tools:json-schema-validator:2.2.8")
+    implementation("org.springframework.boot:spring-boot-starter-web")
+    implementation("org.springframework.boot:spring-boot-starter-thymeleaf")
+    implementation("org.springframework.boot:spring-boot-starter-undertow")
+    implementation("org.springframework.boot:spring-boot-starter-actuator")
+    implementation("org.springframework.boot:spring-boot-starter-data-jpa")
+    implementation("org.thymeleaf:thymeleaf-spring4")
+    implementation("org.flywaydb:flyway-core:4.1.2")
+    implementation("org.postgresql:postgresql:42.1.1")
+    implementation("org.hibernate:hibernate-core:$hibernateVersion")
```
cberg-zalando, line 60: In general, overwriting dependency version, which are managed by Spring, is not a good practice as it can lead to breakages. But I see, that you are still using Spring Boot 1.5.*, which might be the reason. Given that it is not maintained anymore, it might make sense to also look into switching to a supported version.

github-integration/build.gradle.kts
```diff
@@ -46,57 +46,57 @@ dependencies {
...
+    implementation("org.hibernate:hibernate-entitymanager:$hibernateVersion")
+    implementation("org.jadira.usertype:usertype.core:6.0.1.GA")
+    implementation("com.fasterxml.jackson.module:jackson-module-parameter-names:$jacksonVersion")
+    implementation("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:$jacksonVersion")
+    implementation("org.zalando.stups:stups-spring-oauth2-server:1.0.18")
+    implementation("org.zalando:problem-spring-web:0.19.0")
+    implementation("org.zalando:twintip-spring-web:1.1.0")
```
cberg-zalando, line 67: Twintip was already obsoleted since at least 2019.

server/zally-server/build.gradle.kts
```diff
@@ -24,33 +24,33 @@ dependencies {
...
+    implementation("org.springframework.boot:spring-boot-starter-actuator:$springBootVersion")
+    implementation("org.springframework.boot:spring-boot-starter-data-jpa:$springBootVersion") {
         exclude("org.hibernate", "hibernate-entitymanager")
     }
-    compile("org.flywaydb:flyway-core:5.1.4")
-    compile("org.hsqldb:hsqldb:2.4.1")
-    compile("org.postgresql:postgresql:42.2.4")
-    compile("org.hibernate:hibernate-core:5.3.5.Final")
-    compile("org.jadira.usertype:usertype.core:7.0.0.CR1") {
+    implementation("org.flywaydb:flyway-core:5.1.4")
```
cberg-zalando, line 36: The same that I wrote above about overwriting Spring Boot managed dependencies applies here as well, even though the Spring Boot version is newer.

github-integration/build.gradle.kts
tkrop: Yes, supporting this is not necessary, but this is a separate issue.

github-integration/build.gradle.kts
vadeg: @cberg-zalando thank you for the update. We would definitely switch, but not on this PR. There are a lot of changes already and I don't want to add another "problem" in the scope.

cberg-zalando: Sure, I can understand this. Just thought, I give the PR's a view as I was somehow involved in getting the scope problem solved. So just thought to `give something back` as I am the `dependency bump person` in my team. :laughing:

vadeg:
`NullPointerExceptionTest` fails. I am not sure of the need of this test. It simply set `null` or removes values of properties from the schema and checks if rules run without NPE. In fact those removals break Swagger parser itself, because some of the properties can not be removed, e.g. you can not define a [Server Variable](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#serverVariableObject) without a reference or value:
```yaml
...
servers:
  - url: '{scheme}://example.com/'
    description: Example Server
    variables:
      scheme:
...
```
This causes the error in `swagger-parser`:
```kotlin
class com.fasterxml.jackson.databind.node.NullNode cannot be cast to class com.fasterxml.jackson.databind.node.ObjectNode (com.fasterxml.jackson.databind.node.NullNode and com.fasterxml.jackson.databind.node.ObjectNode are in unnamed module of loader 'app')
java.lang.ClassCastException: class com.fasterxml.jackson.databind.node.NullNode cannot be cast to class com.fasterxml.jackson.databind.node.ObjectNode (com.fasterxml.jackson.databind.node.NullNode and com.fasterxml.jackson.databind.node.ObjectNode are in unnamed module of loader 'app')
	at io.swagger.v3.parser.util.OpenAPIDeserializer.getServerVariables(OpenAPIDeserializer.java:532)
	at io.swagger.v3.parser.util.OpenAPIDeserializer.getServer(OpenAPIDeserializer.java:465)
	at io.swagger.v3.parser.util.OpenAPIDeserializer.getServersList(OpenAPIDeserializer.java:435)
	at io.swagger.v3.parser.util.OpenAPIDeserializer.parseRoot(OpenAPIDeserializer.java:192)
	at io.swagger.v3.parser.util.OpenAPIDeserializer.deserialize(OpenAPIDeserializer.java:149)
...
```
In summary, the test breaks what should not be broken by specification definition. I am thinking of removing it. @roxspring it's been long time ago when you have added this test. Do you remember what was the goal?

roxspring:
> In summary, the test breaks what should not be broken by specification definition. I am thinking of removing it. @roxspring it's been long time ago when you have added this test. Do you remember what was the goal?

The requirement came from somebody at Zalando, I don't remember who, I just implemented it because I was working in that area and was intrigued by the challenge. IIRC it was around the time that we added support for OpenAPI 3. The test is designed to test the rules with a variety of broken, or just partially empty, specifications and confirm that the rules are able to handle them properly. The intention was to force us to build the rules more robustly, such that people submitting a slightly broken specification got a useful response rather than some `500 Server Error`. Prior to this test being in place there were _a lot_ of rules which would fail hard, and it picked up on many null-dereferences in my own rules since.

Personally I'd recommend finding a way to keep the test in place - even though the spec you presented is invalid, a user submitting it deserves better feedback as to what was wrong and where.

Hope that helps - unfortunately I don't have capacity for much more than answering the odd question these days - my current role doesn't even involve web APIs so I can't really justify time spent on Zally 😞

server/build.gradle.kts
```diff
@@ -2,7 +2,8 @@ import org.jetbrains.dokka.gradle.DokkaTask
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
-    val kotlinVersion = "1.3.21"
+    val kotlinVersion = "1.4.32"
```
zeitlinger, line 5: https://github.com/JetBrains/kotlin/releases/tag/v1.5.20 is already out

vadeg:
👍

cberg-zalando:
:+1:

vadeg:
@tkrop do you want to have a look?

server/build.gradle.kts
vadeg: I know. I don't want to upgrade from 1.3. directly to 1.5. I was fighting with many libs and plugins to make them work. Don't want to spend one more month on doing this. This could be a separate ticket.

server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/CommonFieldTypesRule.kt
```diff
@@ -19,15 +22,22 @@ class CommonFieldTypesRule(rulesConfig: Config) {
...
                 if (violationDesc != null) {
-                    context.violations(violationDesc, schema)
+                    if (schema.isObjectSchema()) {
+                        context.violations(violationDesc, schema)
+                    } else {
+                        context.violations(
+                            violationDesc,
+                            context.getJsonPointer(parentObjectSchema) + "/properties/$name".toJsonPointer()
+                        )
+                    }
```
tkrop, line 40: Not sure, whether this is a good idea. Isn't the effect, that a violation would be notified twice - once in the defining section, once in the referencing section?

I would suggest, to try to limit this on a single violation on the defining location. However, I must admit, I see the problem of references to external locations that should also create violations. So if this covers the last situation - even if the test is not addressing this issue, I would be okay with this change. Else, may be we should separate this logic change.

server/zally-core/src/main/kotlin/org/zalando/zally/core/ast/ReverseAstBuilder.kt
```diff
@@ -16,9 +17,10 @@ import kotlin.collections.component2
 import kotlin.collections.set
 
 class ReverseAstBuilder<T : Any> internal constructor(root: T) {
+    private val log = LoggerFactory.getLogger(ReverseAstBuilder::class.java)
     private val extensionMethodNames = HashSet<String>()
 
-    private val nodes = ArrayDeque<Node>(listOf(Node(root, EMPTY_JSON_POINTER, null)))
+    private val nodes = ArrayDeque(listOf(Node(root, EMPTY_JSON_POINTER, null)))
```
tkrop, line 23: Why is removing the type here necessary? Can't we keep it?

tkrop (commented):
Looks good. Only two questions left.

server/zally-core/src/main/kotlin/org/zalando/zally/core/ast/ReverseAstBuilder.kt
vadeg: There is no need to keep the generic type here. The generic type is derived from the constructor argument (`listOf(...)`).

tkrop: I see. Nice trade of Kotlin, but some times I miss the explicit types for readability - especially when it starts looking inconsistent as here. Anyhow, I'm okay with dropping it.

server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/CommonFieldTypesRule.kt
vadeg: There will be only one violation in this case because of how the Swagger parser resolves references. Consider the example
```yaml
openapi: 3.0.1
  paths:
    /pets:
      get:
        responses:
          200:
            content:
              application/json:
                schema:
                  properties:
                    id:
                      $ref: "#/components/schemas/CustomId"
  components:
    schemas:
      CustomId:
        type: integer
        format: int64
```

After Swagger parser will parse this schema, it will replace all references with real values. In our case it will be
```yaml
openapi: 3.0.1
  paths:
    /pets:
      get:
        responses:
          200:
            content:
              application/json:
                schema:
                  properties:
                    id:
                      # Here the $ref is replaced with an actual value
                      type: integer 
                      format: int64
  components:
    schemas:
      CustomId:
        type: integer
        format: int64
```

Without the condition I added the violation path will be `/components/schemas/CustomId`. This definition itself doesn't violate [rule 174](https://opensource.zalando.com/restful-api-guidelines/#174). Only a combination of the property name `id` and the property type `integer` violates the rule. 
This root cause of the problem is how Zally stores and lookups JSON pointers. I have described this problem in #1255 in details. Until it will be fixed this is the only mitigation strategy I have found to keep JSON pointers point to exact property (`/paths/~1pets/get/responses/200/content/application~1json/schema/properties/id`) rather than to the property (`/components/schemas/CustomId`) definition leaving to the client guessing what is wrong with the custom type.

tkrop: The begin of your explanation sound good, but the last sentence confuses me. I would expect the visible location to be `/components/schemas/CustomId` since this is where the problem can be found, and not `/paths/~1pets/get/responses/200/content/application~1json/schema/properties/id`. Can you check your last sentence?

tkrop: Thanks for the discussion. Now I understand the problem here.

tkrop:
:+1:

PR #1269, opened 2021-06-24
Fix `content` field validation

### LINKED ISSUE #1268: Incorrect and documented Zally rule introduced failures
We have recently started getting violations reported from Zally such as:

```
MUST Path parameters validation
	Parameter blocklist_id: "content" property should have exactly one entry
	https://github.com/zalando/zally/blob/master/server/rules.md#z001-path-parameters-validation
		components > parameters > BlocklistId (lines 31-40)
```

The corresponding path parameter has a `schema` property. The OpenAPI spec defines `schema` and `content` to be mutually exclusive. So this is simply wrong.

The link in the error message is also wrong and doesn't provide an explanation on this anywhere on the linked page.

vadeg:
Fixes #1268


tkrop:
:+1:

tkrop (approved):
Looks good.

vadeg:
👍

PR #1270, opened 2021-06-28
Add an optional flag to skip SSL request verification

### LINKED ISSUE #1185: Cannot use CLI against server with self signed SSL certificate
We are running a Zally server instance proxied by an Nginx VHost that also does the SSL offloading.
From our CI we try to run a Zally validation against this server using the CLI:

```bash
zally -l https://zally.theserver.org lint ./build/openapi.json
```
Causing this error:
```bash
Post "https://zally.theserver.org/api-violations": x509: certificate signed by unknown authority
```

As of now, it is not possible to skip the SSL validation.
Please add an argument for disabling the validation to the CLI (e.g. `--skip-ssl-validation`).

vadeg:
Fixes #1185

### DIFF
--- a/cli/zally/app.go
+++ b/cli/zally/app.go
@@ -34,6 +34,10 @@ func CreateApp() *cli.App {
 			Usage: "Output format `[pretty|markdown|text]`",
 			Value: "pretty",
 		},
+		cli.BoolFlag{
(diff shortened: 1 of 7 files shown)


tkrop:
:+1:

vadeg:
👍

PR #1277, opened 2021-07-06
fix: supported references for problem+json (#1274)

### LINKED ISSUE #891: Incorrect Severity of Rule [151] (MUST Specify Success and Error Responses)
Currently MAY instead of MUST.

### LINKED ISSUE #1274: Zally validation for rule 151 does not accept reference code from rule 151
https://opensource.zalando.com/restful-api-guidelines/#151 provides this example for the default:

```yaml
responses:
  ...
  default:
    description: error occurred - see status code and problem object for more information.
    content:
      "application/problem+json":
        schema:
          $ref: 'https://opensource.zalando.com/restful-api-guidelines/models/problem-1.0.1.yaml#/Problem'
```

This leads to:

```
MAY – Specify Success and Error Responses

problem json has to be used as default response (https://zalando.github.io/problem/schema.yaml#/Problem)

Rule: https://zalando.github.io/restful-api-guidelines/#151

Location: paths > /pdfs > post > responses > default > content > application/problem+json
```

Substituting the ref resolves the problem:

```yaml
responses:
  ...
        default:
          description: error occurred - see status code and problem object for more information.
          content:
            application/problem+json:
              schema:
```

Either the reference from the guidelines should be accepted or the guideline should be adapted to show the "acceptable" URL.

tkrop:
fixes #1274 
fixes #891

Tricky! As it actually turned out, not the missing reference broke the check, but the change in the problem specification the check was not adapted to. However, the reference check does play a role as soon as the reference cannot be resolved for any reason. It is just not tested while validating the check.

@vadeg if you know how to switch of reference resolution for one test, I would like to add this as an additional test case.

Anyhow, I added a separate check method as well as full test coverage for checking and notifying the media-type violations.


server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/JsonProblemAsDefaultResponseRule.kt
```diff
@@ -17,23 +17,35 @@ import io.swagger.v3.oas.models.responses.ApiResponse
...
+        .map { context.violation("media-type application/problem+json should be used as default response", it.value) }
+
+    @Check(severity = Severity.SHOULD)
+    fun checkDefaultResponseIsProblemJsonSchema(context: Context): List<Violation> = responsesPerOperation(context)
         .filter { "default" in it.second.keys }
         .flatMap { it.second.getValue("default").content.orEmpty().entries }
         .filter { (contentType, _) -> contentType in validContentTypes }
-        .filterNot { it.value?.schema?.`$ref` in validRefs || isProblemJsonSchema(it.value?.schema) }
-        .map { context.violation("problem json has to be used as default response (${validRefs[0]})", it.value) }
+        .filterNot { it.value?.schema?.`$ref` in validRefs }
```
tkrop, line 46: This filter is actually only active, when the references are not resolved, which we currently have no happy path test case for.

vadeg: Swagger parser in Zally always fully resolves references (applies both [resolve](https://github.com/swagger-api/swagger-parser#1-resolve) and [resolveFully](https://github.com/swagger-api/swagger-parser#2-resolvefully) options). I think we can remove completely the ref check from this rule and check only that the default response is a Problem.json object regardless of the reference, because refs will be removed from the schema.

tkrop: I think, keeping it is more robust. If we drop it, the server will fail, when the `problem+json` resource is unreachable - not sure how often this may happen. I would more like to add a test that enforces non-loading ...

tkrop:
:+1:

server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/JsonProblemAsDefaultResponseRule.kt
vadeg: Non-loading would never happen in reality unless we allow to do so :) If the resource is unreachable, then the response will be resolved to nullable response without properties in the schema. This is what the rule can check.

vadeg:
👍

PR #1282, opened 2021-07-06
Drop `json_payload` column.

### LINKED ISSUE #890: api_review.json_payload table is being filled with non-json
Just took a look at the `api_review` table's `json_payload` column in the database and discovered it's currently being populated with `ApiDefinitionRequest.toString()` - which is definitely not JSON formatted. 

We do have a separate `api_review.api_definition` column with the typically YAML formatted body which makes me wonder whether the correct fix is to just drop the `json_payload` column entirely (or at least use of it).

vadeg:
Fixes #890

### DIFF
--- a/server/zally-server/src/main/resources/db/migration/V006__drop_json_payload_column.sql
+++ b/server/zally-server/src/main/resources/db/migration/V006__drop_json_payload_column.sql
@@ -1,0 +1,1 @@
+ALTER TABLE api_review DROP COLUMN json_payload;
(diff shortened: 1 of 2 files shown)


vadeg:
👍

server/zally-server/src/main/resources/db/migration/V006__drop_json_payload_column.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE api_review DROP COLUMN json_payload;
```
tkrop, line 1: Before we drop this, we should consequently look into our database, whether `api_definition` is really filled as expected for the entire review history and merge the content if necessary.

tkrop: @vadeg I had a first look at the data. It looked actually, as if storing this data has been a dumb idea in the first place. I will approve as soon as the merge conflict is resolved.

vadeg:
@tkrop the conflict has been resolved.

vadeg:
👍

vadeg:
@tkrop I have resolved conflicts. Please review again.

tkrop:
:+1:

PR #1284, opened 2021-07-07
Fix project directories

vadeg:
Move all Kotlin sources to `kotlin` directories instead of `java` in `server` application.

Related to #560


vadeg:
👍

tkrop:
:+1:

Issue #1291, opened 2021-07-10
not able to run the rules validator in non spring envrionment


thiyagu06:
I'm trying to write a gradle plugin to be able to lint the spec during the build time itself for faster feedback to the developers. 

When I run the ruleValidator as non spring managed beans, I'm getting the below error when creating the instance of UseOpenApiRule. find the project here to reproduce the error. https://github.com/thiyagu06/zally-gradle-plugin.

When I change the class loader on this line, https://github.com/zalando/zally/blob/master/server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/UseOpenApiRule.kt#L32 as below, it seems to be working. Any inputs on any other ways to solve this.

`javaClass.classLoader.getResource("schemas/${name.toLowerCase()}-schema.json")`.

```
Unable to load schema: jar:file:/Users/thiyagu/.gradle/caches/jars-8/bdc69e2ad1135374ffd290de15007275/zally-core-2.0.0.jar!/schemas/swagger-schema.json
java.lang.NullPointerException: Cannot invoke "String.endsWith(String)" because the return value of "java.net.URI.getPath()" is null
        at com.github.fge.jsonschema.core.util.URIUtils$6.check(URIUtils.java:201)
        at com.github.fge.jsonschema.core.util.URIUtils$6.check(URIUtils.java:192)
        at com.github.fge.jsonschema.core.util.Registry.put(Registry.java:99)
        at com.github.fge.jsonschema.core.load.uri.URITranslatorConfigurationBuilder.addSchemaRedirect(URITranslatorConfigurationBuilder.java:152)
        at com.github.fge.jsonschema.core.load.uri.URITranslatorConfigurationBuilder.addSchemaRedirect(URITranslatorConfigurationBuilder.java:174)
        at org.zalando.zally.core.JsonSchemaValidator.createLoadingConfiguration(JsonSchemaValidator.kt:97)
        at org.zalando.zally.core.JsonSchemaValidator.createValidatorFactory(JsonSchemaValidator.kt:87)
        at org.zalando.zally.core.JsonSchemaValidator.<init>(JsonSchemaValidator.kt:29)
        at org.zalando.zally.ruleset.zalando.UseOpenApiRule.getSchemaValidators(UseOpenApiRule.kt:95)
        at org.zalando.zally.ruleset.zalando.UseOpenApiRule.<init>(UseOpenApiRule.kt:38)
[...]

PR #1293, opened 2021-07-12
Replace Guava classloader with a default Java classloader

vadeg:
Triggered by #1291, but not a fix to this problem. Simply clean up not needed dependency from `guava` in the code.


vadeg:
👍

tkrop:
:+1:

PR #1331, opened 2021-10-21
Update web UI to Node 14

### LINKED ISSUE #1328: zally giving errors due to which.js
I am trying to set up the zally with the new release and it is giving the error on the which.js file which is present inside the web-ui node modules folder. Attached are logs for reference.

[logs.txt](https://github.com/zalando/zally/files/7323703/logs.txt)

image

_Using java8 and python3 already installed in the system._

vadeg:
* Update Node version for `web-ui`
* Update documentation

Fixes #1328


tkrop (commented):
Looks good.

PR #1367, opened 2022-03-02
feat: change severity of proprietary header rule (#1367)

tkrop:
It seems we have overlooked a two years old guideline change on the severity of proprietary header usage that was changed from MUST to SHOULD (see https://opensource.zalando.com/restful-api-guidelines/#183).


tkrop:
:+1:

hjacobs:
:+1:

PR #1368, opened 2022-03-09
fix: missing annotation changes (#1367)

tkrop:
fixes #1367.

Changing the severity in the `@Rule` is not enough. Need to change severity in `@Check` as real source of truth too.


zeitlinger:
:+1:

tkrop:
:+1:

PR #1343, opened 2022-01-03
feat: update dependencies (#1318)

tkrop:
fixes #1318.
substitutes #1327.

Issue #101, opened 2017-01-11
Upgrade to swagger-parser 1.0.25 breaks rule


fmueller:
There is an issue with newest version of swagger-parser library (1.0.25). One test breaks. Needs to be investigated and fixed.

PR #1382, opened 2022-04-07
feat: upgrade swagger dependencies (#1382)

### LINKED ISSUE #1230: Zally not validating APIs for Open API 3.1.0.
It seems as if there is a switch preventing the rules to get into action on Open API version 3.1.0, that has been released on Feb 15th, 2021. Checking APIs with `openapi: 3.0.3` that show violations show now violations after updating to `openapi: 3.1.0` without other changes.

tkrop:
fixes #1343.
contributes #1230.

Upgrade swagger dependencies to enable support for Open API 3.1. As it turned out while testing #1343 while migrating to JUnit 5 was actually disabling all tests through the missing dependency to the runtime. So we had to fix this first.

I also added a small fix of the Zally API actually defining Bearer Authentication.

Unfortunately, updating to the latest library is not sufficient for OpenAPI 3.1. Adding smoke test for Open API 3.1 support showed, that support for OpenAPI 3.1 in the latest swagger-parser version is still missing. Only swagger-core is ready for OpenAPI 3.1.

I finally added an improved failure handling on unsupported OpenAPI versions and unparsable files: They consistently now violate rule [#101](https://opensource.zalando.com/restful-api-guidelines/#101) of the guideline.

### DIFF
--- a/server/zally-core/src/main/kotlin/org/zalando/zally/core/ContextRulesValidator.kt
+++ b/server/zally-core/src/main/kotlin/org/zalando/zally/core/ContextRulesValidator.kt
@@ -2,6 +2,7 @@ package org.zalando.zally.core
 
 import com.fasterxml.jackson.core.JsonPointer
 import org.zalando.zally.rule.api.Context
+import org.zalando.zally.rule.api.Violation
 
 /**
  * This validator validates a given OpenAPI definition based
@@ -14,13 +15,16 @@ class ContextRulesValidator(
 
     override fun parse(content: String, authorization: String?): ContentParseResult<Context> {
         // first try to parse an OpenAPI (version 3+)
-        return when (val parsedAsOpenApi = defaultContextFactory.parseOpenApiContext(content, authorization)) {
-            is ContentParseResult.NotApplicable ->
-                // if content was no OpenAPI, try to parse a Swagger (version 2)
-                defaultContextFactory.parseSwaggerContext(content)
-            else ->
-                parsedAsOpenApi
+        val parsedAsOpenApi = defaultContextFactory.parseOpenApiContext(content, authorization)
+        if (parsedAsOpenApi !is ContentParseResult.NotApplicable) {
+            return parsedAsOpenApi
         }
+        // if content was no OpenAPI, try to parse a Swagger (version 2)
+        val parsedAsSwagger = defaultContextFactory.parseSwaggerContext(content)
+        if (parsedAsSwagger !is ContentParseResult.NotApplicable) {
+            return parsedAsSwagger
+        }
+        return ContentParseResult.ParsedWithErrors(listOf(Violation("No valid Open API specification", EMPTY_JSON_POINTER)))
     }
 
     override fun ignore(root: Context, pointer: JsonPointer, ruleId: String) = root.isIgnored(pointer, ruleId)
--- a/server/zally-core/src/main/kotlin/org/zalando/zally/core/ast/ReverseAstBuilder.kt
+++ b/server/zally-core/src/main/kotlin/org/zalando/zally/core/ast/ReverseAstBuilder.kt
@@ -106,7 +106,7 @@ class ReverseAstBuilder<T : Any> internal constructor(root: T) {
                         val nextPath = m.name
                             .takeIf { it !in this.extensionMethodNames }
                             ?.removePrefix("get")
-                            ?.decapitalize()
+                            ?.replaceFirstChar({ it.lowercase() })
                             ?: ""
 
                         nodes.push(Node(value, pointer + nextPath.toEscapedJsonPointer(), marker))
--- a/server/zally-core/src/test/kotlin/org/zalando/zally/core/DefaultContextFactoryTest.kt
+++ b/server/zally-core/src/test/kotlin/org/zalando/zally/core/DefaultContextFactoryTest.kt
@@ -4,6 +4,7 @@ import org.zalando.zally.core.ContentParseResultAssert.Companion.assertThat
 import org.assertj.core.api.Assertions.assertThat
 import org.intellij.lang.annotations.Language
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.Disabled
 
 class DefaultContextFactoryTest {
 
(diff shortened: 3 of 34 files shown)


server/zally-core/src/test/kotlin/org/zalando/zally/core/DefaultContextFactoryTest.kt
```diff
@@ -26,7 +27,59 @@ class DefaultContextFactoryTest {
     }
 
     @Test
-    fun `OPEN API -- openapi specification without info and paths succeeds with messages`() {
+    @Disabled("discovery of OpenAPI 3.1. not working as expected")
+    fun `OPEN API -- OpenAPI 31 without info and paths succeeds with messages`() {
```
tkrop, line 31: Start of disabled test section.

server/zally-ruleset-zally/src/test/kotlin/org/zalando/zally/ruleset/zally/PathParameterRuleTest.kt
```diff
@@ -207,6 +208,7 @@ class PathParameterRuleTest {
     }
 
     @Test
+    @Disabled("Test is failing to produce a violation, however I feel that the rule does not make sense")
     fun `return violation if 'content' field contains more than one key`() {
         @Language("YAML")
         val context = DefaultContextFactory().getOpenApiContext(
```
tkrop, line 214: @ePaul @vadeg @tfrauenstein This is actually unclear why it is failing to show a violation with the new library, since the rule is supposed to create a violation. However, the rule behind semantically seems not to make sense. Why should we create a violation on multi-part response definitions?

vadeg: @ePaul ZXX rules mostly about checking the OpenAPI schema rules rather then API guidelines. I doubt that we want to move all API validation checks to the API.

server/zally-core/src/test/kotlin/org/zalando/zally/core/DefaultContextFactoryTest.kt
```diff
@@ -26,7 +27,61 @@ class DefaultContextFactoryTest {
     }
 
     @Test
-    fun `OPEN API -- openapi specification without info and paths succeeds with messages`() {
+    @Disabled("OpenAPI 3.1 is not supported by latest swagger parser yet")
+    fun `OPEN API -- OpenAPI 31 without info and paths succeeds with messages`() {
```
ePaul, line 31:
```suggestion
    fun `OPEN API -- OpenAPI 3.1 without info and paths succeeds with messages`() {
```

ePaul: Ah, I guess dots are not allowed here, as this is some kind of method name? Can we find some other replacement?

tkrop: Yes, exactly. `.` is reserved character.

server/zally-ruleset-zally/src/test/kotlin/org/zalando/zally/ruleset/zally/PathParameterRuleTest.kt
tkrop: Oh, true. Seems as if I have not given enough attention to the context.

ePaul: @vadeg 
1. If there is a violation, its link should point to somewhere where it's documented. Doesn't have to be the API guidelines, the Zally documentation is also fine for me. (It's pointing to https://github.com/zalando/zally/blob/main/server/rules.md#z001-path-parameters-validation – the page exists, but no section with this ID.)
2. A rule about request body certainly doesn't belong to the "path parameter rule"?
3. Where in the OpenAPI specification does it say that there can only be a single content type in the `content` map? (I.e. where does this specific violation come from)?

server/zally-core/src/main/kotlin/org/zalando/zally/core/ast/ReverseAstBuilder.kt
```diff
@@ -106,7 +106,7 @@ class ReverseAstBuilder<T : Any> internal constructor(root: T) {
                         val nextPath = m.name
                             .takeIf { it !in this.extensionMethodNames }
                             ?.removePrefix("get")
-                            ?.decapitalize()
+                            ?.replaceFirstChar({ it.lowercase() })
```
ePaul, line 109: What happened to the `decapitalize` and `capitalize` methods?

tkrop: was deprecated.

server/zally-core/src/test/kotlin/org/zalando/zally/core/DefaultContextFactoryTest.kt
tkrop: Okay, will improve.

server/zally-ruleset-zally/src/test/kotlin/org/zalando/zally/ruleset/zally/PathParameterRuleTest.kt
tkrop: The problem here is that the latest swagger-parser maps the incorrect `content` to `null` which does not allow us to check for any violation. I suggest to `@Disable` the test for now and have a look on the next upgrade how they proceed on this.

tkrop:
:+1:

server/zally-core/src/main/kotlin/org/zalando/zally/core/ContextRulesValidator.kt
```diff
@@ -14,13 +15,16 @@ class ContextRulesValidator(
...
+        val parsedAsOpenApi = defaultContextFactory.parseOpenApiContext(content, authorization)
+        if (parsedAsOpenApi !is ContentParseResult.NotApplicable) {
+            return parsedAsOpenApi
         }
+        // if content was no OpenAPI, try to parse a Swagger (version 2)
+        val parsedAsSwagger = defaultContextFactory.parseSwaggerContext(content)
+        if (parsedAsSwagger !is ContentParseResult.NotApplicable) {
+            return parsedAsSwagger
+        }
+        return ContentParseResult.ParsedWithErrors(listOf(Violation("No valid Open API specification", EMPTY_JSON_POINTER)))
```
tkrop, line 27: This is how the violation is triggered on the failed parsing.

zeitlinger:
:+1:

PR #1386, opened 2022-04-27
fix: remove duplicate hostname violations on invalid audience (#1386)

tkrop:
Fixing the duplicate functional hostname violations that appear in case of missing, `null`, or invalid `x-audience` that are preconditions for this check. The idea is to silently ignore the functional hostname check if no proper audience is provided.

Example output copied from new Backstage API Linter:
image
image
image

### DIFF
--- a/server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/FunctionalNamingForHostnamesRule.kt
+++ b/server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/FunctionalNamingForHostnamesRule.kt
@@ -62,11 +62,12 @@ class FunctionalNamingForHostnamesRule(rulesConfig: Config) {
         val apiAudience = try {
             context.api.info?.apiAudience()
         } catch (e: UnsupportedAudienceException) {
-            return context.violations(e.message!!, context.api.info)
+            // case is covered by "APIAudienceRule"
+            return emptyList()
         }
         return when {
-            apiAudience !in audiencesToCheck -> emptyList()
-            apiAudience == null -> context.violations("API info section is not defined", context.api)
+            // cases are covered by "APIAudienceRule"
+            apiAudience == null || apiAudience !in audiencesToCheck -> emptyList()
             context.swagger != null -> checkHostnamesInSwaggerHost(context, apiAudience)
             else -> checkHostnamesInOpenAPIServers(context, apiAudience)
         }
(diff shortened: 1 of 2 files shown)


tfrauenstein:
@tkrop Perhaps I misunderstood the problem situation, and I cannot correlate the screenshots with the error description. The screen shot shows multiple problems: a) rule 224 violation duplicates (for MUST, SHOULD, MAY variants), b) the rule 224 is just a MUST rule and c) why is API audience provided as hint? -- functionally and syntactically Hostname Rule 224 and Audience Rule 219 are not connected. So ad-hoc I do not understand why invalid audience definition leads to duplication of hostname rule violation. Is the hostname correct? Is there no easy way to avoid duplicate hostname rule check in case of audience check violation? I would prefer it, instead of not executing the check at all, if some other check fails.

ePaul:
@tfrauenstein  As I understand the problem (without looking at the code):
* To decide which of the three versions of rule 224 to use, Zally needs to check the audience field (because this determines whether it's a MUST/SHOULD/MAY rule).
* If that field is null, in the current implementation, all three violations appear (with a message of "audience null is not supported"), without even looking at the hostname.

@tkrop's solution is "make this violation silent when there is no audience" (there is another violation for missing audience anyways, and when that one is fixed, one of the three versions of the rule will be used).

server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/FunctionalNamingForHostnamesRule.kt
```diff
@@ -62,11 +62,12 @@ class FunctionalNamingForHostnamesRule(rulesConfig: Config) {
...
             context.api.info?.apiAudience()
         } catch (e: UnsupportedAudienceException) {
-            return context.violations(e.message!!, context.api.info)
+            // cases covered by API Meta Information rule
+            return emptyList()
         }
         return when {
-            apiAudience !in audiencesToCheck -> emptyList()
-            apiAudience == null -> context.violations("API info section is not defined", context.api)
+            // cases covered by API Meta Information rule
```
vadeg, line 69:
```suggestion
            // case is covered by the "ApiAudienceRule"
```

vadeg:
👍

server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/FunctionalNamingForHostnamesRule.kt
tkrop: changed.

tkrop:
:+1:

PR #1414, opened 2022-10-16
Improve violation message and code readability

### LINKED ISSUE #1399: Better violation message for "wrong status code for this HTTP method"
**Current behavior**

When I have a GET operation with a 202 response, Zally gives me a SHOULD violation with message "Operations should use well-understood HTTP status codes", pointing to [rule 150](https://opensource.zalando.com/restful-api-guidelines/index.html#150).

The list in the rule contains 202.

The issue is that 202 does not apply to GET, but the error message doesn't tell this.

If I have a status code which is completely missing in the list, like 302, I instead get "302 is not a well-understood HTTP status code", which is clearer.

**Expected behavior**

The violation message says something like "HTTP status code 202 should not be used for GET operations".

**How to reproduce**

Here is a minimal example (when I remove more fields, there are more violations):

```yaml
openapi: 3.0.0
info:
  x-api-id: 351db032-f0d2-11ec-8ea0-0242ac120002
  version: 1.0.0
  title: Example API
  description: "Showcase for Zally issue #1399"
  contact:
    name: Paul Ebermann
    url: about:blank
    email: nobody@example.com
  x-audience: component-internal
paths:
  /examples/{example_id}:
    get:
      summary: Sales module and sales campaign KPIs as a csv file.
      security:
        - oauth2:
            - example.read
      parameters:
        - name: example_id
          in: path
          required: true
          schema:
            type: string
      responses:
        "202":
          description: Csv is generating but not ready for download.
          headers:
            Retry-After:
              description: "Retry contacting the endpoint after *at least* x seconds."
              schema:
                format: "int32"
                type: "integer"

components:
  securitySchemes:
    oauth2:
      type: oauth2
      flows:
        clientCredentials:
          tokenUrl: https://identity.zalando.com/oauth2/token
          scopes:
            example.read: Access right needed to read from the API
```

vadeg:
* Update violation message for incorrect usage of well understood status codes
* Rename functions in tests to have clear meaning
* Reformat map filters in several rule checks

Fixes #1399

### DIFF
--- a/server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/UseStandardHttpStatusCodesRule.kt
+++ b/server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/UseStandardHttpStatusCodesRule.kt
@@ -1,12 +1,12 @@
 package org.zalando.zally.ruleset.zalando
 
 import com.typesafe.config.Config
+import io.swagger.v3.oas.models.PathItem
 import org.zalando.zally.rule.api.Check
 import org.zalando.zally.rule.api.Context
 import org.zalando.zally.rule.api.Rule
 import org.zalando.zally.rule.api.Severity
 import org.zalando.zally.rule.api.Violation
-import io.swagger.v3.oas.models.PathItem
 
 /**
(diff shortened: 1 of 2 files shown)


vadeg:
👍

vadeg:
@tkrop please review

PR #1415, opened 2022-10-16
Add rule to check NoContent response

### LINKED ISSUE #1403: feat: new rule: Check 204 for successful requests w/o response payload.
HTTP standards requires delivering of 204 (ok, no content), instead of 200 (ok) in case of successful PUT/POST/PATCH operations without response payload. It is defined in [SHOULD only use most common HTTP status codes [150]](https://opensource.zalando.com/restful-api-guidelines/#150).
Let's extend Zally -- it looks like an easy to check rule.

vadeg:
Add a rule which checks that `204` response has no content defined.

Fixes #1403

### DIFF
--- a/cli/zally/integration_test.go
+++ b/cli/zally/integration_test.go
@@ -1,3 +1,4 @@
+//go:build integration
 // +build integration
 
 package main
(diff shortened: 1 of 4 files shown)


vadeg:
👍

tkrop:
:+1:

PR #1435, opened 2023-01-05
feat: upgrade to kotlin 1.8 (#1435)

tkrop:
Updates Kotlin version and prepares the switching of the linter plugin to apply the newest ktlint standard. Also fixes the broken build that needs a bigger host for building the docker container.


natalyapeixoto:
👍

PR #1433, opened 2023-01-04
feat: update to java 17 to prepare spring 3 support (#1433)

tkrop:
Switch to Java 17 to continue on upgrading Spring and Hibernate dependencies as started in #1432.

PR #1437, opened 2023-01-05
feat: activate and apply new linter (#1437)

tkrop:
This pull request activates and applies the new linter rules using the automated code formatting it provides. In addition, it fixes the docker image build that was broken in #1433 when migrating to Java 17.

PR #1438, opened 2023-01-05
feat: fix container image build (#1437)

tkrop:
Fixes the container image build change from #1437 which was supposed to fix #1433.


tkrop:
👍

natalyapeixoto:
👍

PR #1439, opened 2023-01-06
fix: docker image build (#1437)

tkrop:
Finally fixes the docker build problem.

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -6,17 +6,18 @@ pipeline:
     vm_config:
       type: linux
       image: cdp-runtime/jdk17
+      size: large
     commands:
       - desc: Build server
         cmd: |
-          cd $PWD/server
-          ./gradlew --console=plain build
+          cd ${PWD}/server;
+          ./gradlew --console=plain build;
           if [ -n "${CDP_PULL_REQUEST_NUMBER}" ]; then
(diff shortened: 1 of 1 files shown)


tkrop:
👍

delivery.yaml
```diff
@@ -28,13 +29,7 @@ pipeline:
     commands:
       - desc: Build Web UI
         cmd: |
-          cd $PWD/web-ui
-          yarn install -q ; yarn test ; yarn build
-          if [ -n "${CDP_PULL_REQUEST_NUMBER}" ]; then
-            echo "We're in a pull request, aborting."; exit 0;
-          fi;
-          PREFIX="pierone.stups.zalan.do/builder-knowledge"
-          IMAGE="${PREFIX}/zally-web-ui-dummy:${CDP_BUILD_VERSION}"
```
kgalli, line 37: We do not have to build the docker image anymore?

tkrop: No we do not need this any more. Zally-Web-UI is retired. This image is not accessible for the public, so we can stop deploying it.

natalyapeixoto:
👍

delivery.yaml
kgalli: Why do we even build this step then? Is it to ensure it can be build?

Issue #1445, opened 2023-02-21
Update Swagger-Parser Dependency to 2.1.12


jeffrey-ng:
Zally's current version Swagger Parser (2.1.9) incorrectly throws exceptions with relative `$ref` within a spec.
Updating to 2.1.12 should fix the issue.

https://github.com/swagger-api/swagger-parser/issues/1865
https://github.com/swagger-api/swagger-parser/commit/138687c5385556a648ed2a14528f522670242f99

tkrop:
Thank you very much for the information. Will update version as soon as possible.

ktsypkina:
the Swagger Parser dependency was upgraded to 2.1.12 @jeffrey-ng

PR #1480, opened 2024-02-19
Address issue #1445 and fix failed test

ktsypkina:
Addresses https://github.com/zalando/zally/issues/1445
As Zally does not support validation for relative refs, no new test cases to cover this scenario

After upgrading to io.swagger.parser.v3:swagger-parser:2.1.9 (from 2.0.32) the test case was failing due to different error message.

This error occured because of changes in OpenAPIDeserializer:
`rootMap = new ObjectMapper().convertValue(rootNode, Map.class);`
the convert to rootMap was introduced before trying to parse the rootNode. That's why it is not setting the correct message in the result. 

Before the upgrade the message was
`attribute openApi is not of type "object"`

After upgrade the `IllegalArgumentException` is handled by the library itself, setting the message to 
```
Cannot construct instance of `java.util.LinkedHashMap` (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value ('no swagger definition') at [Source: UNKNOWN; byte offset: #UNKNOWN]
```

Aligned with @tkrop on the fix:
handle this exception as violation of [rule #101](https://opensource.zalando.com/restful-api-guidelines/#101)
Since there is 0 attributes in test example `"no swagger definition"` it is handled as ParseResult.NotApplicable with following details:
```
{
    externalId: "<some id>",
    title: "provide API specification using OpenAPI",
    description: "attribute  is not of type `object`",
    ruleLink: "https://zalando.github.io/restful-api-guidelines/#101",
   ....
}
```

### DIFF
--- a/server/zally-core/build.gradle.kts
+++ b/server/zally-core/build.gradle.kts
@@ -2,7 +2,7 @@ dependencies {
     kapt("com.google.auto.service:auto-service:1.0.1")
 
     api(project(":zally-rule-api"))
-    api("io.swagger.parser.v3:swagger-parser:2.1.9")
+    api("io.swagger.parser.v3:swagger-parser:2.1.12")
     api("io.github.config4k:config4k:0.5.0")
     implementation("com.google.auto.service:auto-service:1.0.1")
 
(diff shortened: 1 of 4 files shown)


tkrop:
:+1:

tkrop:
Oh, by the way, there is a drop in coverage. Can you think of a new test case covering it?

ktsypkina:
> Oh, by the way, there is a drop in coverage. Can you think of a new test case covering it?

I've added a new test case before but it's not recognising the new test case as one covering string input scenario

tkrop:
:+1:

ktsypkina:
👍

PR #1517, opened 2025-02-16
Update action/cache to v4

bocytko:
There is current a build warning. Will become an error soon.

> Your workflow is using a version of actions/cache that is scheduled for deprecation, actions/cache@v2. Please update your workflow to use either v3 or v4 of actions/cache to avoid interruptions. Learn more: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down

https://github.com/actions/cache/discussions/1510:

> We are deprecating some versions of this action. We recommend upgrading to version v4 or v3 as soon as possible before March 1st, 2025.

### DIFF
--- a/.github/workflows/build-zally.yml
+++ b/.github/workflows/build-zally.yml
@@ -17,7 +17,7 @@ jobs:
           java-version: '17'
           distribution: 'adopt'
       - name: Cache Gradle packages
-        uses: actions/cache@v2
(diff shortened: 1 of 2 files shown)


bocytko:
👍

tkrop:
:+1:

Issue #1528, opened 2025-06-12
References are not properly loaded when using OpenAPI version 3.1.0


fellipeale:
## Description
When sending a OpenApi 3.1.0 YAML to be validated, if it has $ref on it, Zally will not be able to process the request and return a NPE telling that the parameter.name must not be null.

## Environment
- **Application Version**: 2.1.1
- **OpenAPI schema version**: 3.1.0

## Steps to Reproduce
1. With the service running execute the following request"
`curl --location --request POST 'localhost:8000/api-violations' \
--header 'Content-Type: application/json' \
--data-raw '{
    "api_definition_string": "openapi: '\''3.1.0'\''\ninfo:\n  title: API 1\n  contact: \n    name: '\''Team One'\''\n  version: 1.0.0\npaths:\n  /items/{item-id}:\n    get:\n      parameters:\n        - $ref: '\''#/components/parameters/PathParameter'\''\n      responses:\n        default:\n          description: Response\n          content:\n            application/json:\n              schema:\n                type: string\ncomponents:\n  parameters:\n    PathParameter:\n      in: path \n      name: item-id\n      description: The id of the pet to retrieve\n      required: true\n      schema:\n        type: string"`

## Expected Behavior
We should get a 200 OK response with the validation errors

## Actual Behavior
A 500 error is returned
`
{
    "title": "Internal Server Error",
    "status": 500,
    "detail": "check invocation failed: id=Z001 title=Path parameters validation checkName=checkSchemaOrContentProperty reason=java.lang.NullPointerException: parameter.name must not be null"
}
`

## Additional Context
This issue appears to be specific to OpenAPI 3.1.0 schemas that use $ref for parameters, if the same schema is validated on version 3.0.0 it works without any issues.

fellipeale:
The change was deployed and it is working as expected, closing the issue.

PR #1529, opened 2025-06-23
Fix parsing of OpenApi 3.1 schemas

fellipeale:
The application wasn't being able to correctly parse OpenApi 3.1 schemas, the parameters references wasn't being properly fetched which was causing an unexpected NPE when doing the validation.

More info in the issue https://github.com/zalando/zally/issues/1528

### DIFF
--- a/server/zally-core/src/test/kotlin/org/zalando/zally/core/DefaultContextFactoryTest.kt
+++ b/server/zally-core/src/test/kotlin/org/zalando/zally/core/DefaultContextFactoryTest.kt
@@ -3,14 +3,15 @@ package org.zalando.zally.core
 import com.fasterxml.jackson.core.JsonPointer
 import org.assertj.core.api.Assertions.assertThat
 import org.intellij.lang.annotations.Language
-import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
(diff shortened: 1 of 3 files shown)


server/zally-core/src/test/kotlin/org/zalando/zally/core/DefaultContextFactoryTest.kt
```diff
@@ -3,14 +3,15 @@ package org.zalando.zally.core
...
-import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 import org.zalando.zally.core.ContentParseResultAssert.Companion.assertThat
+import org.zalando.zally.core.util.getAllParameters
 import org.zalando.zally.rule.api.Violation
 
 class DefaultContextFactoryTest {
 
     private val defaultContextFactory = DefaultContextFactory()
+    private val ref = "\$ref"
```
tkrop, line 14: Nit: While it is nice to know that this is working, I'm much for consistency. Since we else use `{'$'}`, I vote for sticking to this.

tkrop:
:+1:

tkrop:
:+1:

fellipeale:
:+1:

PR #1536, opened 2025-07-03
feat: adjust date time property conventions rule (#1536)

### LINKED ISSUE #1501: Relax naming convention for date/time properties
Please check if the PR https://github.com/zalando/restful-api-guidelines/pull/811 introduces changes which are relevant to the Zally project.

tkrop:
This pull request adjusts the date time property suffix rule according to the rule change conducted in https://github.com/zalando/restful-api-guidelines/pull/811.

fixes #1501.

### DIFF
--- a/server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/DateTimePropertiesSuffixRule.kt
+++ b/server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/DateTimePropertiesSuffixRule.kt
@@ -13,7 +13,7 @@ import org.zalando.zally.rule.api.Violation
     ruleSet = ZalandoRuleSet::class,
     id = "235",
     severity = Severity.SHOULD,
-    title = "Name date/time properties using the common suffix"
+    title = "Name date/time properties using the common conventions"
 )
 class DateTimePropertiesSuffixRule(rulesConfig: Config) {
 
--- a/server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/DateTimePropertiesSuffixRule.kt
+++ b/server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/DateTimePropertiesSuffixRule.kt
@@ -13,7 +13,7 @@ import org.zalando.zally.rule.api.Violation
     ruleSet = ZalandoRuleSet::class,
     id = "235",
     severity = Severity.SHOULD,
-    title = "Name date/time properties using the \"_at\" suffix"
+    title = "Name date/time properties using the common suffix"
 )
 class DateTimePropertiesSuffixRule(rulesConfig: Config) {
 
@@ -22,7 +22,7 @@ class DateTimePropertiesSuffixRule(rulesConfig: Config) {
         .map { Regex(it) }
         .toSet()
 
-    private val propertyFormats = setOf("date", "date-time")
+    private val propertyFormats = setOf("date", "time", "date-time")
 
     @Check(severity = Severity.SHOULD)
     fun validate(context: Context): List<Violation> {
--- a/server/zally-ruleset-zalando/src/main/resources/reference.conf
+++ b/server/zally-ruleset-zalando/src/main/resources/reference.conf
@@ -216,7 +216,10 @@ ProprietaryHeadersRule {
 
 DateTimePropertiesSuffixRule {
   patterns: [
-    ".*_at"
+    ".+_at",
+    "(.+_)*time(stamp)?(_.+)?",
+    "(.+_)*date(_.+)*",
+    "(.+_)*day(_.+)*"
   ]
 }
 
--- a/server/zally-ruleset-zalando/src/test/kotlin/org/zalando/zally/ruleset/zalando/DateTimePropertiesSuffixRuleTest.kt
+++ b/server/zally-ruleset-zalando/src/test/kotlin/org/zalando/zally/ruleset/zalando/DateTimePropertiesSuffixRuleTest.kt
@@ -2,7 +2,6 @@
 
 package org.zalando.zally.ruleset.zalando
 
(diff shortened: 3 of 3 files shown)


server/zally-ruleset-zalando/src/main/resources/reference.conf
```diff
@@ -216,7 +216,11 @@ ProprietaryHeadersRule {
 
 DateTimePropertiesSuffixRule {
   patterns: [
-    ".*_at"
+    ".*_at",
+    ".*_time",
+    ".*_timestamp",
+    ".*_date",
+    ".*_day",
```
SmilyOrg, line 223: Hmm, the rule says _contains_, but this needs it to be a suffix, should these be something like `".*_time(_.*)?` instead?

tkrop: Valid point, but I think it requires `.*_time$` instead, since it is using a regular expression matcher and there is no default `$` added to the pattern. I will create a regression test that proofs whether the current implementations is incorrect and how to fix it.

Thanks for making me think.

tkrop: Confirmed. The current implementation is for some reason only matching suffixes and fails for property patterns like `<my-var>_[time,time-date,date,day]_suffix`.

However, i'm also getting the impression that a `contains` does not make sense. We neither have we listed an example nor can I find one.

ePaul: I could think of things like "estimated_time_of_arrival", which has the `time` in the middle.

tkrop: Makes sense. I will try to adjust the regular expressions and tests to support the contains approach. Do we actually also `time`..., etc. to be positioned as prefix?

SmilyOrg: Just `time_of_arrival` might also make sense, in which case it would be a prefix. Maybe the check could be that it uses one of these words with an underscore on at least one side. Just `date`, `time` alone is probably not very useful.

tkrop: I fear, the last one is hard to express in regular expression, but I think that `(.*_)*<key>(_.*)*` would be an appropriate prefix/infix/suffix regular expression for our case.

SmilyOrg: What about https://regex-vis.com/?r=%28.%2B_date%28_.%2B%29%3F%7C%28.%2B_%29%3Fdate_.%2B%29&t=%5B%22date%22%2C%22date_of_arrival%22%2C%22creation_date%22%2C%22_date%22%2C%22date_%22%2C%22estimated_date_of_arrival%22%5D?

tkrop: I think that `(.+_)*<key>(_.+)*` is a simplified 100% equivalent version.

SmilyOrg: Almost, but not 100%. It's a bit too permissive for edge cases, but if we don't care about them, then it's also good.
image

tkrop: Okay, convinced. I'm using your expression.

ePaul:
:+1:

server/zally-ruleset-zalando/src/main/resources/reference.conf
tkrop: @SmilyOrg following your suggestion I see the disadvantage of the readability of the supported patterns in the error message:

```
[ .+_at, (.+_time(stamp)?(_.+)?|(.+_)?time(stamp)?_.+), (.+_date(_.+)?|, .+_)?date_.+), (.+_day(_.+)?|(.+_)?day_.+) ]
```

than

```
[ .+_at, (.+_)*time(_.+)*, (.+_)*time(stamp)?(_.+)*, (.+_)*date(_.+)*, (.+_)*date(_.+)* ]
```

tkrop: After extending the test to cover all edge-cases, I can say that both alternatives are equivalent. So I choose the simpler looking version.

server/zally-ruleset-zalando/src/main/kotlin/org/zalando/zally/ruleset/zalando/DateTimePropertiesSuffixRule.kt
```diff
@@ -13,7 +13,7 @@ import org.zalando.zally.rule.api.Violation
     ruleSet = ZalandoRuleSet::class,
     id = "235",
     severity = Severity.SHOULD,
-    title = "Name date/time properties using the \"_at\" suffix"
+    title = "Name date/time properties using the common suffix"
```
SmilyOrg:
```suggestion
    title = "Name date/time properties using the common conventions"
```

SmilyOrg (approved):

server/zally-ruleset-zalando/src/test/kotlin/org/zalando/zally/ruleset/zalando/DateTimePropertiesSuffixRuleTest.kt
```diff
@@ -54,25 +56,58 @@ class DateTimePropertiesSuffixRuleTest {
...
         val violations = rule.validate(DefaultContextFactory().getOpenApiContext(content))
-        assertThat(violations).isEmpty()
+        assertThat(violations.map { it.description }).containsExactly(
+            rule.generateMessage("was_created_at_suffix", "string", "date-time")
+        )
     }
 
     @Test
-    fun `should ignore fields with non-date and time types`() {
+    fun `should fail on prefix _at and succeed on prefix patterns`() {
```
ePaul:
```suggestion
    fun `should fail on prefix at_ and succeed on prefix patterns`() {
```

tkrop:
:+1:

ePaul:
:+1:

PR #1567, opened 2026-07-08
problem-meta-schema.json: remove exclusiveMaximum for status

ePaul:
* Related: #1566  (from https://github.com/zalando/restful-api-guidelines/pull/870)

We updated the schema to be compatible with the JSON schema version in OpenAPI 3.1, but forgot to update Zally.


ePaul:
Failing test:
```
> Task :zally-ruleset-zalando:test

UseProblemJsonRuleTest > should return violation if a wrong schema is used for Problem Details Object() FAILED
    java.lang.AssertionError at UseProblemJsonRuleTest.kt:71

UseProblemJsonRuleTest > should return violation if a incorrect schema is used for Problem Details Object (OpenAPI 2)() FAILED
    java.lang.AssertionError at UseProblemJsonRuleTest.kt:109
```
I guess I'll need to actually check this out locally ...

ePaul:
I'm not sure whether we should add a test that the old version still works?

ePaul:
:+1:

aswinramesh:
:+1:

